PR #13, opened 2024-01-29
Update install instructions

### LINKED ISSUE #12: Cargo install procedure needs updating
Following the instructions in the README I get the following error when executing `cargo install`:

```
error: Using `cargo install` to install the binaries for the package in current working directory is no longer supported, use `cargo install --path .` instead. Use `cargo build` if you want to simply build the package.
```

Seems it needs path adding.

jatkinson1000:
Closes #12 

Update README to add `--path .` to `cargo install` command as is now required.


LiamPattinson:
Cheers for catching that!

PR #45, opened 2024-09-24
Print context of violations

ZedThree:
Example:

```text
warning: test.f90:90:3: P011 prefer 'real(real64)' to 'double precision' (see 'iso_fortran_env')
   |
89 |   ! Should trigger for use of 'double precision'
90 |   double precision function double_prec(x)
   |   ^ P011
91 |     double precision, intent(in) :: x
92 |     double_prec = 2 * x
   |

warning: test.f90:91:5: P011 prefer 'real(real64)' to 'double precision' (see 'iso_fortran_env')
   |
89 |   ! Should trigger for use of 'double precision'
90 |   double precision function double_prec(x)
91 |     double precision, intent(in) :: x
   |     ^ P011
92 |     double_prec = 2 * x
93 |   end function
   |

warning: test.f90:97:1: M001 subroutine not contained within (sub)module or program
   |
96 | ! This function should trigger for missing enclosing module
97 | subroutine triple(x)
   |  ^ M001
98 |   integer, intent(inout) :: x
99 |   x = x * 3
   |
```

A couple of annoying things:

- Something really weird going on here, where I can't get it to
  put the annotation in the first column: it's either in column 2
  or the end of the previous line. But does appear to be right
  for other columns!
- Some annoyance here: we *have* to have some level prefix to our
  message. Might be fixed in future version of annotate-snippets
  -- or we use an earlier version with more control.
- Also, we could use `.origin(path)` to get the filename and
  line:col automatically, but see above about off-by-one error

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -333,7 +333,7 @@ fn format_violation_line_col(
     // put the annotation in the first column: it's either in column 2
     // or the end of the previous line. But does appear to be right
     // for other columns!
-    let label_offset = offset_up_to_line + col.saturating_sub(1).max(1);
+    let label_offset = offset_up_to_line + col.saturating_sub(1);
 
     // Some annoyance here: we *have* to have some level prefix to our
     // message. Might be fixed in future version of annotate-snippets
@@ -344,7 +344,11 @@ fn format_violation_line_col(
     let snippet = Level::Warning.title(&message_line).snippet(
         Snippet::source(&content_slice)
             .line_start(start_index)
-            .annotation(Level::Error.span(label_offset..label_offset).label(code)),
+            .annotation(
+                Level::Error
+                    .span(label_offset..label_offset.saturating_add(1))
+                    .label(code),
+            ),
     );
 
     let renderer = Renderer::styled();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,9 +4,10 @@ pub mod cli;
 pub mod explain;
 mod rules;
 mod settings;
+use annotate_snippets::{Level, Renderer, Snippet};
 use anyhow::Context;
 use ast::{named_descendants, parse};
-use colored::Colorize;
+use colored::{ColoredString, Colorize};
 use lazy_regex::regex_captures;
 use settings::Settings;
 use std::cmp::Ordering;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -269,15 +270,88 @@ impl fmt::Display for Diagnostic {
...
+    // Some annoyance here: we *have* to have some level prefix to our
+    // message. Might be fixed in future version of annotate-snippets
+    // -- or we use an earlier version with more control.
+    // Also, we could use `.origin(path)` to get the filename and
+    // line:col automatically, but see above about off-by-one error
+    let message_line = format!("{}:{}:{}: {} {}", path, line, col, code, message);
+    let snippet = Level::Warning.title(&message_line).snippet(
+        Snippet::source(&content_slice)
+            .line_start(start_index)
+            .annotation(Level::Error.span(label_offset..label_offset).label(code)),
```
LiamPattinson, line 347:
```suggestion
            .annotation(Level::Error.span(label_offset..label_offset.saturating_add(1)).label(code)),
```

src/lib.rs
```diff
@@ -269,15 +270,88 @@ impl fmt::Display for Diagnostic {
...
+    // appear in `lines`!
+    let offset_up_to_line = lines[start_index.saturating_sub(1)..line.saturating_sub(1)]
+        .iter()
+        .fold(0, |acc, line| acc + line.chars().count() + 1);
+
+    // Something really weird going on here, where I can't get it to
+    // put the annotation in the first column: it's either in column 2
+    // or the end of the previous line. But does appear to be right
+    // for other columns!
+    let label_offset = offset_up_to_line + col.saturating_sub(1).max(1);
```
LiamPattinson, line 336:
```suggestion
    let label_offset = offset_up_to_line + col.saturating_sub(1);
```

LiamPattinson (commented):
Amazing, this looks so much better.

I think I've figured out a fix to the weird off-by-one error, but I haven't tested it enough to be 100% sure.

LiamPattinson:
As an aside, it might be nice to introduce a `--concise` mode that doesn't include the snippet, as it seems to be taking about 4 times as long when applied to a project I'm using for simple benchmarking (although 2 seconds to produce over 10000 annotated code snippets is still impressively fast as far as I'm concerned).

PR #47, opened 2024-09-26
Update tree-sitter-fortran to v0.1.0

ZedThree:
Mostly fine, except some breaking changes for the literal kind detection.

Starting to get my head around the rust ts bindings. Worked out that we can write:

```rust
if let Some(foo) = <something> {
  ...
}
```

as just:

```rust
let foo = <something>?;
```

More straight-line code, less indentation! That's the power of monads! or something

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,8 +353,9 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-fortran"
-version = "0.0.1"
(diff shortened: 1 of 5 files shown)


LiamPattinson:
Thanks for this, the regex checks in that part of the code were there just to simplify the parsing of 'size' nodes, but it looks like that's improved in the latest tree-sitter-fortran version.

Also a huge thanks for the improved handling of preprocessor statements! Running on Epoch, it's down from over 3000 warnings (mostly syntax errors) to just over 400.

PR #49, opened 2024-10-01
Return multiple violations from rules

### LINKED ISSUE #48: Returning multiple violations from the same rule
For example, when checking for missing `intent`s we need to find variable declarations with identifiers in the procedure's parameter list, and then check for the missing attribute. This means it's probably more sensible to collect all such violations for a given procedure at the same time.

Two suggestions:

1. Add a new rule trait that returns a `Vec` of violations, and `append` to the current list
2. Something like ruff's structure, with a `checker` object that's passed into each rule, which can then modify it as appropriate

ZedThree:
Closes #48


LiamPattinson:
All looks good to me!

PR #50, opened 2024-10-01
Add rule for missing `intent` attribute

### LINKED ISSUE #40: Catch missing `intent` attribute
Procedure arguments should have `intent` attributes. Default is `intent(inout)`, this could be an auto-fix. In principle it should also be possible to work out if this could be more specifically `intent(in)` or `intent(out)`, based on whether the arguments are only ever read from/written to

ZedThree:
Closes #40 

~~Includes #49~~

### DIFF
--- a/src/rules/typing/intent.rs
+++ b/src/rules/typing/intent.rs
@@ -1,4 +1,4 @@
-use crate::ast::to_text;
+use crate::ast::{child_with_name, to_text};
 use crate::settings::Settings;
 use crate::{ASTRule, Rule, Violation};
 use tree_sitter::Node;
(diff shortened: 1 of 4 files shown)


src/rules/typing/intent.rs
```diff
@@ -0,0 +1,153 @@
...
+                !decl
+                    .children_by_field_name("attribute", &mut decl.walk())
+                    .any(|attr| {
+                        to_text(&attr, &src)
+                            .unwrap_or("")
+                            .to_lowercase()
+                            .starts_with("intent")
+                    })
+            })
+            .filter_map(|decl| {
```
LiamPattinson, line 70:
```suggestion
            .flat_map(|decl| {
```

Needs to be coupled with the next comment...

src/rules/typing/intent.rs
```diff
@@ -0,0 +1,153 @@
...
+                        None
+                    })
+                    .filter_map(|(dummy, name)| {
+                        let msg = format!(
+                            "{procedure_kind} argument '{name}' missing 'intent' attribute"
+                        );
+                        Some(Violation::from_node(&msg, &dummy))
+                    })
+                    .collect();
+                Some(violations)
```
LiamPattinson, line 97:
```suggestion
                violations
```

... and the next one...

src/rules/typing/intent.rs
```diff
@@ -0,0 +1,153 @@
...
+            })
+            .flatten()
```
LiamPattinson, line 99:
```suggestion

```

LiamPattinson (changes requested):
Looks like the rule is working perfectly, but I found some of the logic hard to follow. After staring at it for a while I realised it's because a couple of the `filter_map` calls aren't actually filtering anything, as their functions can only return `Some(x)`.

If you check the Clippy CI results, there are also quite a few warnings for things like question marks, reference symbols, and lifetimes that aren't needed. Please could you patch those up before the merge?

ZedThree:
Ah, I didn't see the clippy notes! Might need to turn the warnings into errors to fail the build?

Thanks for the comments -- the `filter_map`s are a remnant of the first attempt to get this working.

src/rules/typing/intent.rs
ZedThree: This is actually required, because we build up a `Vec<Vec<Violation>>`: the inner `Vec` comes from the potentially multiple variables on the same declaration line, and the outer one comes from each separate declaration line.

ZedThree: Ah, but you switch to `flat_map` above too, I see

LiamPattinson:
Looks good, and I'm happy for some of the other Clippy warnings to be fixed in https://github.com/PlasmaFAIR/fortitude/pull/51

PR #51, opened 2024-10-01
Fix clippy warnings

ZedThree:
also error on warnings in CI


LiamPattinson:
Thanks for this! I didn't know about the `--all-targets` flag, so didn't realise that the tests weren't being linted this whole time.

PR #53, opened 2024-10-02
Add rule for missing label on `exit`/`cycle`

ZedThree:
Mostly following the equivalent rule in `stylist` except:

- also capture `cycle` as well as `exit`
- only applies to `do` loops with a name
- unfortunately, warns twice for unlabelled `exit/`cycle` in nested loops

Last point could probably be fixed with some more sophisticated logic: walking the tree, and only descending into non-`do` nodes. Then I think we'd be able to confidently suggest the correct label for a fix.

Second point could be turned into another rule: `exit`/`cycle` in nested loops should be labelled.

Also, I've just put this into the "style" category, but I'm not sure that's completely appropriate. Maybe we need some community input into the categorisation of rules.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -93,7 +93,8 @@ impl<'a> Iterator for DepthFirstIteratorExcept<'a> {
     }
 }
 
-/// Iterate over all nodes beneath the current node in a depth-first manner.
+/// Iterate over all nodes beneath the current node in a depth-first manner, never going deeper
+/// than any node types in the exceptions list.
 pub fn descendants_except<'a, I>(node: &'a Node, exceptions: I) -> impl Iterator<Item = Node<'a>>
 where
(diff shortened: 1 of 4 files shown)


LiamPattinson:
> Second point could be turned into another rule: exit/cycle in nested loops should be labelled.

I think this is a good idea, especially since the PR I raised into this branch will result in no warning if an inner loop containing an exit/cycle is unlabelled. Please could you raise a new issue for it?

> Also, I've just put this into the "style" category, but I'm not sure that's completely appropriate. Maybe we need some community input into the categorisation of rules.

I'm also not sure about this, as this feels more like ensuring correctness instead of enforcing purely stylistic rules. However, it doesn't fit any of the other categories, and I can't immediately think of new category to introduce, so I think it would be best to leave it here for now.

PR #55, opened 2024-10-02
Fix repeated warnings for unlabelled exit in nested loops

LiamPattinson:
Adds a variant on the depth-first tree iterator that doesn't go deeper than any nodes within an exception list. Used to avoid repeated warnings in the new rule that warns for unlabelled exit/cycle statements (https://github.com/PlasmaFAIR/fortitude/pull/53). From here, it should be easy to also report which label is most likely missing.

---

@ZedThree Is this the sort of solution you were looking for? If not, feel free to close this PR without merging.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -93,7 +93,8 @@ impl<'a> Iterator for DepthFirstIteratorExcept<'a> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


ZedThree (approved):
Great, thanks @LiamPattinson!

PR #57, opened 2024-10-03
Add rule for assumed size dummy arguments

ZedThree:


ZedThree:
Rebased on `main` with the new trait

LiamPattinson (approved):
It might be worth expanding the explanation a little, but otherwise all looks good.

src/rules/typing/assumed_size.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn new(_settings: &Settings) -> Self {
+        AssumedSize {}
+    }
+
+    fn explain(&self) -> &'static str {
+        "
+        Assumed size dummy arguments declared with a star `*` as the size should be
+        avoided, and assumed shape used instead. There are several downsides to assumed
+        size, the main one being that the compiler is not able to determine the array
+        bounds, so it is not possible to check for array overruns.
```
LiamPattinson, line 21:
```suggestion
        Assumed size dummy arguments declared with a star `*` as the size should be
        avoided, and assumed shape with a colon `:` should be used instead. There are
        several downsides to assumed size, the main one being that the compiler is
        not able to determine the array bounds, so it is not possible to check for
        array overruns.
```

ZedThree:
Ok, let me add a bit more too

src/rules/typing/assumed_size.rs
```diff
@@ -0,0 +1,139 @@
...
+        AssumedSize {}
+    }
+
+    fn explain(&self) -> &'static str {
+        "
+        Assumed size dummy arguments declared with a star `*` as the size should be
+        avoided, and assumed shape with a colon `:` should be used instead. There are
+        several downsides to assumed size, the main one being that the compiler is
+        not able to determine the array bounds, so it is not possible to check for
+        array overruns or to use the array in whole-array expressions.
```
LiamPattinson, line 22: Looks good! With the addition of the second paragraph and example code, I think this can now be shortened:

```suggestion
        Assumed size dummy arguments declared with a star `*` as the size should be
        avoided. There are several downsides to assumed size, the main one being
        that the compiler is not able to determine the array bounds, so it is not
        possible to check for array overruns or to use the array in whole-array 
        expressions.
```

PR #58, opened 2024-10-03
Column number for text rules

ZedThree:
A little more similar treatment between test and AST based rules

(Also add tests for the trailing whitespace and too-long line rules)


LiamPattinson (approved):
All looks good!

PR #59, opened 2024-10-03
Convert AST free functions into a trait

ZedThree:
Not completely sure about this to be honest, adds a bit of overhead in terms of code, but end results feels more rusty and is I think nicer.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -54,18 +54,6 @@ impl<'a> Iterator for DepthFirstIterator<'a> {
     }
 }
 
-/// Iterate over all nodes beneath the current node in a depth-first manner.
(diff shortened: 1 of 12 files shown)


ZedThree:
One alternative is the "newtype" pattern:

```rust
pub struct FortitudeNode(Node);

impl FortitudeNode {
...
```

The _big_ downside of this is accessing `Node` methods requires `node.0.<method>`. We could use methods to delegate these calls, but then that's suddenly a _lot_ more code.

This is unfortunately probably the shortest implementation!

ZedThree:
Probably get this in first and fix up the other PRs

LiamPattinson (approved):
It would be nice if there were a way to avoid the need to pre-declare the extra functions, but otherwise looks much better than before.

PR #72, opened 2024-10-04
Add rule for initialisation in declarations

### LINKED ISSUE #60: Initialisation in declaration should require explicit `save`
Initialisation in declarations gets an implicit `save`, meaning the initialisation is not done on subsequent entry into the routine.

Bad:

```fortran
subroutine foo()
  integer :: bar = 1
end subroutine
```

Good:

```fortran
subroutine foo()
  integer, save :: bar = 1
end subroutine
```

Also good:

```fortran
subroutine foo()
  integer :: bar
  bar = 1
end subroutine
```

ZedThree:
Closes #60

The explanation is longer than the check!

Might be useful to pull out the "attributes as text" bit into a method on the `FortitudeNode` trait, as it's used in a couple of rules now. Maybe also the "find a specific parent" too


LiamPattinson (approved):
All looks good!

Issue #73, opened 2024-10-04
Assumed size on `character` needs more nuance


ZedThree:
This is actually a little more complicated than the rule I implemented.

Here's a similar rule from `stylist`:

```fortran
program cases
    ! A char array outside a function or subroutine, no exception
    character (*) :: autochar_glob

    contains

    subroutine char_input(autochar_in, autochar_inout, autochar_out, fixedchar)
        ! A char array with proper intent, no exception
        character(*), intent(in)       :: autochar_in
        ! A char array with disallowed intent, exception
        character(*), intent(inout)    :: autochar_inout
        ! A char array with disallowed intent, exception
        character(len=*), intent(out)  :: autochar_out
        ! A char array not passed as a parameter, no exception
        character(*)                   :: autochar_var
        ! A char array with fixed length, no exception
        character(len=10), intent(out) :: fixedchar
        ! A declaration with non-intent attribute, no exception
        character(len=*), parameter :: alt_attr = "sample"
    end subroutine char_input

end program cases
```

Weirdly, `character(len=*)` _is_ the correct way to pass a variable-length string, and the compiler does know how long it is.

If it's not a dummy argument, `character(*)` doesn't actually compile. I think we could actually catch that anyway, and recommend `character(:), allocatable` instead.

`character*(*)` is still unacceptable, but should probably be handled separately as well.

So:

1. fix current rule to ignore characters
2. add new rule to check `intent` of character dummys
3. add new rule to check `character*(*)`

Issue #1, opened 2024-01-12
Avoid `character*(*)`


LiamPattinson:
Already catching non-standard extensions like `real*8`, but not catching a deprecated way to specify an assumed size string, `character*(*)`. [Discussion on fortran-lang](https://fortran-lang.discourse.group/t/what-does-character-mean/330/2).

PR #74, opened 2024-10-08
Assumed size rule doesn't catch character len

LiamPattinson:
Avoid catching `character(len=*)`, `character(*)`, and `character*(*)` when detecting assumed sizes.

Partly resolves https://github.com/PlasmaFAIR/fortitude/issues/73, though a further rule is needed to check that the intent of `character(*)` is `intent(in)`, and another is needed to avoid `character*(*)` (this has been in need of its own rule since Issue https://github.com/PlasmaFAIR/fortitude/issues/1!)


ZedThree:
Oops, sorry, didn't realise you were already doing this! I have pretty much the same fix in #75, plus the two new rules

LiamPattinson:
No worries, your fix looks more robust anyway, so let's go with that.

PR #82, opened 2024-10-09
Add rule for deprecated relational operators

### LINKED ISSUE #63: Prefer `>` over `.gt.` etc
Should be pretty easy to capture, it's just the `relational_expression` node. Also very easy to supply a fix, there's a straight mapping from the text operator to the symbol one

ZedThree:
Closes #63 

One more easy rule for 0.3.0! Explanation could do with a little work though

### DIFF
--- a/src/rules/mod.rs
+++ b/src/rules/mod.rs
@@ -15,7 +15,7 @@ register_rules! {
     (Category::Style, "S001", TEXT, style::line_length::LineTooLong, LineTooLong),
     (Category::Style, "S021", AST, style::exit_labels::MissingExitOrCycleLabel, MissingExitOrCycleLabel),
     (Category::Style, "S041", AST, style::old_style_array_literal::OldStyleArrayLiteral, OldStyleArrayLiteral),
-    (Category::Style, "S051", AST, style::relational_symbols::RelationalSymbol, RelationalSymbol),
+    (Category::Style, "S051", AST, style::relational_operators::DeprecatedRelationalOperator, DeprecatedRelationalOperator),
     (Category::Style, "S101", TEXT, style::whitespace::TrailingWhitespace, TrailingWhitespace),
     (Category::Typing, "T001", AST, typing::implicit_typing::ImplicitTyping, ImplicitTyping),
     (Category::Typing, "T002", AST, typing::implicit_typing::InterfaceImplicitTyping, InterfaceImplicitTyping),
--- a/src/rules/style/mod.rs
+++ b/src/rules/style/mod.rs
@@ -1,5 +1,5 @@
 pub mod exit_labels;
 pub mod line_length;
 pub mod old_style_array_literal;
-pub mod relational_symbols;
+pub mod relational_operators;
 pub mod whitespace;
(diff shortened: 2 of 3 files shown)


src/rules/style/relational_symbols.rs
```diff
@@ -0,0 +1,87 @@
...
+        BTreeMap::from([
+            (".gt.", ">"),
+            (".ge.", ">="),
+            (".lt.", "<"),
+            (".le.", "<="),
+            (".eq.", "=="),
+            (".ne.", "/="),
+        ])
+    };
+}
```
LiamPattinson, line 20:
```suggestion
fn map_relational_symbols(name: &str) -> Option<&'static str> {
    match name {
        ".gt." => Some(">"),
        ".ge." => Some(">="),
        ".lt." => Some("<"),
        ".le." => Some("<="),
        ".eq." => Some("=="),
        ".ne." => Some("/="),
        _ => None,
    }
}
```

src/rules/style/relational_symbols.rs
```diff
@@ -0,0 +1,87 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::settings::default_settings;
+    use crate::violation;
+    use pretty_assertions::assert_eq;
+    use textwrap::dedent;
+
+    #[test]
+    fn test_old_style_array_literal() -> anyhow::Result<()> {
```
LiamPattinson, line 60:
```suggestion
    fn test_relational_symbol() -> anyhow::Result<()> {
```

src/rules/style/relational_symbols.rs
```diff
@@ -0,0 +1,87 @@
...
+        `>=`, `<`, and so on.
+        "
+    }
+}
+
+impl ASTRule for RelationalSymbol {
+    fn check(&self, node: &Node, src: &str) -> Option<Vec<Violation>> {
+        let relation = node.child(1)?;
+        let symbol = relation.to_text(src)?.to_lowercase();
+        let new_symbol = RELATIONAL_NAMES.get(symbol.as_str())?;
```
LiamPattinson, line 41: If using the match function I suggested...
```suggestion
        let new_symbol = map_relational_symbols(symbol.as_str())?;
```

src/rules/style/relational_symbols.rs
```diff
@@ -0,0 +1,87 @@
+use std::collections::BTreeMap;
+
```
LiamPattinson, line 2:
```suggestion
```

src/rules/style/relational_symbols.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::ast::FortitudeNode;
+use crate::settings::Settings;
+use crate::{ASTRule, Rule, Violation};
+use lazy_static::lazy_static;
```
LiamPattinson, line 6:
```suggestion
```

LiamPattinson (commented):
I've added a recommendation to switch out the lazy_static BTreeMap for a simpler function using a match statement. Also fixed the test name.

Otherwise all looks good! I'm not really sure what else needs to be added for the explanation.

src/rules/style/relational_symbols.rs
ZedThree: Ah nice, yes that's much more idiomatic! Although we know it will be exactly one of these, so can return just `&'static str` and use `_ => unreachable!()`

LiamPattinson:
All looks good!

LiamPattinson (approved):

Issue #10, opened 2024-01-23
`end` statements should be explicit


LiamPattinson:
The following are all equivalent:

```f90
function f(...)
   ...
end function f

function f(...)
  ...
end function

function f(...)
  ...
end
```

Similarly for subroutines and modules. However, the first form is preferred by convention. Should add a `BestPractices` rule to check that the first form is in use (perhaps activated under strict mode rather than default), and an `Error` rule to check that the names match.

PR #84, opened 2024-10-10
Add rule for non-explicit end statements

LiamPattinson:
Closes https://github.com/PlasmaFAIR/fortitude/issues/10. I think I've caught every type of end statement here -- let me know if I've missed one.

### DIFF
--- a/src/rules/style/end_statements.rs
+++ b/src/rules/style/end_statements.rs
@@ -1,0 +1,220 @@
+use crate::ast::FortitudeNode;
+use crate::settings::Settings;
+use crate::{ASTRule, Rule, Violation};
+use tree_sitter::Node;
+
+pub struct AbbreviatedEndStatement {}
+
+impl Rule for AbbreviatedEndStatement {
+    fn new(_settings: &Settings) -> Self {
+        Self {}
+    }
+
+    fn explain(&self) -> &'static str {
+        "
+        End statements should specify what kind of construct they're ending, and the
+        name of that construct. For example, prefer this:
+
+        ```
+        module mymodule
+          ...
+        end module mymodule
+        ```
+
+        To this:
+
+        ```
+        module mymodule
+          ...
+        end
+        ```
+
+        Or this:
+
+        ```
+        module mymodule
+          ...
+        end module
+        ```
+
+        Similar rules apply for submodules, programs, functions, subroutines, and
+        derived types.
+        "
+    }
+}
+
+fn map_start_statement(kind: &str) -> &'static str {
+    match kind {
+        "module" => "module_statement",
+        "submodule" => "submodule_statement",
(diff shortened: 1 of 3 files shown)


src/rules/style/end_statements.rs
```diff
@@ -0,0 +1,220 @@
...
+
+impl ASTRule for AbbreviatedEndStatement {
+    fn check<'a>(&self, node: &'a Node, src: &'a str) -> Option<Vec<Violation>> {
+        let kind = node.kind();
+        let end_kind = map_end_statement(kind);
+        let end_node = node.child_with_name(end_kind)?;
+
+        // If end node is named, move on.
+        // Not catching incorrect end statement name here, as the compiler should
+        // do that for us.
```
ZedThree, line 93: I do think that would be a good (separate) warning though -- if we can do easy compiler warnings, we can catch them immediately as people write them (and apply autofixes!)

src/rules/style/end_statements.rs
```diff
@@ -0,0 +1,220 @@
...
+        end module
+        ```
+
+        Similar rules apply for submodules, programs, functions, subroutines, and
+        derived types.
+        "
+    }
+}
+
+fn map_start_statement(kind: &str) -> &'static str {
```
ZedThree, line 46: _Could_ squash all three functions into one and return a tuple?

ZedThree (approved):

ZedThree:
There _are_ other block constructs that also can end with just `end`: `if`, `do`, `where`, `associate`, `block`, `select case/rank/type`

Not sure if they should be covered by this rule or not, as some of them can have block labels (`label: do`), and some can't have any name at all, so they would certainly need different logic.

LiamPattinson:
Managed to mess up the commit history a little by forgetting to pull the merge conflict before doing a bit more work.

I've made it trigger on the end statement itself instead of the declaration, which results in somewhat cleaner code and it means the errors in the test are generated in a more sensible order. I've renamed it to `UnnamedEndStatement` to differentiate it from the use of plain `end` compared to `end <something>` for unlabelled constructs.

At least with gfortran, using `end` instead of `end if` or `end do` results in an error. We probably should still catch the ones that can be labelled though, as `mydo: do i=1,10...end do mydo` does work. For now I've left that as a TODO.

src/rules/style/end_statements.rs
LiamPattinson: Good point, when we get around to language server bindings that sort of error would be good to catch.

src/rules/style/end_statements.rs
LiamPattinson: Done in my latest changes, and I simplified it a little so that I only need two outputs instead of three.

PR #85, opened 2024-10-15
Add ranges to `Violation`

### LINKED ISSUE #44: `Violation` should have `start` and `end`
Rather than:

```rust
pub enum ViolationPosition {
    None,
    Line(usize),
    LineCol((usize, usize)),
}
```

which just points to a specific place in the source, having a start and end allows pointing to exactly the offending piece of code and underlining the relevant part(s)

I'm not sure what `None` is supposed to represent here, but it could be encoded as `(0, 0)`. Similarly, `Line` could be the start & end of the whole line.

It would also be most useful if this was in terms of offset into the file, with some conversion utility between offset and line/column numbers. (This is at least partly what all that extra structure is for in ruff)

ZedThree:
Closes #44 

Uses the `ruff_source_file` and `ruff_text_size` crates to implement ranges for violations. Among other things, using these directly should make it easier to adapt other parts of ruff.

A bit rough but it works. Required some rejigging of the tests that I'm not overly happy with -- namely using zero-indexed line/column for the start/end ranges. Might be better to just use offsets in the tests as well, but then we probably want something to make it easier to work out the expected ranges. Maybe a way to optionally print the parse tree for the test snippet? Or to print the snippet with line-start-offsets, e.g.:

```
     0    5    1
               0   
0  | program test
13 |   implicit none
30 | end program test
```

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -10,7 +10,6 @@ use crate::{Diagnostic, Violation};
 use colored::Colorize;
 use itertools::{chain, join};
 use ruff_source_file::{SourceFile, SourceFileBuilder};
-use std::fs::read_to_string;
 use std::path::{Path, PathBuf};
 use walkdir::WalkDir;
 
@@ -114,6 +113,27 @@ fn check_file(
     Ok(violations)
 }
 
+/// Wrapper around `std::fs::read_to_string` with some extra error
+/// checking.
+///
+/// Check that the file length is representable as `u32` so
+/// that we don't need to check when converting tree-sitter offsets
+/// (usize) into ruff offsets (u32)
+fn read_to_string(path: &Path) -> std::io::Result<String> {
+    let metadata = path.metadata()?;
+    let file_length = metadata.len();
+
+    if TryInto::<u32>::try_into(file_length).is_err() {
+        #[allow(non_snake_case)]
+        let length_in_GiB = file_length as f64 / 1024.0 / 1024.0 / 1024.0;
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::Other,
+            format!("larger than maximum 4 GiB ({length_in_GiB} GiB)"),
+        ));
+    }
+    std::fs::read_to_string(path)
+}
+
 /// Check all files, report issues found, and return error code.
 pub fn check(args: CheckArgs) -> i32 {
     let settings = Settings {
@@ -141,17 +161,6 @@ pub fn check(args: CheckArgs) -> i32 {
                 };
 
                 let file = SourceFileBuilder::new(filename.as_ref(), source.as_str()).finish();
-
-                // If we can convert the total file size to u32, then
-                // we don't need to check when converting tree-sitter
-                // offsets (usize) into ruff offsets (u32)
-                if TryInto::<u32>::try_into(file.source_text().len()).is_err() {
-                    let violation = violation!(format!("File too large"));
-                    let diagnostic = Diagnostic::new(&empty_file, "E000", &violation);
-                    println!("{diagnostic}");
-                    total_errors += 1;
-                    continue;
-                }
 
                 match check_file(&path_rules, &text_rules, &ast_entrypoints, &path, &file) {
                     Ok(violations) => {
--- a/src/check.rs
+++ b/src/check.rs
@@ -142,6 +142,17 @@ pub fn check(args: CheckArgs) -> i32 {
 
                 let file = SourceFileBuilder::new(filename.as_ref(), source.as_str()).finish();
 
+                // If we can convert the total file size to u32, then
+                // we don't need to check when converting tree-sitter
+                // offsets (usize) into ruff offsets (u32)
+                if TryInto::<u32>::try_into(file.source_text().len()).is_err() {
+                    let violation = violation!(format!("File too large"));
+                    let diagnostic = Diagnostic::new(&empty_file, "E000", &violation);
+                    println!("{diagnostic}");
+                    total_errors += 1;
+                    continue;
+                }
+
                 match check_file(&path_rules, &text_rules, &ast_entrypoints, &path, &file) {
                     Ok(violations) => {
                         let mut diagnostics: Vec<Diagnostic> = violations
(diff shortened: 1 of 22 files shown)


src/check.rs
```diff
@@ -124,8 +125,15 @@ pub fn check(args: CheckArgs) -> i32 {
             let text_rules = text_rule_map(&rules, &settings);
             let ast_entrypoints = ast_entrypoint_map(&rules, &settings);
             let mut total_errors = 0;
-            for file in get_files(&args.files) {
-                match check_file(&path_rules, &text_rules, &ast_entrypoints, &file) {
+            for path in get_files(&args.files) {
+                // FIXME!!
+                let source = read_to_string(&path).unwrap();
```
LiamPattinson, line 130: I'd prefer to fix this now rather than later. Maybe something like this:

```suggestion
            let err_source = "kj23brdf9kshb23owihqb";
            for path in get_files(&args.files) {
                let source = read_to_string(&path).unwrap_or(err_source.to_string());
                if source == err_source {
                    let file = SourceFileBuilder::new(path.to_string_lossy().as_ref(), "")
                        .finish();
                    let err_msg = format!("Failed to process: {}", path.display());
                    let violation = violation!(&err_msg);
                    let diagnostic = Diagnostic::new(&file, "E000", &violation);
                    println!("{}", diagnostic);
                    total_errors += 1;
                    continue;
                }
```
The diagnostic handling bit is very similar to a section lower down in this function, so should probably be pulled out into a new function.

src/lib.rs
```diff
@@ -70,31 +72,53 @@ pub struct Violation {
...
     }
 
     pub fn from_node<T: AsRef<str>>(message: T, node: &tree_sitter::Node) -> Self {
-        let position = node.start_position();
         Violation::new(
             message,
-            ViolationPosition::LineCol((position.row + 1, position.column + 1)),
+            ViolationPosition::Range(TextRange::new(
+                TextSize::try_from(node.start_byte()).unwrap(),
+                TextSize::try_from(node.end_byte()).unwrap(),
```
LiamPattinson, line 91: These unwraps feel a bit scary. Should this instead return `Result<Self>` and be handled with `?` everywhere?

src/lib.rs
```diff
@@ -157,145 +178,157 @@ pub trait ASTRule: Rule {
...
-/// Read filename into vec of strings
-fn read_lines(filename: &PathBuf) -> Vec<String> {
-    std::fs::read_to_string(filename)
-        .unwrap() // panic on possible file-reading errors
-        .lines() // split the string into an iterator of string slices
-        .map(String::from) // make each slice into a string
-        .collect() // gather them together into a vector
-}
-
 fn format_violation_line_col(
```
LiamPattinson, line 249:
```suggestion
fn format_violation_range(
```

src/lib.rs
```diff
@@ -157,145 +178,157 @@ pub trait ASTRule: Rule {
...
-    let snippet = Level::Warning.title(&message_line).snippet(
-        Snippet::source(&content_slice)
-            .line_start(start_index)
-            .annotation(
-                Level::Error
-                    .span(label_offset..label_offset.saturating_add(1))
-                    .label(code),
-            ),
+
+    // let message_line = format!("{path}:{st}:{}: {} {}", path, line, col, code, message);
```
LiamPattinson, line 301: Can this be removed?

src/lib.rs
```diff
@@ -157,145 +178,157 @@ pub trait ASTRule: Rule {
...
+        .count();
+    let end_char = source[TextRange::up_to(message_range.end())]
+        .chars()
+        .count();
 
     // Some annoyance here: we *have* to have some level prefix to our
     // message. Might be fixed in future version of annotate-snippets
     // -- or we use an earlier version with more control.
     // Also, we could use `.origin(path)` to get the filename and
     // line:col automatically, but see above about off-by-one error
```
LiamPattinson, line 299: This refers back to earlier comments that were removed. Can this go too, or does it need to be reworded?

LiamPattinson (commented):
All generally looks good, with just a few comments.

Overall, I'm a little uneasy with the direct use of Ruff internals from a maintainability point of view. It makes it harder for others to contribute if they have to figure out how to use largely-undocumented internals of another project (although it's not like ours is particularly well documented...).

On the other hand, this _works_, and I'd much prefer to make fast progress now than to get bogged down essentially reimplementing the same features ourselves. If it becomes a major headache in future, I imagine we'll be able to swap out the Ruff code for our own implementation fairly easily, especially if we aim for a similar API.

LiamPattinson:
Should we also swap around where we write the error code and where we write the message in our diagnostics?

src/check.rs
ZedThree: Yep, or pull the loop body out into a function that returns `Result`, and then can propagate the error up like that?

src/lib.rs
ZedThree: What we can do is check the file length as soon as we open it to ensure it's representable as `u32` (i.e. < 4GB), and then we know all these casts/unwraps are fine.

src/check.rs
LiamPattinson: Yes, I think that would be a better approach

src/lib.rs
LiamPattinson: Ah, if that's the only time that this can panic, then it should be fine to leave them in, although maybe with a comment to explain why.

src/check.rs
```diff
@@ -124,8 +125,35 @@ pub fn check(args: CheckArgs) -> i32 {
...
+                        continue;
+                    }
+                };
+
+                let file = SourceFileBuilder::new(filename.as_ref(), source.as_str()).finish();
+
+                // If we can convert the total file size to u32, then
+                // we don't need to check when converting tree-sitter
+                // offsets (usize) into ruff offsets (u32)
+                if TryInto::<u32>::try_into(file.source_text().len()).is_err() {
```
ZedThree, line 148: Can probably do this without calling `.len()` by asking the filesystem

LiamPattinson (approved):
All looks good to me! Cheers for getting this in, it should make things like auto-fixes much easier to implement. Up to you if you'd like to make a few final tweaks before merging.

ZedThree:
Ok switched to using `Metadata.len()` so we don't need to actually open the file, and gave it a vaguely informative error message

LiamPattinson:
Nice, looks like a really clean solution. Happy to merge if you are.

PR #87, opened 2024-10-18
Add support for config files

### LINKED ISSUE #5: Project-wide settings
It should be possible to choose settings for a whole Git project via a top-level settings file -- perhaps `.fortitude.toml`? This should include the same fields as the optional CLI arguments. The following features will be required:

- [ ] Toml reader, or some other choice of settings file format.
- [ ] Function to merge command line settings with those from the settings file.
  - Command line arguments should always take precedence, but in some cases should be additive, such as when including rules.
- [ ] A function to find a `.fortitude.toml` file for a project.
  - Search up directories until finding one that contains `.fortitude.toml` or the `.git` directory.

(edit: 2024-09-19, updated bullet points that no longer make sense)

ZedThree:
Closes #5 

Adds support for `fortitude.toml` and `fpm.toml` files. Still needs tests and documentation.

`fortitude.toml` file:

```toml
global_options = "when we have any"

[check]
ignore = ["S001", "T041"]
line_length = 132
```

`fpm.toml`:

```toml
[extra.fortitude]
global_options = "when we have any"

[extra.fortitude.check]
ignore = ["S001", "T041"]
line_length = 132
```

Looks for settings files in the current directory, or any of its parents.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -117,8 +117,7 @@ fn from_clap_config_subsection<P: AsRef<Path>>(path: P) -> Result<Cli> {
     let matches = <Cli as CommandFactory>::command().get_matches();
 
     let config_str = if path.as_ref().ends_with("fpm.toml") {
-        let config = std::fs::read_to_string(path)?
-            .parse::<Table>()?;
+        let config = std::fs::read_to_string(path)?.parse::<Table>()?;
 
         // Unwrap should be ok here because we've already checked this
         // file as these tables
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,21 +1,29 @@
-use clap::{Parser, Subcommand};
-use std::path::PathBuf;
+use anyhow::{Context, Result};
+use clap::{CommandFactory, Parser, Subcommand};
+use clap_config::ClapConfig;
+use serde::{Deserialize, Serialize};
+use std::path::{Path, PathBuf};
+use toml::Table;
 
-#[derive(Debug, Parser)]
+#[derive(Debug, Parser, ClapConfig)]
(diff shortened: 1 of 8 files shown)


src/cli.rs
```diff
@@ -42,6 +50,93 @@ pub struct CheckArgs {
...
+/// Find the path to the `pyproject.toml` or `ruff.toml` file, if such a file
+/// exists. Adapated from ruff
+pub fn find_settings_toml<P: AsRef<Path>>(path: P) -> Result<Option<PathBuf>> {
+    for directory in path.as_ref().ancestors() {
+        if let Some(pyproject) = settings_toml(directory)? {
+            return Ok(Some(pyproject));
+        }
+    }
+    Ok(None)
+}
```
LiamPattinson, line 114: Comments and variable names need updating here

LiamPattinson (commented):
Generally looks good, and I'm surprised at how little extra code was needed to add this.

An issue is that the current behaviour is for the command line to always overwrite whatever is in the config file. While this is fine for things like `line_length` and `select`, it would be better for things like `include` and `ignore` if the config file and command line args could be combined in some way (or maybe `extend-ignore` and `extend-include` would be better on the command line?). Do you think it would be better to add that layer of complexity here, or in a later PR?

ZedThree:
I think this a cheap and cheerful solution for now that gets the job done. We'll definitely want things like `extend-ignore`, but that will need special handling which means we might end up abandoning this solution entirely.

ZedThree:
Done and done!

LiamPattinson (approved):

PR #96, opened 2024-10-31
Use ruff's diagnostics crate

ZedThree:
This is about 80/90% complete. I'd like to finish removing (the renamed) `Fortitude{Violation,Diagnostic}`. Pretty sure `FortitudeViolation` can now go entirely, and `FortitudeDiagnostic` will probably be replaced with a `Message` for output to different formats.

I think passing `Settings` into `Rule::check` will allow me to remove `Rule::new` completely, just not managed to thread that needle yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,17 +376,16 @@ checksum = "2b15c43186be67a4fd63bee50d0303afffcef381492ebe2c5d87f324e1b8815c"
 [[package]]
 name = "ruff_source_file"
 version = "0.0.0"
-source = "git+https://github.com/astral-sh/ruff.git?tag=0.6.8#ae39ce56c0cc1f8ac15f980c0b457b16b67c1f2a"
+source = "git+https://github.com/astral-sh/ruff.git?tag=0.7.1#337af836d3b2d88413c1391495a94755af46f574"
 dependencies = [
  "memchr",
- "once_cell",
  "ruff_text_size",
 ]
 
 [[package]]
 name = "ruff_text_size"
 version = "0.0.0"
-source = "git+https://github.com/astral-sh/ruff.git?tag=0.6.8#ae39ce56c0cc1f8ac15f980c0b457b16b67c1f2a"
+source = "git+https://github.com/astral-sh/ruff.git?tag=0.7.1#337af836d3b2d88413c1391495a94755af46f574"
 
 [[package]]
 name = "rustix"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+
+[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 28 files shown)


ZedThree:
Ok, think this is ready.

This change is not particularly interesting by itself, but it allows us to start:

- suggesting edits
- adding output formats

We probably want to switch to snapshot testing to simplify changing these things.

Also, we might want to rethink `Rules`. Currently we put instances into a map and look up codes, retrieving the instance and calling `.check()` on it. Instead, we might store the `check` function, and pass `Settings` into that.

LiamPattinson:
There's quite a lot to go over here, and I have a lot on right now, so I might not get through everything until sometime next week. From a first pass over it, I think it generally looks better, but I'd like to get a better handle on things before merging.

A very small request: Could `FromTSNode` be renamed `FromASTNode`? Just for consistency with `ASTRule` etc.

---

> Also, we might want to rethink `Rules`. Currently we put instances into a map and look up codes, retrieving the instance and calling `.check()` on it. Instead, we might store the `check` function, and pass `Settings` into that.

I was previously passing `Settings` directly into the `check` functions, but switched it to the current method during my last big rewrite. I honestly can't remember why -- maybe I was anticipating we would need to do more complex set-up down the line, and this would make things faster?

I'm also not a fan of the current system in which structs are both `Rule` and `Violation`, and some of the data is set to default values depending on which of the two `Self` needs to be in any given context. Removing `Rule::new()` and passing settings to the check function would help to simplify this a lot, but I'm not sure about removing the `Rule` traits altogether, as it's nice to have something to bind `entrypoints` to the rule type and enforce correct function signatures.

If we don't need to construct a `Rule` before checking it, perhaps we should replace `BTreeMap`'s of `{codes: instances}` with `BTreeSet`'s of just `{codes}`. We could then use macros to match the rule codes to the right function call, which would be defined on the rule types instead of the rule instances. For example, we could implement an `entrypoints` function to get the entry-points for each AST rule:

```rust
// Define in rules/macros, which will fill in the branches
fn entrypoints(rule_code: &str) -> Vec<&'static str> {
    match rule_code {
        ...
        "T001" => ImplicitTyping::entrypoints()
        ...
    }
}
```

I haven't actually tried implementing this approach, and knowing Rust there's probably some complicated issue with the type system or macro rules that would make this much more difficult than I'm expecting, but it might result in a cleaner solution if it works. I'd be happy to have a look at this next Thursday/Friday if you don't get there first.

ZedThree:
> There's quite a lot to go over here, and I have a lot on right now, so I might not get through everything until sometime next week. From a first pass over it, I think it generally looks better, but I'd like to get a better handle on things before merging. 

Yep, that's partly why I didn't want to start actually implementing new features in this PR, just switch the types over.

> A very small request: Could `FromTSNode` be renamed `FromASTNode`? Just for consistency with `ASTRule` etc.

Done.

> > Also, we might want to rethink `Rules`. Currently we put instances into a map and look up codes, retrieving the instance and calling `.check()` on it. Instead, we might store the `check` function, and pass `Settings` into that.
> 
> I was previously passing `Settings` directly into the `check` functions, but switched it to the current method during my last big rewrite. I honestly can't remember why -- maybe I was anticipating we would need to do more complex set-up down the line, and this would make things faster?

I think passing in the settings is more in line with ruff with their `checker` object, so that seems reasonable?

> I'm also not a fan of the current system in which structs are both `Rule` and `Violation`, and some of the data is set to default values depending on which of the two `Self` needs to be in any given context. Removing `Rule::new()` and passing settings to the check function would help to simplify this a lot, but I'm not sure about removing the `Rule` traits altogether, as it's nice to have something to bind `entrypoints` to the rule type and enforce correct function signatures.

Yep, agreed. It's nice to have our own traits to extend the functionality.

LiamPattinson (approved):

PR #97, opened 2024-11-01
Make `check` functions static

ZedThree:
This is a PR into #96

I think in rust parlance these are now "associated functions", but they're now the equivalent of C++'s static methods. This lets us do several things:

- call `check` without actually having to make an instance
- store the `check` functions in the rule lookup maps
- remove the `Rule` trait entirely

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,25 +108,18 @@ impl FromStartEndLineCol for Diagnostic {
 // Rule trait
 // ----------
 
-/// Implemented by all rules.
-pub trait Rule {
-    fn new(settings: &Settings) -> Self
-    where
-        Self: Sized;
-}
-
 /// Implemented by rules that act directly on the file path.
-pub trait PathRule: Rule {
+pub trait PathRule {
     fn check(settings: &Settings, path: &Path) -> Option<Diagnostic>;
 }
 
 /// Implemented by rules that analyse lines of code directly, using regex or otherwise.
-pub trait TextRule: Rule {
+pub trait TextRule {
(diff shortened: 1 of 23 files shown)


src/lib.rs
```diff
@@ -108,38 +108,32 @@ impl FromStartEndLineCol for Diagnostic {
...
     /// Return list of tree-sitter node types on which a rule should trigger.
-    fn entrypoints(&self) -> Vec<&'static str>;
+    fn entrypoints() -> Vec<&'static str>;
 
     /// Apply a rule over some text, generating all violations raised as a result.
-    fn apply(&self, source: &SourceFile) -> anyhow::Result<Vec<Diagnostic>> {
-        let entrypoints = self.entrypoints();
+    fn apply(source: &SourceFile) -> anyhow::Result<Vec<Diagnostic>> {
+        let entrypoints = Self::entrypoints();
+        let default_settings = default_settings();
```
LiamPattinson, line 131: I think it would be better if `apply` took settings as an argument. To avoid the need to repeatedly import `default_settings` throughout the tests, maybe we could add a convenience function of `apply_default` on `ASTRule`?

LiamPattinson (approved):
Nice, this is pretty much what I had in mind in https://github.com/PlasmaFAIR/fortitude/pull/96#issuecomment-2451996248. Just one little thing that I think could improve it, but otherwise all looks good.

src/lib.rs
ZedThree: Yes, that's a good idea. I'd like to rewrite the tests to use snapshots first as this potentially might change then anyway

Issue #95, opened 2024-10-31
Snapshot testing


ZedThree:
We can (somewhat) simplify all our testing by switching to `insta` and capturing snapshots: https://insta.rs/docs/cli/

This is how rust tests individual rules: https://github.com/astral-sh/ruff/blob/main/CONTRIBUTING.md#rule-testing-fixtures-and-snapshots

Related: #29

PR #99, opened 2024-11-08
Add fix for superfluous implicit none

LiamPattinson:
Adds a correction for `SuperfluousImplicitNone`, and prints the recommendation to the user. There are still a few things I want to get right before the merge:

- [x] The new `edit_delete` function on `Node` should automatically clean up whitespace around the deleted node.
- [x] Snippet bounds should be adjusted when printing corrections.

This also demonstrates why https://github.com/PlasmaFAIR/fortitude/issues/95 will soon be necessary: the modifications required of the tests to accommodate fixes will be far too unwieldy.


LiamPattinson:
Fix suggestions from `fix_title` are now printed, and a diff of the proposed changes is shown. I've used a simple coloured diff generator for now, but it doesn't match the style of `annotate_snippets`.

ZedThree (approved):

PR #102, opened 2024-11-10
Update dev README to provide correct commands for installing from source

### LINKED ISSUE #101: Outdated guidelines for building from source using cargo in README.dev.md
As per the title, the guidelines are currently:
```
cargo install --path .
```

When running from the root directory this results, for me, in:
```
error: found a virtual manifest at `<path>/fortitude/Cargo.toml` instead of a package manifest
```

I believe the solution is instead to run:
```
cargo install --path fortitude
```
as the project is split into two crates, `fortitude` and `fortitude-macros`, the latter being a dependency of the former.
If I had to guess I'd presume this is a hangover from before the macros crate existed which fell through the cracks.

I have a PR ready for you to review and will open shortly.

jatkinson1000:
Thanks again for this project.
I am looking at building for development, but on my way came across this issue in the developer guidelines, so thought I'd submit a small patch.

Hopefully fixes #101 

I have updated to point cargo to `fortitude` rather than the local directory.

I also added a clarification that this should be run from the root directory of the project.
This is because `cargo install --path .` works from `fortitude/fortitude/`, but based on the pip installation instructions I assumed both are to be run from the root directory.

### DIFF
--- a/README.dev.md
+++ b/README.dev.md
@@ -3,7 +3,8 @@
 ## Installation from source
 
 To install from source, you must first have a working Rust environment (see
-[rustup](https://rustup.rs/)). The project may then be installed either using `pip`:
+[rustup](https://rustup.rs/)). The project may then be installed from the root directory
+using either `pip`:
 
 ```bash
 python -m venv venv # Or use your preferred virtual environment method...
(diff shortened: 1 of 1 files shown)


LiamPattinson (approved):
Thanks for this, I hadn't realised that `cargo install` had stopped working. Everything else like `cargo build` and `cargo test` was working just fine, and it's not something we're checking in the CI tests. I'll raise an issue to update them.

PR #107, opened 2024-11-12
Add `Rule` enum

ZedThree:
Adapted from ruff's `map_codes` procedural macro, this creates a `Rule` enum with all of the different violations as variants. This lets us pass around `Rule` instances instead of code strings, and enables some future work (like stealing ruff's `RuleSelector` machinery).


LiamPattinson (approved):
I've had a look through and no issues immediately jumped out at me. I think it would take a long time for me to understand `map_codes` in full -- I came close to figuring out ruff's version a few times, but there were always some sections that didn't make sense to me. I'll probably have to figure it out piecemeal whenever I need to alter it.

It seems a shame to remove the macros I wrote so soon, but I was anticipating that we'd have to move to proc-macros at some point, and it's probably better to do it sooner than later.

PR #108, opened 2024-11-12
Enable more sophisticated rule selection

### LINKED ISSUE #33: Select/include/ignore rulesets with just category ID
Currently rules must be selected/included/ignored individually. It would be more convenient if it were possible to use categories instead:

```bash
fortitude check --ignore=S my_files/
```

ZedThree:
Adds `RuleSelector` from ruff to select or ignore whole groups (or subgroups) of rules:

```bash
# Just check for typing and module rules
fortitude check --select=T,M
# Ignore all styling rules
fortitude check --ignore=S
# Only check for typing rules, but ignore superfluous implicit none
fortitude check --select=T --ignore=T003
```

Closes #33


ZedThree:
Ah, the ruff equivalent is `--extend-select`, which does indeed allow you to add to the default ruleset.

For reference, here's the full list of ruff's rule selection settings:

```
Rule selection:
      --select <RULE_CODE>                                 Comma-separated list of rule codes to enable (or ALL, to enable all rules)
      --ignore <RULE_CODE>                                 Comma-separated list of rule codes to disable
      --extend-select <RULE_CODE>                          Like --select, but adds additional rule codes on top of those already specified
      --per-file-ignores <PER_FILE_IGNORES>                List of mappings from file pattern to code to exclude
      --extend-per-file-ignores <EXTEND_PER_FILE_IGNORES>  Like `--per-file-ignores`, but adds additional ignores on top of those already specified
      --fixable <RULE_CODE>                                List of rule codes to treat as eligible for fix. Only applicable when fix itself is enabled (e.g., via `--fix`)
      --unfixable <RULE_CODE>                              List of rule codes to treat as ineligible for fix. Only applicable when fix itself is enabled (e.g., via `--fix`)
      --extend-fixable <RULE_CODE>                         Like --fixable, but adds additional rule codes on top of those already specified
```

We probably want to have all of these eventually. Options, in ascending order of more work, are:

1. Just remove `--include` for now
2. Get `--include` working with new rule selection, possibly rename later
3. Do 2. but also rename to `--extend-select` here
4. Implement all the ruff selection settings here


Thoughts?

LiamPattinson:
Provided it doesn't add too much complexity, I think renaming `--include` to `--extend-select` and adding it here would be best. Per-file rule selection should probably wait until another PR, and per-rule fixability should probably wait until we have a decent number of fixes implemented.

ZedThree:
Will need to check, as `--extend-select` should allow the user to extend the selection from the config file. I suspect that is where all the complexity comes in!

LiamPattinson:
Yes, if it would require rewriting most of the clap code, then perhaps that's something for another PR, and deletion would make more sense at this stage. As the default ruleset is currently just the full ruleset, I suppose `--extend-select` won't do anything interesting until the CLI/settings file interface is upgraded.

ZedThree:
Ok, I've just removed `--include` for now. I'll make an issue for `extend-select`

ZedThree:
Actually, I think I now how to implement `extend-select`, will have a quick try now

LiamPattinson:
It's looking really good! I'm surprised how little extra machinery was needed to get `--extend-select` to work.

I was playing around with it and found that the `explain` command was no longer working, so I've raised [a small PR](https://github.com/PlasmaFAIR/fortitude/pull/110) to correct this (and made some additional improvements while I was at it, because it seems I can't help myself).

LiamPattinson (approved):

Issue #106, opened 2024-11-12
Human-friendly rule names


ZedThree:
[ruff has discussed this](/astral-sh/ruff/issues/1773) previously, see there for more details. Basically, we should allow e.g. `implicit-typing` as an alternative to `T001` for input, and display  both.

As we currently don't have suppression comments, adding this (hopefully) won't be a massive hardship, it should certainly be pretty easy for output.

Two comments to pay attention to though, in terms of implementation:

- [comment one](/astral-sh/ruff/issues/1773#issuecomment-2163951669)
- [comment two](/astral-sh/ruff/issues/1773#issuecomment-2164799403)

So then the design decision is how do register these names? Easiest option is to convert the violation name to snake/hyphen case, but that doesn't allow for alternatives. Maybe that's the primary name, and alternatives must be registered in the big registration macro

LiamPattinson:
Regarding ways to automatically generate different cases:

I tried playing around with [convert_case](https://docs.rs/convert_case/latest/convert_case/) and [Inflector](https://docs.rs/Inflector/latest/inflector/), both of which can convert from camel-case rule/violation names to kebab-case or title-case. Inflector will also do sentence-case.

The problem is that neither library understands the language, and each will simply draw boundaries where they find capital letters/spaces/hyphens/etc. They get pretty close, but there are some issues, e.g. converting `IOError` to sentence-case results in `Ioerror` instead of `IO error`. It looks like you can define custom patterns with `convert_case`, but I'm not sure if you can make something sophisticated enough to do everything we need automatically.

LiamPattinson:
I just double-checked, and convert_case correctly handles acronyms with kebab-case:

```rust
use convert_case::{Case, Casing};

assert_eq!("io-error", "IOError".to_case(Case::Kebab));
```

That's probably the easiest way to generate the human-readable identifier from the `Violation` name.

ZedThree:
Is this done? Or do we want to leave it open till we cover multiple aliases?

PR #110, opened 2024-11-13
Fix explain bug with RuleSelector, improve explain output

LiamPattinson:
While trying out https://github.com/PlasmaFAIR/fortitude/pull/108/commits/7a4b965983001ab387c96d121a0c4c231723afb1 in https://github.com/PlasmaFAIR/fortitude/pull/108, I found that the `explain` command was no longer working properly. This PR fixes the bug, and while I was at it I also improved the output for `explain`, updated the README, and improved the help text that's printed when things go wrong.

`explain` now prints the name of the rule alongside the code, but it's in camel-case. There are a few Rust libraries that allow you to convert to other cases, like [sentence case](https://docs.rs/Inflector/latest/inflector/cases/sentencecase/index.html), but these aren't clever enough to rename `IOError` to `IO error` or `NonStandardFileExtension` to `Non-standard file extension`, so for now I've left it at camel-case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,10 +53,10 @@ fortitude check --select=T --ignore=T003
 The `explain` command can be used to get extra information about any rules:
 
 ```bash
-fortitude explain T001
+fortitude explain --rules=T001,T011,...
 ```
 
-If no rules are provided, this will print all rule descriptions to the terminal.
+If `--rules` is not provided, this will print all rule descriptions to the terminal.
 
 To see further commands and optional arguments, try using `--help`:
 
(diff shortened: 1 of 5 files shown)


LiamPattinson:
I added a small function to the `map_codes` macro so you can get the name of a `Rule` from the enum, but this feels like something that should have already been there. Have I overlooked where this is defined?

ZedThree (approved):
Nice, thanks! I think this is a step towards #106, though maybe we eventually want to use `kebab-case` names rather than `CamelCase`

PR #111, opened 2024-11-14
Select rules by name

LiamPattinson:
Adds a basic implementation for https://github.com/PlasmaFAIR/fortitude/issues/106, so you can now specify the full names of rules:

```
fortitude check --select=implicit-typing
```

Also works in settings files.

Missing some features:

- Can't specify categories by full name, so `--select=typing` doesn't work.
- Human-readable names are mapped 1-to-1 with rule codes. In the future we'll likely need multiple human-readable names and multiple codes mapping to each rule.

Ultimately I think we'll need to add more macro machinery to do it properly, but right now I'm still having issues figuring out how everything fits together.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,6 +192,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "convert_case"
+version = "0.6.0"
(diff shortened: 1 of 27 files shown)


ZedThree (approved):
Amazing! :tada:

PR #112, opened 2024-11-14
Select categories by name

LiamPattinson:
Extends https://github.com/PlasmaFAIR/fortitude/pull/111, rule categories can also be selected by name, allowing things like:

```
fortitude check --select=typing,S --ignore=unnamed-end-statement,T001
```

I'd argue that this mostly resolves https://github.com/PlasmaFAIR/fortitude/issues/106, but we're still lacking the ability to map multiple aliases to one rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,19 +44,28 @@ You can select or ignore individual rules or whole groups with
 ```bash
 # Just check for missing `implicit none`
 fortitude check --select=T001
+# Also check for missing `implicit none` in interfaces
+fortitude check --select=T001,T002
 # Ignore all styling rules
 fortitude check --ignore=S
 # Only check for typing rules, but ignore superfluous implicit none
 fortitude check --select=T --ignore=T003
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -44,20 +44,29 @@ You can select or ignore individual rules or whole groups with
...
 ```
 
 The `explain` command can be used to get extra information about any rules:
 
 ```bash
-fortitude explain --rules=T001,T011,...
+# Print extra information for all rules
+fortitude explain
+# Only get information for selected rules
+fortitude explain --rules=T001,T011
```
ZedThree, line 63: Separate issue, but should this actually just be a positional argument?

LiamPattinson: It was before, but was changed when `RuleSelector` was brought in. I thought it was an intentional change so was just rolling with it :laughing:

ZedThree: Oops, that's entirely my bad! Fixed in #113

PR #113, opened 2024-11-14
Make `explain` arguments positional

ZedThree:
I think this is a bit easier to use


LiamPattinson (approved):
All looks good to me, which may be because I just pushed a near-identical branch to solve the same problem.

PR #116, opened 2024-11-14
Refactor: Use `strum_macros` to get rule and category names

LiamPattinson:
After looking into how ruff handles conversion to-and-from human-readable names for rules, I realised that most of the machinery needed to do so was already in the code: the line `#[strum(serialize_all="kebab-case")]` was already generating nice names for each rule, and the derive macro `strum_macros::AsRefStr` was adding `.to_ref()`, which converts the rule to `&'a str`. By adding the macro `strum_macros::EnumString` I was able to also gain the string-to-enum mapping, and `strum_macros::Display` adds `.to_string()`. I also added `strum_macros::IntoStaticStr` for the sake of completing the set.

This PR adds extra strum functionality and removes the extra functions added in https://github.com/PlasmaFAIR/fortitude/pull/111 and https://github.com/PlasmaFAIR/fortitude/pull/112, but leaves the functionality to select rules and categories by name on the command line and in settings files. It also cleans up a few imports.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,15 +192,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "convert_case"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec182b0ca2f35d8fc196cf3404988fd8b8c739a4d270ff118a398feb0cbec1ca"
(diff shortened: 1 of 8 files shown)


ZedThree (approved):
Nice! More stuff for less code is great

Issue #118, opened 2024-11-14
Errors in toml config file are silently ignored


MatthAlex:
Added the following, one after the other.

```bash
$ grep -B 1 -F "line-length" *.toml
fortitude.toml-[check]
fortitude.toml:line-length = 132
--
fpm.toml-[extra.fortitude.check]
fpm.toml:line-length = 132
```

```bash
(.venv) $ fortitude check app/
...
-- Number of errors: 22

(.venv) $ fortitude check app/ --line-length 132
...
-- Number of errors: 6
```

The command line option is respected, but the files are not.

ZedThree:
Thanks for the bug report @MatthAlex! It looks like we're expecting the setting in the config file to be `line_length` with an underscore, but we're not raising an error on unknown variables in the file.

We should probably use `line-length` for consistency with the CLI, and definitely error on unknown settings.

MatthAlex:
Yup, that works, confirmed!

Having it be `line-length` also matches the README example.

LiamPattinson:
Thanks @MatthAlex, I spotted this one myself in https://github.com/PlasmaFAIR/fortitude/issues/247 and patched it in https://github.com/PlasmaFAIR/fortitude/pull/251. The latest commit on main should include the fix, but we haven't released a patch version yet.

@ZedThree Do you think we should make a new release for v0.6.3 now? If so, we should probably also update the file counting issue so that the reported number of files scanned is correct even when the error diagnostics are switched off. I'm a bit concerned that we've made a few too many other changes to `main`, so pushing through to v0.7 might be preferable instead.

PR #119, opened 2024-11-15
Use kebab-case in config files

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/118 (though please confirm before merging!).

We're using a macro library https://github.com/gibfahn/clap_config to automatically generate the config file parser from the CLI args, but while Clap prefers kebab-case for CLI args, this library converts them to snake-case. This PR brings a modified version of that macro into Fortitude, and changes the serialisation to kebab-case. It's licensed under MIT, so as far as I know there shouldn't be any licensing issues. We could continue to build on this macro going forward, or may instead opt to use a more complicated setup that allows for different arguments in the config file versus the command line.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,9 +280,12 @@ dependencies = [
 name = "fortitude_macros"
 version = "0.1.0"
 dependencies = [
+ "clap",
+ "heck",
  "itertools 0.12.1",
  "proc-macro2",
  "quote",
+ "serde",
  "syn",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ authors = [
 license = "MIT"
 
 [workspace.dependencies]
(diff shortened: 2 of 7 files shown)


LiamPattinson:
Something I forgot to check before raising this PR: It still doesn't catch unknown args in the config file, so while `line-length` is now successfully picked up, `line_length` is now quietly ignored, which isn't ideal.

ZedThree (approved):

LiamPattinson:
Currently struggling to figure out why `#[serde(deny_unknown_fields)]` isn't working in this case, and I can't see where to add any checks to the macro.

LiamPattinson:
Another point against this solution: with `#[serde(deny_unknown_fields)]`, any error in the TOML file results in every field being ignored. I think now might be a good time to come up with a more robust solution!

PR #117, opened 2024-11-14
Add some basic integrated tests

ZedThree:
Closes #29 

TODO:

- [x] Test `check`
- [x] Test configuration file

PR #121, opened 2024-11-18
Convert unit tests to snapshot tests

### LINKED ISSUE #95: Snapshot testing

ZedThree:
Closes #95 

Builds on #117


ZedThree:
The `SuperfluousImplicitNone` test might have issues depending on how you run it, as the diff output may or may not have colours in it. I didn't see an easy way to disable the colour for the tests.

fortitude/src/lib.rs
```diff
@@ -263,6 +277,7 @@ fn format_violation(
         let fix_title = Level::Note.title(fix_msg);
         writeln!(f, "\n{}", renderer.render(fix_title))?;
 
+        // FIXME: turn off colour in tests?
```
LiamPattinson, line 280: I think this FIXME has already been fixed

README.dev.md
```diff
@@ -54,6 +51,42 @@ the rule is intended to fix. Words such as 'forbid' should be omitted. For examp
 name for the rule that warns of overly long lines is `LineTooLong`, and not something
 like `AvoidLineTooLong` or `KeepLinesShort`.
 
+#### Rule testing: fixtures and snapshots
+
+To test rules, Fortitude uses snapshots of Fortitude's output for a given file (fixture). Generally, there
+will be one file per rule (e.g., `E402.py`), and each file will contain all necessary examples of
```
LiamPattinson, line 57: References a `.py` file, but should be a `.f90` file

README.dev.md
```diff
@@ -54,6 +51,42 @@ the rule is intended to fix. Words such as 'forbid' should be omitted. For examp
...
+both violations and non-violations. `cargo insta review` will generate a snapshot file containing
+Fortitude's output for each fixture, which you can then commit alongside your changes.
+
+Once you've completed the code for the rule itself, you can define tests with the following steps:
+
+1. Add a Python file to `fortitude/resources/test/fixtures/[category]` that contains the code you
```
LiamPattinson, line 63: Similarly, should be a Fortran file. There are a few other Python references in this doc.

LiamPattinson (changes requested):
Nice one! I found a few very minor issues, but overall it's looking much cleaner than before.

LiamPattinson:
Since the diff output is more of a temporary placeholder at this stage, I'd be in favour of disabling it for now until we can write up a more robust method.

ZedThree:
I've just deleted the diff output for the time being, with the intention we'll restore it later

LiamPattinson (approved):

ZedThree:
Oh, the macos-12 runner is deprecated, so that won't work

PR #126, opened 2024-11-20
Update precision explanations

LiamPattinson:
Following feedback on [Fortran Discourse](https://fortran-lang.discourse.group/t/fortitude-a-fortran-linter/8834), updates the explanations given for 'precision' rules.

I based these rules on advice given on Fortran-Lang's [best practices pages](https://fortran-lang.org/en/learn/best_practices/floating_point/), so I'm strongly in favour of keeping them around, but it's clearly a controversial topic, so we may have to exclude these rules from the default set in a future version.

Also updated markdown code snippets to use `f90` over `fortran`, as `fortran` can cause incorrect syntax highlighting:

```fortran
! fortran
real, allocatable, intent(in) :: hello_world
```

```f90
! f90
real, allocatable, intent(in) :: hello_world
```

### DIFF
--- a/fortitude/src/rules/precision/double_precision.rs
+++ b/fortitude/src/rules/precision/double_precision.rs
@@ -24,12 +24,6 @@ use tree_sitter::Node;
 /// - `use, intrinsic :: iso_fortran_env, only: dp => real64`
 /// - `integer, parameter :: dp = selected_real_kind(15, 307)`
 ///
-/// To select a `real` with the same precision as `double precision`, the kind
-/// may be set from a floating point literal constant using `d` or `D` to denote
-/// the exponent:
-///
-/// - `integer, parameter :: dp = kind(0.0d0)
-///
 /// For code that should be compatible with C, you should instead use
 /// `real(c_double)`, which may be found in the intrinsic module `iso_c_binding`.
 ///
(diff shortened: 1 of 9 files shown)


fortitude/src/rules/precision/double_precision.rs
```diff
@@ -9,20 +9,34 @@ use tree_sitter::Node;
...
+/// `real(dp)`, with `dp` set in one of the following ways:
 ///
 /// - `use, intrinsic :: iso_fortran_env, only: dp => real64`
 /// - `integer, parameter :: dp = selected_real_kind(15, 307)`
 ///
+/// To select a `real` with the same precision as `double precision`, the kind
+/// may be set from a floating point literal constant using `d` or `D` to denote
+/// the exponent:
+///
+/// - `integer, parameter :: dp = kind(0.0d0)
```
ZedThree, line 31: This would conflict with #2?

LiamPattinson: It would, but that seems to be the advice given in Modern Fortran 2018. I imagine that line would be a strong case for the use of a `NOLINT` comment.

I'd also be happy to remove that section altogether, as recommending `selected_real_kind`, `iso_fortran_env` and `iso_c_bindings` is confusing enough without also offering a fourth option to achieve (probably) the same thing.

ZedThree (approved):

PR #129, opened 2024-11-20
Parallelise check

LiamPattinson:
Builds on https://github.com/PlasmaFAIR/fortitude/pull/127

Check works in parallel, which makes it marginally faster. For large projects, the IO step at the end can dominate.

Currently failing snapshot tests, and I'm working on the reasons why.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,6 +198,31 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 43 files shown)


LiamPattinson:
Latest changes to the output format branch seem to have fixed the issue.

ZedThree (approved):

PR #131, opened 2024-11-20
CI: Check installation

### LINKED ISSUE #103: Not checking installability in CI
https://github.com/PlasmaFAIR/fortitude/pull/102 fixed a documentation error which recommended an incorrect method for installing the software. To catch these issues in future, we should check that the following work in our CI tests:

```bash
pip install .
cargo install --path fortitude
```

Ideally these should also be followed by a check that calling `fortitude --help` actually works.

ZedThree:
Fixes #103

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,3 +22,27 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+
+  cargo-install:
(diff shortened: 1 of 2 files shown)


ZedThree:
Oh, those tests won't actually run on this PR because no rust files have changed :facepalm:

LiamPattinson (approved):

PR #137, opened 2024-11-21
Better parallel performance

LiamPattinson:
https://github.com/PlasmaFAIR/fortitude/pull/129 introduced parallel file checking, but the speed-up was underwhelming. This is because the parsing step was still single-threaded, as the tree sitter parser was hidden behind a mutex. This was done to permit the use of a simple `parse(file)` function, and originally there was no expectation that parallelisation would be needed.

Now builds a new parser for each file. I'd have expected this to degrade performance, but it makes no noticeable difference on a single thread, and greatly improves parallel performance. 

On my laptop, running the following command on [geos-chem](https://github.com/geoschem/geos-chem):

```bash
time fortitude check --output-format=concise geos-chem
```

takes about 5.8 seconds with `EXPORT RAYON_NUM_THREADS=1`, and just 2.4 second with `EXPORT RAYON_NUM_THREADS=4`. Before this PR, it was taking 4.7 seconds on four threads.

Running on GS2 with one thread takes me 1.3 seconds, and four threads takes just 0.65 seconds!

I think this update is significant enough to justify a version 5.1. Thoughts?


LiamPattinson:
As a further note, on my machine it can parse geos-chem in just 1.6 seconds and GS2 in just 0.35 seconds if the results are directed to a file instead of printed to the terminal.

fortitude/src/lib.rs
```diff
@@ -95,7 +96,14 @@ pub trait AstRule {
     /// Apply a rule over some text, generating all violations raised as a result.
     fn apply(source: &SourceFile) -> anyhow::Result<Vec<Diagnostic>> {
```
ZedThree, line 97: Oh, we should be able to drop this function entirely now?

fortitude/src/check.rs
```diff
@@ -245,7 +246,13 @@ pub(crate) fn check_file(
     }
 
     // Perform AST analysis
-    let tree = parse(file.source_text())?;
+    let mut parser = Parser::new();
+    parser
+        .set_language(&tree_sitter_fortran::LANGUAGE.into())
+        .context("Error loading Fortran grammar")?;
```
ZedThree, line 252:
```suggestion
    let mut parser = Parser::new()
        .set_language(&tree_sitter_fortran::LANGUAGE.into())
        .context("Error loading Fortran grammar")?;
```

ZedThree (approved):
We should probably add an integrated test where we check a whole project.

We might also be able to do something fancy like use a thread pool allocator to avoid making a new parser for every single file

fortitude/src/check.rs
LiamPattinson: Unfortunately you can't do this, as `set_language` returns `Result<(), LanguageError>`.

fortitude/src/lib.rs
LiamPattinson: It's still used in the test for `DeprecatedRelationalOperator`, but otherwise yes.

LiamPattinson: Just saw that you already raised a PR into this branch to fix it, so that's sorted now

LiamPattinson:
> We might also be able to do something fancy like use a thread pool allocator to avoid making a new parser for every single file

This would be ideal, but I was surprised at how little difference it makes to build a new parser each time.

PR #143, opened 2024-11-22
Clarify check help text for recursive file searches

LiamPattinson:


ZedThree (approved):
Thanks! We might need to look into getting clap to wrap long help text at some point

PR #145, opened 2024-11-22
Add fancy progress bar

LiamPattinson:
Since https://github.com/PlasmaFAIR/fortitude/pull/127, diagnostics are not printed as they are created, but instead are printed all at once after checking all files. On large projects, this can give the impression that the program is hanging, and gives no indication of when the program might be done.

This adds a fancy progress bar so the user can keep track of how things are going. Uses [indicatif](https://docs.rs/indicatif/latest/indicatif/), and the bar seems to nicely delete itself after completion.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -120,6 +120,7 @@ pub struct CheckSettings {
     pub line_length: usize,
     pub file_extensions: Vec<String>,
     pub output_format: OutputFormat,
+    pub no_progress_bar: bool,
 }
 
 /// Read either fpm.toml or fortitude.toml into our "known good" file
@@ -146,6 +147,7 @@ fn parse_config_file(config_file: &Option<PathBuf>) -> Result<CheckSettings> {
                 .file_extensions
                 .unwrap_or(FORTRAN_EXTS.iter().map(|ext| ext.to_string()).collect_vec()),
             output_format: value.output_format.unwrap_or_default(),
+            no_progress_bar: value.no_progress_bar.unwrap_or_default(),
         },
         None => CheckSettings::default(),
     };
@@ -368,6 +370,10 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
 
     let output_format = args.output_format.unwrap_or(file_settings.output_format);
 
+    let no_progress_bar = args
+        .no_progress_bar
+        .unwrap_or(file_settings.no_progress_bar);
+
     // At this point, we've assembled all our settings, and we're
     // ready to check the project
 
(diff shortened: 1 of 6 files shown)


ZedThree:
Neat! The contrast is not great in my terminal though, I just get a blue bar appear briefly. Could we adjust that to be clearer?

LiamPattinson:
Please could you check if this is better? Indicatif uses [console::Color](https://docs.rs/console/0.14.1/console/enum.Color.html) for styling, so at best you can provide an 8-bit colour (that's 8 bits for the whole colour, not 8 bits for each of red, green and blue!). I couldn't figure out what their 8-bit colour scheme was, but I came to this by trial and error.

LiamPattinson:
Currently looks like this in my terminal:

Screenshot from 2024-11-22 15-33-38

One issue is that it vanishes too quickly for most projects. It would be nice if there were an option to only show it if it takes longer than 0.5s and the ETA is over 1s (or something like that).

ZedThree:
Yep, that looks much better, thanks! Although the ETA doesn't work for me, just shows `[00:00]` even for large projects.

> One issue is that it vanishes too quickly for most projects. It would be nice if there were an option to only show it if it takes longer than 0.5s and the ETA is over 1s (or something like that).

Could only show it if we have more than, say, 100 files? We know that's only going to take less than a second.

Probably we should also have an option like `--no-progress-bar` to turn it off.

We might need to be a little careful about this when we get to the LSP server.

LiamPattinson:
>  Although the ETA doesn't work for me, just shows [00:00] even for large projects.

I experimented with it a bit more and its usefulness seems to vary. On large projects and a single thread it's quite accurate, but on 4 threads it just stays on 0s, maybe bumping up to 1s. I've reverted it for now.

> Could only show it if we have more than, say, 100 files? We know that's only going to take less than a second.

I'm not sure where to draw that line, because there could be 50 gigantic files in one project that take an age to parse. Plus people could be running this on really slow hardware.

> Probably we should also have an option like `--no-progress-bar` to turn it off.

Done :+1: 

> We might need to be a little careful about this when we get to the LSP server.

Currently I've set it to make a progress bar with a style of `""` under some circumstances. It's not technically switching it off, so that might cause some problems down the line. I think that's something we can deal with later though.

ZedThree (approved):

fortitude/src/check.rs
```diff
@@ -119,6 +120,7 @@ pub struct CheckSettings {
     pub line_length: usize,
     pub file_extensions: Vec<String>,
     pub output_format: OutputFormat,
+    pub no_progress_bar: bool,
```
ZedThree, line 123: If we want to be really rusty (rustic?), this should probably be a new enum:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, CacheKey, is_macro::Is)]
pub enum ProgressBar {
    #[default]
    Disabled,
    Enabled,
}

impl From<bool> for ProgressBar {
    fn from(version: bool) -> Self {
        if version {
            ProgressBar::Enabled
        } else {
            ProgressBar::Disabled
        }
    }
}

impl Display for ProgressBar {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::Disabled => write!(f, "disabled"),
            Self::Enabled => write!(f, "enabled"),
        }
    }
}
```

(just copy-pasting from `PreviewMode`)

This just gives type-safety if passing this around -- can't accidentally pass it to `PreviewMode` for example!

LiamPattinson: I've now done this, but went with three options: `Off`, `Ascii`, and `Fancy`.

LiamPattinson:
A further update: Added ASCII mode, turned the bar off by default, and used alternative unicode characters to achieve half-character resolution (You can get even better if you go for a maximum height bar, but I think the narrow bar looks sleeker).

I've looked into it, and the bar doesn't get printed at all if you're not writing to terminal, so hopefully this shouldn't interfere with things like LSP.

ZedThree (approved):

PR #150, opened 2024-11-27
Add rule for missing `private` statements in modules

### LINKED ISSUE #147: Warn about modules without `private` statements
I think it's a good idea to have a private statement in a module, requiring the programmer to explicitly list the public entities, so I request that fortitude have an option to warn about a module such as the one below.

```
module m
implicit none
integer :: i
end module m
```

If such a warning is added, whether it should be turned on by default is a matter of taste.

ZedThree:
Closes #147

Three questions I guess:

1. Should this be in the default option set?
2. Should this be in "preview" mode to begin with? (We need to expose that as an option!)
3. Should we allow explicitly setting the default visibility to `public`?


LiamPattinson:
Nice, looks good!

> Should this be in the default option set?

I'd argue 'yes', with the caveat that it should also allow `public` (see below)

> Should this be in "preview" mode to begin with? (We need to expose that as an option!)

This would be the ideal, but would it be a lot of work to set up? I imagine we would need to add a formal default rule set and a 'preview default set' containing a subset of the rules added since the last major release. We would then need to modify the default rule set when the user runs with `--preview`, and during each new (major) release we would need to migrate 'preview' defaults over to the regular default set.

> Should we allow explicitly setting the default visibility to public?

I would argue 'yes', but separate this into two rules to let the user choose how strict they want to be:

- `M021: MissingExportStatement` which requires the user to explicitly state `public` or `private`. I agree that `public` is usually bad, but if the user states it explicitly then they probably know what they're doing.
- `M022: PublicExports` which complains if the user is explicitly using `public`.

Maybe the former could be in the default set and the latter could be one the user has to manually activate?

ZedThree:
> This would be the ideal, but would it be a lot of work to set up? I imagine we would need to add a formal default rule set and a 'preview default set' containing a subset of the rules added since the last major release. We would then need to modify the default rule set when the user runs with --preview, and during each new (major) release we would need to migrate 'preview' defaults over to the regular default set.

We already have the machinery from ruff in place that filters out preview rules, it's just a matter of hooking it all up. We'd still want to review rules that are in the preview group on new releases though.

I'm just trying to get it all tested, will be a separate PR

fortitude/src/rules/modules/accessibility_statements.rs
```diff
@@ -0,0 +1,103 @@
...
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for missing `private` or `public` accessibility statements in modules
+///
+/// ## Why is this bad?
+/// The `private` statement makes all entities (variables, types, procedures)
+/// private by default, requiring an explicit `public` attribute to make them
+/// available. As well as improving encapsulation between modules, this also
+/// makes it possible to detect unused entities.
```
LiamPattinson, line 16:
```suggestion
/// makes it possible to detect unused entities.
///
/// A `public` statement in a module does not change the default behaviour,
/// and therefore all entities will be available from outside the module
/// unless they are individually given a `private` attribute. This brings
/// all of the same downsides as the default behaviour, but an explicit
/// `public` statement makes it clear that the programmer is choosing
/// this behaviour intentionally.  
```

fortitude/src/rules/modules/accessibility_statements.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+/// ## What it does
+/// Checks if the default accessibility in modules is set to `public`
+///
+/// ## Why is this bad?
+/// The `public` statement makes all entities (variables, types, procedures)
+/// public by default. This decreases encapsulation and makes it more likely to
+/// accidentally expose more than necessary.
```
LiamPattinson, line 69:
```suggestion
/// accidentally expose more than necessary. Public accessibility also makes
/// it harder to detect unused entities, which can often be indicative of
/// errors within the code.
```

LiamPattinson (approved):
All looks good! I've added a couple of suggestions for the rule explanations, but the rules themselves and the tests look fine.

LiamPattinson:
Now that we're using preview mode, should this be followed by a v0.5.1 release, or should we try to get a few more rules in first?

ZedThree:
> Now that we're using preview mode, should this be followed by a v0.5.1 release, or should we try to get a few more rules in first?

I think getting a few more rules/features in first is a good idea

I'm actually changing my mind on `M022`. Running this check over 160 projects, and it does seem that people are consciously using `public` on purpose, usually to mean "re-export all the imports", or on a module just of constants.

If we keep it, it definitely shouldn't be on by default.

@Beliavsky just wondering if you had an opinion on this?

LiamPattinson:
I don't think I would want `M022` on by default, but it would be nice to have. Once we have noqa/nolint/allow comments, or maybe `per-file-ignores` I think those cases you mentioned would be a good candidate for their use. We could also expand the check in `M022` to see if only constants have been defined within the module.

ZedThree:
Ok, I think we merge these but definitely remove M022 from the default set

Issue #148, opened 2024-11-25
Do not warn about missing kind suffix in argument of `kind` function


Beliavsky:
For the code

```
implicit none
integer, parameter :: sp = kind(1.0), dp = kind(1.0d0)
print*, sp, dp
end
```
`fortitude check xxkind.f90` gives

```
warning: xxkind.f90:2:33: P001 real literal 1.0 missing kind suffix
  |
1 | implicit none
2 | integer, parameter :: sp = kind(1.0), dp = kind(1.0d0)
  |                                 ^^^ P001
3 | print*, sp, dp
4 | end
  |
```
  
Setting kind parameters in this way is common and reasonable, so I suggest that fortitude not warn about a missing suffix in the argument of the `kind` function.

ZedThree:
Very good point, thanks!

Issue #35, opened 2024-09-16
Catch literal kinds in `real(x, kind=k)` and similar functions


LiamPattinson:
The rules `LiteralKind` and `LiteralKindSuffix` catch the use of number literals as kind specifiers for both variable declarations and on number literals themselves. It's still possible for them to slip through the net via type casting functions:

```f90
real(8) :: x        ! This will be caught
x = 2.0_8           ! This will be caught
x = real(2, 8)      ! This will not!
x = real(2, kind=8) ! Should be careful to catch this too
```

PR #156, opened 2024-11-28
Less aggressive warnings for `no-real-suffix`

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/148, and additionally doesn't warn for things like:

```f90
real(dp), parameter :: x = 0.0
```

However, it will still warn for:

```f90
real(dp), parameter :: x = 3.141592654
```

I'm not at all satisfied with my test for whether precision would be lost or not. Currently it's just parsing the literal into a 32-bit float and a 64-bit float and seeing if they're equal, but this can be unreliable. I tried making a more clever test which tests the relative error between the two against `f32::EPSILON`, but this was even less reliable. Any suggestions would be appreciated!

### DIFF
--- a/fortitude/resources/test/fixtures/precision/P001.f90
+++ b/fortitude/resources/test/fixtures/precision/P001.f90
@@ -1,15 +1,24 @@
 program test
-  use, intrinsic :: iso_fortran_env, only: sp => real32, dp => real64
+  use, intrinsic :: iso_fortran_env, dp => real64
 
-  real(sp), parameter :: x1 = 1.234567
-  real(dp), parameter :: x2 = 1.234567_dp
-  real(dp), parameter :: x3 = 1.789d3 ! rule should ignore d exponentiation
-  real(dp), parameter :: x4 = 9.876
-  real(sp), parameter :: x5 = 2.468_sp
-  real(sp), parameter :: x6 = 2.
-  real(sp), parameter :: x7 = .0
-  real(sp), parameter :: x8 = 1E2
-  real(sp), parameter :: x9 = .1e2
-  real(sp), parameter :: y1 = 1.E2
-  real(sp), parameter :: y2 = 1.2e3
+  integer, parameter :: sp = kind(0.0) ! Okay: Permissible in a kind statement
+
+  real(dp), parameter :: a = 0.0 ! Okay: No loss of precision
+  real(sp), parameter :: c = 1.0e10 ! Okay: No loss of precision, e exponent
+  real(sp), parameter :: d = 2.0E10 ! Okay: No loss of precision, E exponent
+  real(sp), parameter :: e = 1.23456 ! Bad: Loss of precision
+  real(sp), parameter :: f = 1.23456e1 ! Bad: Loss of precision, e exponent
+  real(sp), parameter :: g = 1.23456E1 ! Bad: Loss of precision, E exponent
+  real(dp), parameter :: h = 1.23456_dp ! Okay: Kind suffix
+  real(sp), parameter :: i = 1.23456_sp ! Okay: Loss of precision, but we're explicit
+  real(dp), parameter :: j = 1.23456d1 ! Okay: Ignore d exponent
+  real(dp), parameter :: k = 1.23456D3 ! Okay: Ignore D exponent
+
+  real(dp) :: p, q, x, y, z
+
+  x = sqrt(2.0) ! Bad: Loss of precision
+  y = real(1.0, kind=dp) ! Okay: Type cast with no loss of precision
+  z = real(1.0 + 1.0, kind=dp) ! Bad: Type cast from expression, possible l.o.p
+  p = real(5.0, kind=dp) ! Okay: Type cast with no loss of precision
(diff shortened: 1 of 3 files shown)


fortitude/src/rules/precision/kind_suffixes.rs
```diff
@@ -72,6 +79,46 @@ impl AstRule for NoRealSuffix {
         if !regex_is_match!(r"^(\d*\.\d*|\d*\.*\d*[eE]\d+)$", txt) {
             return None;
         }
+
+        // Ok if being used in a kind statement
+        if let Some(call_expression) = node.ancestors().find(|x| x.kind() == "call_expression") {
```
ZedThree, line 84: I am slightly worried this is a little too general, as we want exactly the grand-parent (or great-grand-parent for complex literals?) -- but I just searched my whole corpus and the most complicated I saw was stuff like `kind(thing(1))`, so probably this is fine in practice

fortitude/resources/test/fixtures/precision/P001.f90
```diff
@@ -5,15 +5,19 @@ program test
...
-  real(sp), parameter :: e = 1.23456 ! Bad: Loss of precision
-  real(sp), parameter :: f = 1.23456e1 ! Bad: Loss of precision, e exponent
-  real(sp), parameter :: g = 1.23456E1 ! Bad: Loss of precision, E exponent
-  real(dp), parameter :: h = 1.23456_dp ! Okay: Kind suffix
-  real(sp), parameter :: i = 1.23456_sp ! Okay: Loss of precision, but we're explicit
-  real(dp), parameter :: j = 1.23456d1 ! Okay: Ignore d exponent
-  real(dp), parameter :: k = 1.23456D3 ! Okay: Ignore D exponent
+  real(dp), parameter :: c = 0.25 ! Okay: No loss of precision
+  real(dp), parameter :: d = 0.000244140625 ! Okay: No loss of precision (=2^-12)
+  real(dp), parameter :: e = 17745.0 ! Okay: No loss of precision
```
ZedThree, line 10: Same as `b`?

ZedThree (approved):
I guess this is one where we could really do with more information to check if there will actually be loss of precision or not. See gfortran for example: https://godbolt.org/z/TPaWG8P6r Removing one digit from the end of the literal no longer triggers the warnings

fortitude/src/rules/precision/kind_suffixes.rs
LiamPattinson: Further down I use `parent()?.parent()?` to do just this, as the first parent is the argument list and the second is the function call. I could use that here too -- I did it this way here in case people used it in cases where the return kind of an expression isn't obvious, like where a C++ programmer might use `decltype`. I suppose in that case, maybe explicit suffixes would be a good idea anyway.

fortitude/resources/test/fixtures/precision/P001.f90
LiamPattinson: Thanks, I think the suggestion might have gotten a little confused, but I removed the repetition in one of my later commits.

fortitude/src/rules/precision/kind_suffixes.rs
LiamPattinson: I've tightened the search through the ancestors now so the following code will still raise a warning:

```f90
integer, parameter :: sp = kind(1.0 + 1.0)
```

fortitude/resources/test/fixtures/precision/P001.f90
```diff
@@ -1,15 +1,28 @@
...
-  real(sp), parameter :: x5 = 2.468_sp
-  real(sp), parameter :: x6 = 2.
-  real(sp), parameter :: x7 = .0
-  real(sp), parameter :: x8 = 1E2
-  real(sp), parameter :: x9 = .1e2
-  real(sp), parameter :: y1 = 1.E2
-  real(sp), parameter :: y2 = 1.2e3
+  use, intrinsic :: iso_fortran_env, dp => real64
+
+  integer, parameter :: sp = kind(0.0) ! Okay: Permissible in a kind statement
```
ZedThree, line 4: Might also want to check `kind((0.0, 0.0))`

LiamPattinson: Good point, I forgot about complex...

LiamPattinson: On further investigation it also fails for negative numbers!

LiamPattinson: Fixed now, and I caught another bug where I'd gotten the type casting functions `cmplx` and `int` wrong (I'd called them `complex` and `integer`)

ZedThree (approved):
~Thi~

LiamPattinson:
I think I've caught most of the edge cases here... Also scanned through the old issues and realised this fixes https://github.com/PlasmaFAIR/fortitude/issues/35

ZedThree (approved):

PR #164, opened 2024-11-29
Move statement-function rule to preview set

LiamPattinson:
Was a little hasty in merging https://github.com/PlasmaFAIR/fortitude/pull/162! New rules should probably all go in the preview set from now on.


ZedThree:
I think this one is probably uncontroversial enough that it could go straight into stable, but happy to have it go into preview

LiamPattinson:
Oh yeah, I totally agree that this should be in the default set, and the PR that bumps the version to 0.6 should probably move it over -- I'm just trying to build good habits now so that I'm less likely to upset users down the line.

ZedThree (approved):

Issue #24, opened 2024-09-13
Prefer modules to common blocks


LiamPattinson:
There are a few edge cases in modern Fortan where common blocks should be preferred as their contents are guaranteed to be ordered in memory the same way they're declared, but generally speaking modules do the same job but better.

PR #165, opened 2024-11-29
Add rule `common-block`

LiamPattinson:
A nice easy one.

Fixes https://github.com/PlasmaFAIR/fortitude/issues/24

### DIFF
--- a/fortitude/src/rules/obsolescent/common_blocks.rs
+++ b/fortitude/src/rules/obsolescent/common_blocks.rs
@@ -59,7 +59,7 @@ use tree_sitter::Node;
 ///   i = 1
 /// end subroutine s
 /// ```
-//.
+///
 /// ## References
 /// - Metcalf, M., Reid, J. and Cohen, M., 2018, _Modern Fortran Explained:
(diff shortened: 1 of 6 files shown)


fortitude/src/rules/obsolescent/common_blocks.rs
```diff
@@ -0,0 +1,85 @@
...
+/// subroutine s()
+///   use mydata, only: i, j
+///   i = 1
+/// end subroutine s
+/// ```
+//.
+/// ## References
+/// - Metcalf, M., Reid, J. and Cohen, M., 2018, _Modern Fortran Explained:
+///   Incorporating Fortran 2018_, Oxford University Press, Appendix B
+///   'Obsolescent and Deleted Features'
```
ZedThree, line 66: Should probably copy this reference for the statement functions one too

ZedThree (approved):

LiamPattinson:
I was trying to also add a rule for `block data`, but I can't even figure out how to implement them, so probably not worth worrying about at this stage!

ZedThree:
Easy way to find an example for things like that: look in the tree-sitter-fortran test suite! There will be a short example in there, including the AST

PR #167, opened 2024-11-29
Bump tree-sitter-fortran

ZedThree:
Several new features, bug fixes, and further support for preprocessor
`#if`s in places


fortitude/Cargo.toml
```diff
@@ -46,7 +46,7 @@ textwrap = "0.16.0"
 thiserror = { version = "1.0.58" }
 toml = "0.8.19"
 tree-sitter = "~0.23.0"
```
LiamPattinson, line 48:
```suggestion
tree-sitter = "~0.24.0"
```

LiamPattinson (approved):
Should we update our `tree-sitter` version while we're at it?

PR #169, opened 2024-11-29
Add rule for obsolescent `entry` statement

ZedThree:
Includes #167 as it requires the updated grammar

### DIFF
--- a/fortitude/src/rules/obsolescent/entry_statement.rs
+++ b/fortitude/src/rules/obsolescent/entry_statement.rs
@@ -1,0 +1,42 @@
+use crate::settings::Settings;
+use crate::{AstRule, FromAstNode};
+use ruff_diagnostics::{Diagnostic, Violation};
+use ruff_macros::{derive_message_formats, violation};
+use ruff_source_file::SourceFile;
+use tree_sitter::Node;
+
+/// ## What it does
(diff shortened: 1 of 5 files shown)


fortitude/src/rules/obsolescent/entry_statement.rs
```diff
@@ -0,0 +1,56 @@
...
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for statement functions.
+///
+/// ## Why is this bad?
+/// Statement functions are an obsolescent feature from Fortran 77, and have been
+/// entirely supplanted by internal procedures. Statement functions are much more
+/// limited in what they can do. They were declared obsolescent in Fortran 90 and
+/// removed in Fortran 95.
```
LiamPattinson, line 15: Uses description text from `statement-function`

LiamPattinson (changes requested):
The rule and the tests look good, just the doc-comment needs updating.

fortitude/src/rules/obsolescent/entry_statement.rs
ZedThree: :facepalm: thanks!

I'm pretty sure ruff has some utility to add new rules -- we should steal that!

LiamPattinson (approved):

PR #171, opened 2024-11-29
Add rule for missing double-colon in variable decls

ZedThree:
Bad:

``` f90
real foo, bar
```

Good:

``` f90
real :: foo, bar
```

I _think_ this is a safe, always available fix.


LiamPattinson (approved):
Looks good!

Issue #120, opened 2024-11-18
Git commit in `--version`


ZedThree:
Looks like this can pretty easily be done with [shadow-rs](https://docs.rs/shadow-rs/latest/shadow_rs/)

PR #174, opened 2024-12-02
Print git commit and extra info with `--version`

LiamPattinson:
Resolves https://github.com/PlasmaFAIR/fortitude/issues/120

`shadow-rs` brings in a surprising amount of new dependencies, but it was easy enough to get working.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -46,7 +46,7 @@ jobs:
           sccache: 'true'
           manylinux: auto
         env:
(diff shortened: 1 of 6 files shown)


LiamPattinson:
That last step seems to have fixed my build problems but I have no idea why

ZedThree (approved):

LiamPattinson:
Whoops, this didn't actually pass the tests. It's just that I mis-formatted the YAML file and the failing tests didn't run.

Issue #175, opened 2024-12-02
Failing CI build and release


LiamPattinson:
PR https://github.com/PlasmaFAIR/fortitude/pull/174 introduced some issues with the build/release CI. Although I thought those issues were fixed within the PR, they're still unsolved. It's something to do with libgit2 not building correctly on manylinux containers.

PR #176, opened 2024-12-02
Fix failing build/release CI action

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/175


LiamPattinson:
That seems to have done it. It turns out that the default libgit2 dependency isn't needed after all, so this also cuts down on the dependency bloat.

ZedThree (approved):

PR #177, opened 2024-12-02
Add (many) more output formats

### LINKED ISSUE #172: Output to SARIF format
The SARIF format (https://sarifweb.azurewebsites.net/) is a standard schema for JSON that is written and read by many different static analysis  tools (https://github.com/oasis-tcs/sarif-spec/wiki/Known-Producers-and-Consumers#known-producers-of-sarif). For instance, it is written by the GCC and Clang static analyzers, and GitHub CodeQL, and it can be read by IDE extensions (e.g. [VSCode](https://marketplace.visualstudio.com/items?itemName=MS-SarifVSCode.sarif-viewer)) and also git forges (e.g., [GitHub Security](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning), [Azure DevOps](https://docs.kics.io/latest/integrations_azurepipelines/), and there is a request for integration into [GitLab](https://gitlab.com/gitlab-org/gitlab/-/issues/452042)). My use-case would primarily be the git forge integration, so that I can eventually include fortitude in a pipeline and have it scan PRs for issues.

It appears there is a Rust library that wrap the structure already - [serde-sarif](https://docs.rs/serde-sarif/latest/serde_sarif/index.html), so there shouldn't be a need to actually write the JSON formatter for this if we can include this dependency.

ZedThree:
Closes #172

Adds the remaining ruff output formats:

- json-lines
- junit
- grouped
- github
- gitlab
- pylint
- rdjson
- azure
- sarif

Also adds diff output of fixes to the "full" format for tests -- this
lets us actually check the fixes in the tests.


ZedThree:
These were basically all copy-pasted from ruff with some fairly mechanical changes

fortitude/src/message/rdjson.rs
```diff
@@ -0,0 +1,138 @@
...
+        &mut self,
+        writer: &mut dyn Write,
+        messages: &[DiagnosticMessage],
+    ) -> anyhow::Result<()> {
+        serde_json::to_writer_pretty(
+            writer,
+            &json!({
+                "source": {
+                    "name": "ruff",
+                    "url": "https://docs.astral.sh/ruff",
```
LiamPattinson, line 29: Left some ruff bits here
```suggestion
                    "name": "fortitude",
                    "url": "https://github.com/PlasmaFAIR/fortitude",
```

LiamPattinson (commented):
Found one error, otherwise looks good.

ZedThree:
I think a new clippy version has just been released or something, we're now getting warnings about bits I've not touched. I'll fix them here as they're very simple

LiamPattinson (approved):

PR #181, opened 2024-12-05
Add automatic fixes

### LINKED ISSUE #70: Auto-fixes
Rules should give suggestions for fixes where possible, with a mode to automatically apply them

ZedThree:
Closes #70

TODO: add docs!


fortitude/src/cli.rs
```diff
@@ -97,16 +97,45 @@ pub struct CheckArgs {
...
+    /// Apply fixes to resolve lint violations, but don't report on, or exit non-zero for, leftover violations. Implies `--fix`.
+    /// Use `--no-fix-only` to disable or `--unsafe-fixes` to include unsafe fixes.
+    #[arg(long, overrides_with("no_fix_only"), action = SetTrue)]
+    pub fix_only: Option<bool>,
+    #[clap(long, overrides_with("fix_only"), hide = true, action = SetTrue)]
+    pub no_fix_only: Option<bool>,
+    /// Avoid writing any fixed files back; instead, output a diff for each changed file to stdout, and exit 0 if there are no diffs.
+    /// Implies `--fix-only`.
+    #[arg(long, conflicts_with = "show_fixes", action = SetTrue)]
+    pub diff: Option<bool>,
```
ZedThree, line 127: This isn't actually implemented yet, maybe remove for now?

fortitude/src/fix/snippet.rs
```diff
@@ -0,0 +1,45 @@
+#![allow(dead_code)]
+// Taken from ruff
+// Copyright 2022 Charles Marsh
+// SPDX-License-Identifier: MIT
+
+use unicode_width::UnicodeWidthStr;
+
+/// A snippet of source code for user-facing display, as in a diagnostic.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SourceCodeSnippet(String);
```
ZedThree, line 10: This isn't used anywhere yet, but potentially useful -- instead of storing a `String` in a `Violation`, store one of these, and it will be automatically truncated if it's too long.

fortitude/src/fix/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+        assert!(source_map.markers().is_empty());
+    }
+
+    #[test]
+    fn apply_one_insertion() {
+        let locator = Locator::new(
+            r#"
+import os
+
+print("hello world")
```
LiamPattinson, line 188: Could the tests be converted to something similar in Fortran?

fortitude/src/printer.rs
```diff
@@ -78,66 +155,171 @@ impl Printer {
     pub(crate) fn write_once(
         &self,
         num_files: usize,
-        diagnostics: &[DiagnosticMessage],
+        diagnostics: &Diagnostics,
         writer: &mut dyn Write,
     ) -> Result<()> {
         // TODO: implement tracking of fixables
```
LiamPattinson, line 161: I think this TODO has been resolved

LiamPattinson (commented):
There's quite a lot of new machinery here, so it's hard to understand everything in full, but it mostly looks good with just a few issues I spotted.

fortitude/src/printer.rs
LiamPattinson: The suggestion is pointing at the wrong lines. It was meant to be the TODO at the start of this function.

fortitude/src/cli.rs
LiamPattinson: Sure, there's already quite a lot of new features to get to grips with, so removing it will reduce the complexity a little.

fortitude/src/check.rs
```diff
@@ -245,14 +268,85 @@ fn get_files<S: AsRef<str>>(files_in: &Vec<PathBuf>, extensions: &[S]) -> Vec<Pa
...
+            path_rules,
+            text_rules,
+            ast_entrypoints,
+            path,
+            file,
+            settings,
+        )?;
+        let fixed = FxHashMap::default();
+        (result, fixed)
+    };
```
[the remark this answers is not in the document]
ZedThree: We have this structure (in psuedo code):

```
if should_fix {
  try_fix
  if try_fix_worked {
     apply_fixes
  } else {
     just_lint
  }
} else {
  just_lint
}

return result
```

The first call to `check_only_file` here is a fall back if we tried to fix things but it didn't work for whatever reason.

The second call is if the user never requested fixes.

I think it could be rearranged, but we still end up with some repeated code:

```
if should_fix {
  try_fix
  if try_fix_worked {
     return result
  }
}

just_lint

return result
```

Ruff can't do this because it does some caching of the results before returning. If we intend to do that too, we'd have to switch back to the current structure

fortitude/src/fix/mod.rs
ZedThree: Yes, but these are actually all language agnostic, just working on `String`s. The magic numbers would need updating too, so I'm not sure it's worth the effort to be honest!

fortitude/src/check.rs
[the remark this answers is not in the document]
LiamPattinson: Okay, that's fair. I think the problem looked bigger than it actually was because rustfmt had put everything on its own line.

LiamPattinson (approved):

PR #182, opened 2024-12-05
Add fix for `deprecated-relational-operators`

ZedThree:
Plus a helper function to make future fixes even easier


LiamPattinson (approved):
Looks good!

PR #185, opened 2024-12-05
Add fix for `unnamed-end-statement`

LiamPattinson:
Builds on https://github.com/PlasmaFAIR/fortitude/pull/182

### DIFF
--- a/fortitude/src/rules/style/snapshots/fortitude__rules__style__tests__unnamed-end-statement_S061.f90.snap
+++ b/fortitude/src/rules/style/snapshots/fortitude__rules__style__tests__unnamed-end-statement_S061.f90.snap
@@ -3,7 +3,7 @@ source: fortitude/src/rules/style/mod.rs
 expression: diagnostics
 snapshot_kind: text
 ---
-./resources/test/fixtures/style/S061.f90:5:3: S061 end statement should read 'end type mytype'
+./resources/test/fixtures/style/S061.f90:5:3: S061 [*] end statement should be named.
   |
 3 |   type mytype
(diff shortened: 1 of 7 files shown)


fortitude/src/rules/style/snapshots/fortitude__rules__style__tests__unnamed-end-statement_S061.f90.snap
```diff
@@ -12,8 +12,19 @@ snapshot_kind: text
...
   |
+  = help: Write as 'end type mytype'.
 
-./resources/test/fixtures/style/S061.f90:9:3: S061 end statement should read 'end subroutine mysub1'
+ℹ Safe fix
+2 2 |   implicit none
+3 3 |   type mytype
+4 4 |     integer :: x
+5   |-  end type                      ! catch this
+  5 |+  end type mytype! catch this
```
ZedThree, line 22: Hmm, seems the node range includes all this whitespace, that's a little annoying.

We probably want a rule, `missing-whitespace-before-comment`, and that would fix this on a subsequent iteration

ZedThree (approved):

fortitude/src/rules/style/snapshots/fortitude__rules__style__tests__unnamed-end-statement_S061.f90.snap
LiamPattinson: I was just thinking the same, though I imagine that would fall under the regex rules, and I'm not sure how we would implement fixes for those. Is this fixable at the tree-sitter level?

PR #187, opened 2024-12-05
Add fix for `trailing-whitespace`

LiamPattinson:
I spent far too long figuring this out from scratch before it occurred to me that Ruff implements the exact same rule and I could just steal most of it. To my credit, I came up with basically the same solution, and barely had to change anything after realising. For example, I had this to get the number of bytes to remove:

```rust
let whitespace_bytes = line
    .chars()
    .rev()
    .take_while(|c| c.is_whitespace())
    .map(|c| c.len_utf8())
    .sum();
```

It's eerily similar to the solution in Ruff, which just switches out `usize` for `TextSize`.

The Ruff version isn't always a safe fix, as trailing whitespace in multiline strings needs to be preserved. Am I correct in thinking this will never be a problem in Fortran, as a line continuation character is always needed?

### DIFF
--- a/fortitude/src/rules/style/whitespace.rs
+++ b/fortitude/src/rules/style/whitespace.rs
@@ -5,7 +5,7 @@ use ruff_source_file::{OneIndexed, SourceFile};
 use ruff_text_size::{TextLen, TextRange, TextSize};
 
 use crate::settings::Settings;
-use crate::{FromStartEndLineCol, TextRule};
+use crate::TextRule;
 
 /// ## What does it do?
 /// Checks for tailing whitespace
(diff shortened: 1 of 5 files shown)


LiamPattinson:
In this comment I was wondering how we could implement a rule that ensures there's a space before a comment: https://github.com/PlasmaFAIR/fortitude/pull/185#discussion_r1871868166. This shows that fixes can be quite easily applied by a `TextRule`!

fortitude/src/rules/style/whitespace.rs
```diff
@@ -16,25 +17,44 @@ use crate::{FromStartEndLineCol, TextRule};
...
+                    line_end_byte,
                 ));
+                violations.push(
+                    Diagnostic::from_start_end_line_col(
+                        Self {},
+                        source_file,
+                        idx,
+                        line.trim_end().len(),
+                        idx,
+                        line.len(),
```
ZedThree, line 54: Should we now have:

```suggestion
                        line_end_byte - whitespace_bytes,
                        idx,
                        line_end_byte,
```

ZedThree (approved):

fortitude/src/rules/style/whitespace.rs
LiamPattinson: No, because `line_end_byte` is a bytes offset into the whole file, while `Diagnostic::from_start_end_line_col` uses zero-indexed rows and cols. However, what we can do is just use Ruff's `Diagnostic::new(kind, text_range)` directly, as we already have the `TextRange` here.

ZedThree: Ah, nice!

ZedThree (approved):

PR #192, opened 2024-12-06
Add rule for implicit external procedures

### LINKED ISSUE #140: Recommend `implicit none (type, external)`
This is a Fortran 2018 feature for preventing implicit procedures. The big compilers all support it: https://godbolt.org/z/hn36GjEqh
Notably `nvfortran` doesn't, so this would be an issue for CUDA projects.

Use case is something like

```f90
program foo
  implicit none

  call unknown_func() ! This compiles! (but may not link)
end program foo
```

```f90
program foo
  implicit none (external)

  call unknown_func() ! This won't compile
end program foo
```

This probably wants to be a separate rule to `implicit-typing` because not all compilers support it

ZedThree:
Closes #140

Includes #191 for important bug fix

### DIFF
--- a/fortitude/resources/test/fixtures/typing/T004.f90
+++ b/fortitude/resources/test/fixtures/typing/T004.f90
@@ -1,0 +1,11 @@
+module a
+  implicit none
+end module a
+
+module b
+  implicit none (type)
+end module b
+
+module c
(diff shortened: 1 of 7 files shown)


LiamPattinson (approved):
All looks good! :+1:

Issue #189, opened 2024-12-06
Whitespace before inline comments


LiamPattinson:
Other language style guides recommend two spaces before inline comments:

```f90
x = x + 1! bad
x = x + 1 ! better
x = x + 1  ! good
```

This more clearly distinguishes code from comments, improving readability.

This would also address an issue introduced by PR https://github.com/PlasmaFAIR/fortitude/pull/185, which introduced automatic fixes for unnamed end statements. It was found that if they're followed by an inline comment these fixes also remove any whitespace between the end statement and the comment, significantly hurting readability (ultimately I think this is a tree-sitter problem).

LiamPattinson:
Ideally this sort of thing should be fixed by a `fortitude format` call, but as it's possible for rule fixes to introduce errors like this, I think it would be best to duplicate the functionality into the linter too.

ZedThree:
Ah, the tree-sitter node is up to the start of the comment because most statements include a grammar rule `_end_of_statement`. This is either:

- semicolon
- end of line
- start of comment

This is a bit of a fiddly thing, unfortunately.

PR #194, opened 2024-12-06
Add rule for incorrect whitespace before comment

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/189

We could probably go further and mandate a maximum of 2 spaces too, but this could mess up deliberately lined up comments:

```f90
integer :: a    ! comment
integer :: bb   ! comment
integer :: ccc  ! comment
```


fortitude/src/rules/mod.rs
```diff
@@ -82,6 +82,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Style, "061") => (RuleGroup::Stable, Ast, style::end_statements::UnnamedEndStatement),
         (Style, "071") => (RuleGroup::Preview, Ast, style::double_colon_in_decl::MissingDoubleColon),
         (Style, "101") => (RuleGroup::Stable, Text, style::whitespace::TrailingWhitespace),
+        (Style, "121") => (RuleGroup::Preview, Ast, style::whitespace::IncorrectSpaceBeforeComment),
```
ZedThree, line 85: We're not very consistent with our rule codes, but should this be 111?

fortitude/resources/test/fixtures/style/S121.f90
```diff
@@ -0,0 +1,8 @@
+! This is fine
+ ! This isn't but we'll let it pass
```
ZedThree, line 2: Yeah, catching this will rely on know the indentation of the following node(s)

ZedThree (approved):

fortitude/src/rules/mod.rs
LiamPattinson: I'm not sure. I honestly just took a guess that there might be 20-ish whitespace rules we might want to add before this one. Maybe just 102 would make more sense?

ZedThree: Yeah, I better there's a few whitespace rules! Let's for for 102 then

LiamPattinson: Fixed in https://github.com/PlasmaFAIR/fortitude/pull/194/commits/ad6455db1917dc15fc752f00757a1ea9e83a8f73

PR #195, opened 2024-12-06
Add check for `external` procedures

### LINKED ISSUE #23: Prefer `interface` to `external`
Explicit interface declarations are far less error prone than `EXTERNAL` declarations:

```f90
interface
  function f(foo, bar)
    implicit none
    integer, intent(in) :: foo
    real, intent(in):: bar
    
    ...

  end function f
end interface

! Alternatively...
external :: f ! No args specified

x = f(foo, bar)
```

ZedThree:
Closes #23

There's a sort of naming clash with `external-function` -- I've called
this `external-statement`, but I think it's maybe better as
`external-procedure`. I wonder if that might be better as something
like `procedure-not-in-module`?

### DIFF
--- a/fortitude/resources/test/fixtures/typing/T061.f90
+++ b/fortitude/resources/test/fixtures/typing/T061.f90
@@ -1,0 +1,14 @@
+program test
+  implicit none (type, external)
+
+  interface
+    integer function f(foo, bar)
+      implicit none
+      integer, intent(in) :: foo
+      integer, intent(in):: bar
+    end function f
+  end interface
+
+  external :: g
+
+end program test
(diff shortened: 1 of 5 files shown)


LiamPattinson (approved):
:+1:

LiamPattinson:
> There's a sort of naming clash with external-function -- I've called
this external-statement, but I think it's maybe better as
external-procedure. I wonder if that might be better as something
like procedure-not-in-module?

I think `procedure-not-in-module` and `external-procedure` would be the best names, but it'll be hard to avoid some confusion since they're both arguing against the same practice but from different angles. We could set up some redirects for this, but I feel like having both `external-function` and `external-procedure` pointing at different rules might be too confusing, so I think a breaking change may be in order. Maybe raise an issue for this?

Issue #197, opened 2024-12-06
Bad slicing of utf-8 in rendering snippets


ZedThree:
Checking this file: https://github.com/grimme-lab/crest/blob/master/src/legacy_algos/reactor.f90
causes a panic:

```
crest/src/legacy_algos/reactor.f90:380:27: S101 [*] trailing whitespace
thread 'main' panicked at /home/peter/.cargo/git/checkouts/annotate-snippets-rs-7b89a4815579cf66/3e1ae5f/src/renderer/display_list.rs:1441:29:
byte index 26 is not a char boundary; it is inside '³' (bytes 25..27) of `  !-- transform into g/cm³   `
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::str::slice_error_fail_rt
   3: core::str::slice_error_fail
   4: alloc::vec::Vec<T,A>::retain::{{closure}}
   5: annotate_snippets::renderer::display_list::format_snippet
   6: annotate_snippets::renderer::display_list::format_message
   7: annotate_snippets::renderer::Renderer::render
   8: <fortitude::message::text::MessageCodeFrame as core::fmt::Display>::fmt
   9: core::fmt::write
  10: <&std::io::stdio::Stdout as std::io::Write>::write_fmt
  11: <std::io::stdio::Stdout as std::io::Write>::write_fmt
  12: <fortitude::message::text::TextEmitter as fortitude::message::Emitter>::emit
  13: fortitude::printer::Printer::write_once
  14: fortitude::check::check
  15: fortitude::main
```

relevant snippet is the first line here:

```f90
 380   │   !-- transform into g/cm³   
 381   │   dens = dens * ( 0.001d0 / (1.0d-30*bohr**3.0d0))
 382   │   !dens=0.001*mass/(1.0d-30*ax(1)*ax(2)*ax(3)*bohr**3)
```

not actually sure if it's the rendering or the rule that's the problem

LiamPattinson:
It looks like other people have encountered the same problem, but apparently there's an unreleased fix in the latest `annotate-snippets`: https://github.com/rust-lang/annotate-snippets-rs/issues/161. Since we started using our own fork, should we expect to already have that fix?

ZedThree:
Yep, we should have that fix already I think

PR #198, opened 2024-12-06
Bugfix: utf-8 multi-byte characters breaking diagnostic output

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/197

Although I need to add a test to make sure...

### DIFF
--- a/fortitude/src/lib.rs
+++ b/fortitude/src/lib.rs
@@ -19,7 +19,7 @@ pub use crate::registry::clap_completion::RuleParser;
 pub use crate::rule_selector::clap_completion::RuleSelectorParser;
 
 use ruff_diagnostics::{Diagnostic, DiagnosticKind};
-use ruff_source_file::{OneIndexed, SourceFile};
+use ruff_source_file::SourceFile;
 use ruff_text_size::{TextRange, TextSize};
 use settings::Settings;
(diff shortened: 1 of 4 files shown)


ZedThree:
Yeah, can you use the comment from the issue in a test? Or check with an emoji or something?

ZedThree:
Weird that ruff doesn't need this here :shrug:

LiamPattinson:
As Ruff is using the old version of `annotate-snippets`, maybe that worked on char ranges instead of byte ranges? It's hard to tell from the docs, as neither the old docs nor the new docs specify which they're using -- both just say it's a slice of `usize`.

ZedThree:
Here's another example that fails (adjusting `S001.f90`):

```f90
program test
  use some_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_really_long_module_name, only : integer_working_precision
  implicit none
  integer(integer_working_precision), parameter, dimension(1) :: a = [1]
  write (*, '("╔════════════════════════════════════════════╗")')
  write (*, '("║  UTF-8 LOGO BOX                            ║")')
  write (*, '("╚════════════════════════════════════════════╝")')

end program test
```

PR #203, opened 2024-12-06
Add sales-pitch to README

LiamPattinson:
Gives new visitors to the page a reason to keep reading. Hopefully doesn't come across too strong -- I'm definitely open to suggestions on this!


ZedThree (approved):
Nice!

PR #217, opened 2024-12-10
Fix link to rules table in Readme

imciner2:
The current link doesn't work because it needs to have the version specifier in the URL.


ZedThree (approved):
Lovely, thanks @imciner2!

PR #227, opened 2024-12-11
Catch syntax errors introduced by fixes

ZedThree:


LiamPattinson:
Fantastic! This would have saved us from that bug yesterday.

Would it be possible to add a test for this? I imagine we would need to add a rule to the test set that deliberately breaks something (maybe just the `unnamed-end-statement` rule pre-bugfix?)

ZedThree:
We currently don't have a good (or any!) mechanism of applying test rules, and I couldn't see how ruff tests their implementation of this, so I'm not quite how to actually add the test.

We might need to do something like pass the full list of selected rules deeper into the `check` functions, rather than splitting out the path/text/AST rules early on. We might need to do this anyway for the per-file rule selection.

LiamPattinson:
> We currently don't have a good (or any!) mechanism of applying test rules, and I couldn't see how ruff tests their implementation of this, so I'm not quite how to actually add the test.

That's fair, I don't think it's worth rewriting half of the code to get this in. I was expecting Ruff would have something we could easily copy over, but if not, I'm happy to go ahead with things as they are.

LiamPattinson (approved):

Issue #205, opened 2024-12-07
Open statements should have an `action` specifier


Beliavsky:
For code such as

```fortran
implicit none
open (unit=10, file="foo.txt")
write (10, *) "hello"
end
```
I suggest that fortitude complain that an `action="write"` was not specified in the open statement. In general, `open` statements should have an `action` specifier.

PR #230, opened 2024-12-12
Add rule `missing-action-specifier` and `Io` rule category

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/205

Other entries to this category would follow the advice in https://fortran-lang.org/en/learn/best_practices/file_io/, e.g. catch `action=read` without `status=old` (Edit: or maybe just catch missing `status` in general)


ZedThree (approved):

LiamPattinson:
I've tried running this rule on some real-world code and it doesn't seem to be catching everything. I'll have to investigate what's going wrong.

LiamPattinson:
> I've tried running this rule on some real-world code and it doesn't seem to be catching everything. I'll have to investigate what's going wrong.

Never mind, the rule works fine, I was just forgetting to run with `--preview` switched on :man_facepalming:

---

> I'm not completely sold on this particular rule to be honest. In some ways it's good to be explicit, but I do think in general it's good to know the defaults and to not repeat them

I'm the other way -- I lean towards always being explicit for file IO. I even write `open("myfile", "r")` in Python.

I think we should probably implement a 'default' and 'pedantic' set soonish, as there are now quite a few controversial rules.

Issue #6, opened 2024-01-22
Per-file ignores


LiamPattinson:
It should be possible to ignore rules on a per-file and per-line basis.

- To ignore rules per file, will need to add project-wide settings file first (https://github.com/PlasmaFAIR/fortitude/issues/5). It will be easier to handle this via a settings file or the command line than to parse comments in the files themselves.
- ~~Per-line ignores could be done the same way as flake8/ruff: search for a comment `! noqa` at the end of a line. The regex `!\w*noqa\w*$` should do it.~~
  - ~~After gathering rule violations for a file, check the returned line numbers, and search for `! noqa` on those lines. Discard violations wherever it's found.~~
  - ~~Could also have fine-grained ignores, such as `! ignore: E123`.~~
  - ~~Things will get trickier if we also aim to account for line continuations with `&`.~~

Edit: To aid project management, this Issue should just be for the per-file ignores, as these features can be tackled independently

jatkinson1000:
Adding a +1 for the option to ignore rules on a file-by-file basis.

LiamPattinson:
This is next up on our priorities list, and we've been giving the per-line ignores some more thought.

- We're thinking about going for something like `! allow(all)` as a general 'nolint' statement, and e.g. `! allow(line-too-long)` for specific allowances.
- This should go on the line _before_ the offending line(s). We probably won't have an inline equivalent in the first instance.
- To implement this, we should find all comments that match the required format, and if a violation is raised within the range of the next AST sibling, it should be discarded.
  - This will allow you to put the comment at the top of a function/module to ignore all issues within it!

PR #232, opened 2024-12-12
Add per-file-ignores

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/6

Allows the user to specify `--per-file-ignores` and `--extend-per-file-ignores` to switch off rules/categories for specific files:

```bash
fortitude check --per-file-ignores=myfile.f90:typing,myotherfile.f90:S001
```

Using `--extend-per-file-ignores` on the command line adds to the list in the config file, while `--per-file-ignores` overwrites it.

As we're just using `serde` to handle the config file vs CLI problem, the `.toml` interface is a bit clunky:

```toml
[check]
per-file-ignores = [
  "myfile.f90:typing",
  "myotherfile.f90:S001",
]
```

Ideally we would have a separate interface like:

```toml
[check.per-file-ignores]
"myfile.f90" = ["typing"]
"myotherfile.f90" = ["S001"] 
```

I'd suggest raising a new issue for this if this PR is merged.

---

I had to copy quite a lot of new Ruff machinery over to get this working, and while I've tried it out on some local files, I haven't written any tests yet. I'll leave this as a draft until those are done.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -660,18 +660,19 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
         extend_select: args.extend_select.unwrap_or(file_settings.extend_select),
     };
 
-    let mut per_file_ignores = args
-        .per_file_ignores
-        .or(file_settings.per_file_ignores)
-        .unwrap_or_default();
-    per_file_ignores.extend(
-        args.extend_per_file_ignores
+    let per_file_ignores = CompiledPerFileIgnoreList::resolve(collect_per_file_ignores(
+        args.per_file_ignores
+            .or(file_settings.per_file_ignores)
             .unwrap_or_default()
             .into_iter()
-            .chain(file_settings.extend_per_file_ignores),
-    );
-    let per_file_ignores =
-        CompiledPerFileIgnoreList::resolve(collect_per_file_ignores(per_file_ignores))?;
+            .chain(
+                args.extend_per_file_ignores
+                    .unwrap_or_default()
+                    .into_iter()
+                    .chain(file_settings.extend_per_file_ignores),
+            )
+            .collect::<Vec<_>>(),
+    ))?;
 
     let output_format = args.output_format.unwrap_or(file_settings.output_format);
     let preview_mode = resolve_bool_arg(args.preview, args.no_preview)
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -6,11 +6,14 @@ use crate::fs;
 use crate::message::DiagnosticMessage;
 use crate::printer::{Flags as PrinterFlags, Printer};
 use crate::registry::AsRule;
-use crate::rule_selector::{PreviewOptions, RuleSelector, Specificity};
+use crate::rule_selector::{
+    collect_per_file_ignores, CompiledPerFileIgnoreList, PreviewOptions, RuleSelector, Specificity,
+};
 use crate::rules::Rule;
 use crate::rules::{error::ioerror::IoError, AstRuleEnum, PathRuleEnum, TextRuleEnum};
 use crate::settings::{
-    FixMode, OutputFormat, PreviewMode, ProgressBar, Settings, UnsafeFixes, DEFAULT_SELECTORS,
+    FixMode, OutputFormat, PatternPrefixPair, PreviewMode, ProgressBar, Settings, UnsafeFixes,
+    DEFAULT_SELECTORS,
 };
 
 use anyhow::{Context, Result};
@@ -22,7 +25,7 @@ use ruff_diagnostics::Diagnostic;
 use ruff_source_file::{Locator, SourceFile, SourceFileBuilder};
 use ruff_text_size::TextRange;
 use rustc_hash::FxHashMap;
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fs::File;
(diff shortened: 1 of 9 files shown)


fortitude/src/check.rs
```diff
@@ -640,6 +660,19 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
         extend_select: args.extend_select.unwrap_or(file_settings.extend_select),
     };
 
+    let mut per_file_ignores = args
+        .per_file_ignores
+        .or(file_settings.per_file_ignores)
+        .unwrap_or_default();
+    per_file_ignores.extend(
```
ZedThree, line 667: Does this not work?

```suggestion
    let per_file_ignores = args
        .per_file_ignores
        .or(file_settings.per_file_ignores)
        .unwrap_or_default()
        .extend(
```

ZedThree:
Sounds like we might need to have the split types for CLI and config file sooner rather than later.

My first thought on doing this was to filter the ruleset for each file beforehand (in the loop, but before we check), but this way looks pretty expedient and clean.

fortitude/src/check.rs
LiamPattinson: I couldn't get that to work, but there was a similar iterator trick to avoid the mutable intermediate

danielhollas:
Hi folks, 

this is a missing feature that's currently blocking us from using Fortitude in CI (ideally we'd like to just exclude all rules for certain files). Feel free to ping me once things start converging and I am happy to take this PR for a spin on our codebase.

Thanks for all the work!

LiamPattinson:
@danielhollas I'm glad to hear that you're finding this useful, and I hope that by our next release we'll have enough configuration options for the most common use-cases. We're aiming to also add per-line rule ignores, the ability to exclude certain file patterns, and the ability to follow rules in the user's `.gitignore` files.

> ideally we'd like to just exclude all rules for certain files

I've just tested this, and the following seems to work:

`fortitude check --per-file-ignores=myfile.f90:ALL src/`

The file pattern can be finicky however. It will match all files with a certain file name (including with wildcards), but I can't get it to match files within a certain directory, which is something that you can do with `ruff`:

`fortitude check --per-file-ignores=./src/myfile.f90:ALL src/`  # Doesn't work for some reason

@ZedThree Would you be in favour of merging this now and raising a new issue for the directory-selection bug, or should I get that fixed first?

ZedThree:
I think that needs to be fixed, `src/myfile.f90` should work, and `myfile.f90` should not match `src/myfile.f90` -- `**/myfile.f90` should match both `myfile.f90` and `src/myfile.f90`

LiamPattinson:
I think I've gotten the directory selection to work, but it's strange in the tests -- it seems we need to specify `/tmp/.tmpDSFGJ725/` (or whatever) in the path to get `**/some_dir/*.f90` to work:

```
fortitude check --per-file-ignores=/tmp/.tmpSFGJ525/**/some_dir/*.f90 /tmp/.tmpSFGJ525
```

Outside of the tests however, the following seems to work just fine:

```
fortitude check --per-file-ignores=**/some_dir/*.f90 src
```

---

> I think that needs to be fixed, src/myfile.f90 should work, and myfile.f90 should not match src/myfile.f90 -- **/myfile.f90 should match both myfile.f90 and src/myfile.f90

Ruff's behaviour is that `myfile.f90` _should_ match `src/myfile.f90`. I could try to get this implemented, but it might take me a little while to work out exactly how the matching works and modify it accordingly.

ZedThree:
> Ruff's behaviour is that `myfile.f90` _should_ match `src/myfile.f90`. I could try to get this implemented, but it might take me a little while to work out exactly how the matching works and modify it accordingly.

Ahh, this is [`.gitignore`-like](https://git-scm.com/docs/gitignore) behaviour:

> - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.
> 
> - If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.
> 
> - For example, a pattern doc/frotz/ matches doc/frotz directory, but not a/doc/frotz directory; however frotz/ matches frotz and a/frotz that is a directory (all paths are relative from the .gitignore file).

fortitude/tests/check.rs
```diff
@@ -647,3 +647,216 @@ end program test
...
+    // Expect:
+    // - Don't overwrite config file
+    // - File types of foo and baz but no bar
+    // - No files with index 2
+    assert_cmd_snapshot!(Command::cargo_bin(BIN_NAME)?
+                         .args(["--config-file", config_file.as_os_str().to_string_lossy().as_ref()])
+                         .arg("check")
+                         .arg("--select=typing")
+                         .arg(ignores)
+                         .arg(path),
```
ZedThree, line 813: Change the directory the command is run in:

```suggestion
    let ignores = format!(
        "--extend-per-file-ignores=**/double_nested/*.f90:implicit-typing",
    );

    apply_common_filters!();
    // Expect:
    // - Don't overwrite config file
    // - File types of foo and baz but no bar
    // - No files with index 2
    assert_cmd_snapshot!(Command::cargo_bin(BIN_NAME)?
                         .args(["--config-file", config_file.as_os_str().to_string_lossy().as_ref()])
                         .arg("check")
                         .arg("--select=typing")
                         .arg(ignores)
                         .current_dir(path),
```

LiamPattinson: Thanks, that seems to have done it! I've done this and made some further simplifications in https://github.com/PlasmaFAIR/fortitude/pull/232/commits/960ea592c805799f1ddafaac01f9dab78ea5e8f2

PR #234, opened 2024-12-12
T042: Fix example code in assumed-size-character-intent rule

### LINKED ISSUE #229: T042: Example in the rule explanation does not compile
Hey :wave: 

First of all, thank you so much for giving us fortitude, it is awesome and great to see a modern linter for Fortran. As someone currently working on legacy Fortran codebases it's going to be quite useful!

The following program is part of `fortitude explain T042`

```f90
program example
  character(len=3) :: short_text
  call set_text(short_text)
  print*, short_text
contains
  subroutine set_text(text)
    character(len=:), allocatable, intent(out) :: text
    text = \"longer than 3 characters\"
  end subroutine set_text
end program
```
(Also note the superfluous backslashed before the quotes)

Trying to compile it with Gfortran 13.3.1 gives me this error

```
❯ gfortran example.f90
example.f90:3:16:

    3 |   call set_text(short_text)
      |                1
Error: Actual argument at (1) to allocatable or pointer dummy argument ‘text’ must have a deferred length type parameter if and only if the dummy has one
```

Making `short_text` allocatable seems to fix the issue and produces the correct output.

```f90
program fixed_example
  character(len=:), allocatable :: short_text
  call set_text(short_text)
  print*, short_text
contains
  subroutine set_text(text)
    character(len=:), allocatable, intent(out) :: text
    text = "longer than 3 characters"
  end subroutine set_text
end program
```
```console
❯ ./fixed_example 
 longer than 3 characters
```

Happy to submit a PR for this if the fix suggested above looks correct.
I'll freely admit that I am still confused about how exactly Fortran handles strings
so I was basically trying to use Fortitude to tell me what to do. :smile:

danielhollas:
See #229 for details.

Closes #229

### DIFF
--- a/docs/rules/assumed-size-character-intent.md
+++ b/docs/rules/assumed-size-character-intent.md
@@ -14,24 +14,24 @@ program example
 contains
   subroutine set_text(text)
     character(*), intent(out) :: text
-    text = \"longer than 3 characters\"
+    text = "longer than 3 characters"
   end subroutine set_text
 end program
 ```
 
-Here, `short_text` will only contain the truncated \"lon\".
+Here, `short_text` will only contain the truncated "lon".
 
 To handle dynamically setting `character` sizes, use `allocatable` instead:
 
 ```f90
 program example
-  character(len=3) :: short_text
-  call set_text(short_text)
-  print*, short_text
+  character(len=:), allocatable :: allocatable_text
+  call set_text(allocatable_text)
+  print*, allocatable_text
 contains
   subroutine set_text(text)
     character(len=:), allocatable, intent(out) :: text
-    text = \"longer than 3 characters\"
+    text = "longer than 3 characters"
   end subroutine set_text
 end program
 ```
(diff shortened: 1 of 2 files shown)


docs/rules/assumed-size-character-intent.md
```diff
@@ -14,24 +14,24 @@ program example
...
+Here, `short_text` will only contain the truncated "lon".
 
 To handle dynamically setting `character` sizes, use `allocatable` instead:
 
 ```f90
 program example
-  character(len=3) :: short_text
-  call set_text(short_text)
-  print*, short_text
+  character(len=:), allocatable :: allocatable_text
```
danielhollas, line 28: I am not sure what would be a good name for this variable.

danielhollas:
Generated docs look good and the example now compiles and runs correctly AFAICT

https://fortitude--234.org.readthedocs.build/en/234/rules/assumed-size-character-intent/

```console
❯ gfortran example.f90 
❯ ./a.out 
 longer than 3 characters
```

docs/rules/assumed-size-character-intent.md
ZedThree: I think this is fine

ZedThree (approved):
Lovely, thanks @danielhollas!

Issue #235, opened 2024-12-13
CLI and config file should be decoupled


LiamPattinson:
PR https://github.com/PlasmaFAIR/fortitude/pull/232 introduces per-file-ignores, which have the following syntax on the command line:

```bash
--per-file-ignores=<pattern>:<ignore>
# e.g.
--per-file-ignores=**/some_dir/*.f90:typing,**/some_dir/*.f90:style
```

As the config file is read using `serde` and the same structs used by `clap` to define the CLI, the same syntax is needed in the config files:

```toml
[check]
per-file-ignores = [
  "**/some_dir/*.f90:typing",
  "**/some_dir/*.f90:style",
]
```

A better syntax, and one familiar to Ruff users, would be:

```toml
[check.per-file-ignores]
"**/some_dir/*.f90" = ["typing", "style"]
```

As more configuration options are added there will likely be more instances where the best practices for the CLI and config files will diverge, and so we should implement a separate set of structs for the config files.

PR #236, opened 2024-12-13
Add rule `magic-number-in-array-size`

### LINKED ISSUE #180: Warn about repeated literals, especially integers used to declare arrays
For the code
```fortran
implicit none
real(8) :: probs(221), cnorm(221), beta(4,221), crits(221), wght(221)
end
```
fortitude does not complain about the repeated `221`, which should be a named constant. The code was taken from

https://github.com/cran/urca/blob/master/src/UnitRootMacKinnon.f

ZedThree:
Closes #180 

This should expose a setting `check.magic-number-in-array-size.ignored-values` -- which I think really requires #235


LiamPattinson (approved):
All looks good!

Issue #233, opened 2024-12-12
Exclude files, directories, and patterns


ZedThree:
We should have a setting to completely exclude files, directories, and patterns. Related to #231 

These options from ruff:

```rust
    pub exclude: Option<Vec<FilePattern>>,
    pub extend_exclude: Vec<FilePattern>,
    pub extend_include: Vec<FilePattern>,
    pub force_exclude: Option<bool>,
    pub include: Option<Vec<FilePattern>>,
    pub respect_gitignore: Option<bool>,
```

Issue #231, opened 2024-12-12
Fortitude should respect `.gitignore`


LiamPattinson:
Ruff has the options `--respect-gitignore` (active by default) and `--no-respect-gitignore`. If running under the latter, `ruff check .` at the top level of a git repo will start looking into folders like `build/`, `target/`, etc (it seems to avoid searching under `venv`/`.venv` unless you specifically request them).

Fortitude should behave similarly.

LiamPattinson:
A complication to this is that there may be nested `.gitignore` files, so we'll have to add to the set of excluded file patterns as we recurse through directories. It'll probably be best to see how Ruff handles this.

ZedThree:
We probably want to use the [ignore](https://crates.io/crates/ignore) crate.

PR #238, opened 2024-12-16
Add options to exclude files

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/233

Opened as a draft as I've yet to add tests.

Doesn't add all features suggested in https://github.com/PlasmaFAIR/fortitude/issues/233:
- `--respect-gitignore` is a separate issue (https://github.com/PlasmaFAIR/fortitude/issues/231).
- `--include` isn't an option for the Ruff CLI, and we'll need to handle config files separately for that (https://github.com/PlasmaFAIR/fortitude/issues/235)
- `--force-exclude` would be a small extension to this, and could be added along with the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,11 @@ fortitude check
 ```
 
 You can also call `check` on individual files, globs, and
-directories. You can configure what extensions `fortitude` searches
+directories. You can configure what extensions Fortitude searches
 for in directories with `--file-extensions`:
 
 ```bash
-fortitude check src --file-extensions=f90,fpp
+fortitude check --file-extensions=f90,fpp
 ```
 
 You can select or ignore individual rules or whole groups with
@@ -77,7 +77,7 @@ fortitude check --select=typing --ignore=superfluous-implicit-none
 Use `--output-format=concise` to get shorter output:
 
 ```bash
-$ fortitude check --concise
+$ fortitude check --output-format=concise
 test.f90:2:1: M001 function not contained within (sub)module or program
 test.f90:5:1: S061 end statement should read 'end function double'
 test.f90:7:1: M001 subroutine not contained within (sub)module or program
@@ -90,11 +90,11 @@ The `explain` command can be used to get extra information about any rules:
 # Print extra information for all rules
 fortitude explain
 # Only get information for selected rules
-fortitude explain T001,T011
+fortitude explain T001 T011
 # Print information on all style rules
 fortitude explain S
 # Rules and categories can also be referred to by name
-fortitude explain style,superfluous-implicit-none
+fortitude explain style superfluous-implicit-none
 ```
 
 To see further commands and optional arguments, try using `--help`:
@@ -152,35 +152,7 @@ preview = true
 
 Run `fortitude explain` to see which rules are in preview mode.
 
-### Excluding Files
 
-> [!NOTE]
-> To be featured in v0.7.0
-
-Fortitude will automatically ignore files in some directories (`build/`, `.git/`,
-`.venv/`, etc.), and this behaviour can be extended using the `--exclude` option. For
-example, to ignore all files in the directory `benchmarks/`:
-
-```console
-$ fortitude check --exclude=benchmarks
-```
-
-Note that this will override any exclusions specified in your `fpm.toml` or
-`fortitude.toml` file. To extend those exclusions, use instead:
-
-```console
-$ fortitude check --extend-exclude=benchmarks
-```
-
-It is also possible to switch off individual rules or rule categories for specific
-files using `--per-file-ignores`:
-
-```console
-$ fortitude check --per-file-ignores=**/*.f95:non-standard-file-extension
-```
-
-Similarly, if these are set in the configuration file, `--extend-per-file-ignores` can
-be used to add additional rule ignores instead of overwriting them.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -152,6 +152,36 @@ preview = true
 
 Run `fortitude explain` to see which rules are in preview mode.
 
+### Excluding Files
+
+> [!NOTE]
+> To be featured in v0.7.0
+
+Fortitude will automatically ignore files in some directories (`build/`, `.git/`,
+`.venv/`, etc.), and this behaviour can be extended using the `--exclude` option. For
+example, to ignore all files in the directory `benchmarks/`:
+
+```console
+$ fortitude check --exclude=benchmarks
+```
+
+Note that this will override any exclusions specified in your `fpm.toml` or
+`fortitude.toml` file. To extend those exclusions, use instead:
+
+```console
+$ fortitude check --extend-exclude=benchmarks
+```
+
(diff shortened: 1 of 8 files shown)


ZedThree:
I think `--include` is essentially the positional arguments, basically `files`?

`--force-exclude` probably good to get in too, it sounds like it's useful for things like precommit

fortitude/tests/check.rs
```diff
@@ -69,7 +69,7 @@ unknown-key = 1
       |
     2 | unknown-key = 1
       | ^^^^^^^^^^^
-    unknown field `unknown-key`, expected one of `files`, `ignore`, `select`, `extend-select`, `per-file-ignores`, `extend-per-file-ignores`, `line-length`, `file-extensions`, `fix`, `no-fix`, `unsafe-fixes`, `no-unsafe-fixes`, `show-fixes`, `no-show-fixes`, `fix-only`, `no-fix-only`, `output-format`, `preview`, `no-preview`, `progress-bar`
+    unknown field `unknown-key`, expected one of `files`, `fix`, `no-fix`, `unsafe-fixes`, `no-unsafe-fixes`, `show-fixes`, `no-show-fixes`, `fix-only`, `no-fix-only`, `output-format`, `preview`, `no-preview`, `progress-bar`, `ignore`, `select`, `extend-select`, `per-file-ignores`, `extend-per-file-ignores`, `file-extensions`, `exclude`, `extend-exclude`, `line-length`
```
ZedThree, line 72: Hmm, this is perhaps getting a little long. Looks like ruff suppresses this somehow. Once we start adding individual rule settings, this might just not be helpful.

ZedThree:
Should we add some default `exclude` values? Ruff has:

```rust
pub(crate) static EXCLUDE: &[FilePattern] = &[
    FilePattern::Builtin(".bzr"),
    FilePattern::Builtin(".direnv"),
    FilePattern::Builtin(".eggs"),
    FilePattern::Builtin(".git"),
    FilePattern::Builtin(".git-rewrite"),
    FilePattern::Builtin(".hg"),
    FilePattern::Builtin(".ipynb_checkpoints"),
    FilePattern::Builtin(".mypy_cache"),
    FilePattern::Builtin(".nox"),
    FilePattern::Builtin(".pants.d"),
    FilePattern::Builtin(".pyenv"),
    FilePattern::Builtin(".pytest_cache"),
    FilePattern::Builtin(".pytype"),
    FilePattern::Builtin(".ruff_cache"),
    FilePattern::Builtin(".svn"),
    FilePattern::Builtin(".tox"),
    FilePattern::Builtin(".venv"),
    FilePattern::Builtin(".vscode"),
    FilePattern::Builtin("__pypackages__"),
    FilePattern::Builtin("_build"),
    FilePattern::Builtin("buck-out"),
    FilePattern::Builtin("dist"),
    FilePattern::Builtin("node_modules"),
    FilePattern::Builtin("site-packages"),
    FilePattern::Builtin("venv"),
];
```

We probably want at least the version control ones, maybe some of the others too.

LiamPattinson:
> I think `--include` is essentially the positional arguments, basically `files`?

I think it's just a way to put an inclusive filter on the file choices. If you set it to some subset of a project in `ruff.toml` (`src/myproject/some_dir/*.py` and then try linting some other non-overlapping subset (`src/myproject/some_other_dir/`) , it responds by saying no Python files were found. I'm not sure why it's an option in the config file but not on the command line though.

> Should we add some default exclude values?

Yes, it's probably worth excluding things like `.git` and `.vscode` by default. Perhaps also `build` and `_build`, as I think that would work for 90% of `fpm` and Cmake users?

fortitude/src/cli.rs
```diff
@@ -145,4 +106,96 @@ pub struct CheckArgs {
...
+    /// Like --exclude, but adds additional files and directories on top of those already excluded.
+    #[arg(
+        long,
+        value_delimiter = ',',
+        value_name = "FILE_PATTERN",
+        help_heading = "File selection"
+    )]
+    pub extend_exclude: Option<Vec<FilePattern>>,
+
+    /// Enforce exclusions, even for paths passed to Ruff directly on the command-line.
```
ZedThree, line 190:
```suggestion
    /// Enforce exclusions, even for paths passed to Fortitude directly on the command-line.
```

LiamPattinson: Whoops! It was only a matter of time before I let one of those slip through

LiamPattinson:
I've added a list of default exclude directories and tweaked the logic in `get_files` a bit to make sure the file patterns work as expected. I also got a few tests in, but `fortitude/tests/check.rs` is beginning to get a bit unwieldy.

I think that respecting `.gitignore` will require swapping out `walkdir` for the `ignore` crate and rewriting the `get_files` function accordingly. I think I'd prefer to leave that for its own PR.

README.md
```diff
@@ -152,6 +152,36 @@ preview = true
 
 Run `fortitude explain` to see which rules are in preview mode.
 
+### Excluding Files
```
ZedThree, line 155: Maybe instead this should be in `docs/index.md`?

Actually, looking at ruff, they have some tools for extracting portions of the `README.md` into the site docs. We should look at that at some point

README.md
```diff
@@ -152,6 +152,36 @@ preview = true
...
+
+> [!NOTE]
+> To be featured in v0.7.0
+
+Fortitude will automatically ignore files in some directories (`build/`, `.git/`,
+`.venv/`, etc.), and this behaviour can be extended using the `--exclude` option. For
+example, to ignore all files in the directory `benchmarks/`:
```
ZedThree, line 162: Oh, does `--exclude` not override the builtins?

fortitude/tests/check.rs
```diff
@@ -851,3 +852,244 @@ per-file-ignores = [
...
+    fs::write(&config_file, config).unwrap();
+    base_path
+}
+
+#[test]
+fn check_exclude() -> anyhow::Result<()> {
+    let tempdir = TempDir::new()?;
+    apply_common_filters!();
+    // Expect:
+    // - Overwrite 'foo.f90' in config file, see 'base.f90' and 'foo.f90' but not 'bar.f90'
```
ZedThree, line 897: Lowest possible priority, but you meant this?

```suggestion
    // - Override 'foo.f90' in config file, see 'base.f90' and 'foo.f90' but not 'bar.f90'
```

ZedThree (approved):

README.md
LiamPattinson: From playing around with Ruff, apparently not, as it behaves as if the built-ins are a mandatory set of `--extend-exclude` instead of `--exclude`. If you want it to lint everything in your `venv`, you have to manually pass it `venv/lib/Python3.10/site-packages`, and setting your own `exclude` doesn't change this behaviour.

I haven't played around with `include` in a config file, so I'm not sure how that might affect it.

README.md
LiamPattinson: For now, I've moved that section to the docs along with some other things from the README.

ZedThree (approved):

PR #239, opened 2024-12-16
Add rules for magic/non-portable IO units

### LINKED ISSUE #146: Suggest the use of `input_unit` and `output_unit` instead of units 5 and 6
For the code below I suggest that the warnings in the comments be given.

```
implicit none
integer :: i
write (6,*) "enter an integer" ! use output_unit from iso_fortran_env instead
read (5,*) i ! use input_unit from iso_fortran_env instead
print*,"i =",i
end
```

### LINKED ISSUE #206: Units should not be hardcoded in procedures
For the code

```fortran
subroutine foo()
implicit none
integer :: i
read (10, *) i
end subroutine foo
```

fortitude could warn about the use a hard-coded unit number in a procedure, since unit 10 may or may not be opened for reading or writing elsewhere in the program. Unit numbers in procedures should be passed as arguments or obtained using the `newunit` specifier.

ZedThree:
Closes #206 
Closes #146 

Unfortunately these are overlapping rules, although some projects might have their own equivalent of `input_unit` and so would want `magic-io-unit` but not `non-portable-io-unit`. I think this is one place where ruff's approach of handling check functions would work quite well. We'd have one function that we'd call if either rule was enabled, and it could do the check once and return the apropriate `Violation`.

### DIFF
--- a/fortitude/src/ast.rs
+++ b/fortitude/src/ast.rs
@@ -209,7 +209,7 @@ pub fn dtype_is_plain_number(dtype: &str) -> bool {
     )
 }
 
-/// If `node` is a keyword argument with name `keyword`, return the corresponding value
+/// Returns `true` if `node` is a keyword argument with name `keyword`.
 pub fn is_keyword_argument<S: AsRef<str>>(node: &Node, keyword: S, src: &str) -> bool {
     if node.kind() != "keyword_argument" {
         return false;
--- a/fortitude/src/ast.rs
+++ b/fortitude/src/ast.rs
@@ -208,3 +208,17 @@ pub fn dtype_is_plain_number(dtype: &str) -> bool {
         "integer" | "real" | "logical" | "complex"
     )
 }
+
+/// If `node` is a keyword argument with name `keyword`, return the corresponding value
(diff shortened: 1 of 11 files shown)


fortitude/src/ast.rs
```diff
@@ -208,3 +208,17 @@ pub fn dtype_is_plain_number(dtype: &str) -> bool {
         "integer" | "real" | "logical" | "complex"
     )
 }
+
+/// If `node` is a keyword argument with name `keyword`, return the corresponding value
```
LiamPattinson, line 212:
```suggestion
/// Returns `true` if `node` is a keyword argument with name `keyword`.
```

LiamPattinson (commented):
Looks good, thanks for this!

Could we perhaps rename `magic-io-unit` to `literal-io-unit`? It would be more consistent with rules such as `literal-kind`.

LiamPattinson:
Good points! I agree that there's a bit of a difference between 'this is a literal' and 'this is a magic number' -- for literal kinds (almost), everyone agrees what the `8` means in `real(8)`, so I wouldn't consider it a 'magic' number. However, in `open(513)` the main problem is that `513` has been pulled out of thin air.

I'm happy to leave the name as-is.

LiamPattinson (approved):

PR #240, opened 2024-12-16
Add `missing-default-case`

### LINKED ISSUE #213: Warn if there is no default case in a select statement
It is good practice to include a default case in a select statement to catch unknown values and either handle them appropriately or flag an error.

ZedThree:
Closes #213


LiamPattinson (approved):
Looks good!

LiamPattinson:
For the output, a similar problem arises for `old-style-array-literal`. It would be nice if there were a way for `annotate-snippets` to automatically abbreviate very tall annotations.

PR #242, opened 2024-12-17
Add "allow" comments

### LINKED ISSUE #209: Per-line ignore
Hi, thanks for creating Fortitude - I'm finding it really useful!

Would it be feasible to introduce per-line ignores? My use case is that I have a program that uses operator overloading and I'm not aware of a way to import the overloaded operators without using a bare `use` statement (no `only` specifications). So in this case I'd put a comment after the `use` statement telling Fortitude to ignore M011.

ZedThree:
Closes #209

Comments are of the form:

    ! allow(rule)

Usual rule selection applies, list of comma separated selectors

Comments apply to the whole of the next statement, including
constructs, procedures, modules, etc

Needs some proper docs, but I'm not sure where it belongs. Maybe we
need to add a whole tutorial section?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,6 +90,30 @@ files using `--per-file-ignores`:
 fortitude check --per-file-ignores=**/*.f95:non-standard-file-extension
 ```
 
+as well as for individual statements through "allow" comments of the
+form:
+
+```f90
+! allow(rule-or-category)
+```
+
+Multiple rules or categories can be given as a comma-separated
+list. Allow comments apply to the next statement and its contents. So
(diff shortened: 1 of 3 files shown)


LiamPattinson:
Amazing! I was expecting this to be much harder to achieve, and it works great.

I'd be in favour of adding a tutorial to the docs, so I've just raised an issue (https://github.com/PlasmaFAIR/fortitude/issues/243). For now, perhaps just a section in the README?

LiamPattinson (approved):
Happy to merge, provided a tutorial can be added soon-ish

LiamPattinson (approved):
All looks good!

Issue #64, opened 2024-10-04
Avoid multiple statements on the same line


ZedThree:
This hurts readability. Might be easiest with a text rule, just find `';'`

LiamPattinson:
Do you think it would be worth also catching multiple variable declarations on the same line? There are times when I think it doesn't hurt readability:

```f90
integer :: i, j, k
```

And others where it's just confusing:

```f90
real :: pi, some_array(:, :) time
```

ZedThree:
There's probably some useful variations like that that could be done as separate rules:

- Avoid mixing array and scalar variable declarations on the same line
- Avoid declaring multiple dummy arguments on the same line (maybe)

Potentially also "avoid declaring dummy arguments and local variables on the same line", but that would/should actually be fixed by "missing intent"

PR #246, opened 2024-12-18
Rule: `multiple-statements-per-line`

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/64

This was a lot harder to implement than I expected!

- Adds a function that replaces all text in strings and comments with whitespace/'!' respectively, allowing text rules to avoid matching on characters in that context.
  - The difficulty here is catching multiline strings, and the fact that the following is valid:
```f90
"This is a character string &
  & going over &
  !!! surprise comment !!!
  & multiple lines"
```
- Adds rule that finds semi-colons in code and replaces them with a newline and an appropriate amount of indentation.

I've left this as a draft as I'd like to add a companion rule, `superfluous-semicolon`, for a semicolon at the end of a line, and I still need to add tests.


ZedThree:
Could you do this without the blanking by something like: find `;` nodes in the AST, then check if the next non-comment sibling is on the same line?

LiamPattinson:
As far as I can tell `;` doesn't show up in the AST. If you switch the search in `check.rs` from `root_node.named_descendants()` to just `root_node.descendants()` and print every node kind, things like `::`, `,` and `=` show up, but there's no sign of `;`. I'm not sure if I'm just missing something though.

LiamPattinson:
I've added a rule `superfluous-semicolon`, but ultimately I don't think it would be a good idea to merge these rules in their current state.

Blanking all comments and strings is a very expensive operation, and as we can't have one rule return multiple kinds of violation, it has to be done twice per file if both rules are active. Benchmarking suggests this slows the code by ~7%:
```bash
❯ hyperfine -i './target/release/fortitude check --output-format=concise --preview --ignore=S081,S082 ../geos-chem'
Benchmark 1: ./target/release/fortitude check --output-format=concise --preview --ignore=S081,S082 ../geos-chem
  Time (mean ± σ):      1.763 s ±  0.046 s    [User: 8.042 s, System: 0.396 s]
  Range (min … max):    1.684 s …  1.808 s    10 runs
 
❯ hyperfine -i './target/release/fortitude check --output-format=concise --preview ../geos-chem'
Benchmark 1: ./target/release/fortitude check --output-format=concise --preview ../geos-chem
  Time (mean ± σ):      1.883 s ±  0.075 s    [User: 8.973 s, System: 0.414 s]
  Range (min … max):    1.761 s …  1.996 s    10 runs
 
```

The difference would be a lot less noticeable in a realistic IO-constrained run (`geos-chem` takes almost 10 seconds on my machine with full output!), but this still indicates that we need a less expensive solution to this problem.

@ZedThree Would it be possible to have `tree-sitter` recognise semicolons in the AST?

ZedThree:
Ah, I think the issue is that "end-of-statement" is an unnamed node and done entirely through an external scanner.

We could change the grammar to something like:

```
_end_of_statement: $ => choice(';', $._scan_end_of_statement),
```

ZedThree:
We've just released `tree-sitter-fortran` 0.4.0 which has `;` as an unnamed node, so we should now be able to significantly simplify this rule

LiamPattinson:
Fantastic, I'll try to get that going now.

LiamPattinson:
I've got it working with the new `tree-sitter-fortran` features -- this is a huge improvement!

TODO:

- [ ] Add tests
- [ ] Match local indentation when applying fixes -- I'll add some generic functions for this, as it's likely to come up again.

fortitude/src/ast.rs
```diff
@@ -161,6 +164,18 @@ impl FortitudeNode<'_> for Node<'_> {
...
+    fn indentation(&self, source_file: &SourceFile) -> String {
+        let src = source_file.to_source_code();
+        let start_byte = TextSize::try_from(self.start_byte()).unwrap();
+        let start_index = src.line_index(start_byte);
+        let start_line = src.line_start(start_index);
+        let start_offset = start_byte - start_line;
+        let line = src
+            .slice(TextRange::new(start_line, start_line + start_offset))
+            .to_string();
+        line.chars().take_while(|&c| c.is_whitespace()).collect()
```
ZedThree, line 176: We might have some fun with tabs here! (although tabs are technically not valid characters in fortran source)

ZedThree (approved):

LiamPattinson:
I think I'm finally there. The fixes behave strangely with regards to indentation if the user is doing anything really weird, like using multiple semicolons in a row, but should hold up in most cases.

fortitude/src/ast.rs
LiamPattinson: I figured tabs might be present in some codes, and hence I wrote `indentation` to return a string of the current indentation level rather than a count of the number of spaces. That way it can include a mix of tabs and spaces. I _think_ this is the safest way to handle them, but could be totally wrong on that.

On this topic, we should probably have a 'no tabs allowed' rule.

Issue #247, opened 2024-12-19
`line-length` and `file-extensions` ignored in config files


LiamPattinson:
As these options have default values specified in the CLI, lines such as the following always return the default value over those specified in the config file:

https://github.com/PlasmaFAIR/fortitude/blob/a185b11962091813f77827486b92f1ad2beb16bb/fortitude/src/check.rs#L696-L698

The following lines are at fault:

https://github.com/PlasmaFAIR/fortitude/blob/a185b11962091813f77827486b92f1ad2beb16bb/fortitude/src/cli.rs#L100-L105

I think the solution will be to manually state the default values in the doc-comments and fall back on the stated defaults in `src/check.rs`

PR #251, opened 2024-12-20
Bugfix: `line-length` and `file-extensions` ignored in toml files

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/247

It took me an embarrassingly long time to realise that `#[derive(Default)]` on `CheckSettings` was setting `line_length = 0` in cases where a config file wasn't present.

The fix here is pretty ugly, but it should be quite short-lived, as the solution to https://github.com/PlasmaFAIR/fortitude/issues/235 will likely mean giving back default values to `line-length` and `file-extensions` when they're deserialised from TOML files.


ZedThree (approved):
Can we add a test to catch this?

LiamPattinson:
Sorry for the delay on this one, I've added a simple test for this.

A merge conflict with main had tried to interleave several test snapshots and was almost unreadable, so I found it easier to remove the test, perform the merge, and then add it back again at the end of the file.

ZedThree (approved):

PR #254, opened 2024-12-26
Add check for use of specific names for intrinsic functions

imciner2:
This check uses a lot of the machinery that the comparator check uses, but just applies it to the call nodes to identify the functions.

There is still one point I am not that sure how to handle though - how to do a fix for multiple issues on one line. This can be seen in the example `test3`, where each violation is getting its own fix instead of both being included in a single fix (since they are all on one line).

### DIFF
--- a/fortitude/resources/test/fixtures/obsolescent/OB031.f90
+++ b/fortitude/resources/test/fixtures/obsolescent/OB031.f90
@@ -1,0 +1,21 @@
+subroutine test()
+    use, intrinsic :: iso_fortran_env, dp => real64
+    real(kind=dp) :: x, y
+
+    y = ASIN(x)
+    y = DSIN(x)
+end subroutine test
+
+subroutine test1()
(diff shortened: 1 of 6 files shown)


ZedThree:
Nice, thanks @imciner2! I think having a separate fix for each instance is fine, and likely the correct thing to do, rather than trying to merge all instances on the same line. The end result will be the same anyway.

My only comment really is that we should preferably try to match the case of what we're replacing -- or if that's not possible, prefer lowercase.

imciner2:
I've changed it to try to preserve the case of the function, at least based on what the case of the first character of the existing function is. This same paradigm should probably be applied to the other rules that replace strings (e.g., the `unamed-end-statements` rule at least), but that can be a later PR.

ZedThree (approved):
Lovely, thanks @imciner2!

Yes, agreed that matching-case is more generally useful for other rules too.

PR #265, opened 2025-01-09
Add `RuleSet`, `RuleTable` from ruff; allow ignoring `IoError`

ZedThree:
`RuleSet` and `RuleTable` are massive overkill for just this feature, but this will also allow ignoring allow-comment warnings, properly testing preview/deprecated/removed rules, and converting to the more explicit/verbose ruff checking structure.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -925,6 +925,12 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
                         )]));
                     } else {
                         // TODO: log::warn
+                        eprintln!(
+                            "{}{}{} {error}",
+                            "Error opening file ".bold(),
+                            fs::relativize_path(path).bold(),
+                            ":".bold()
+                        );
                         return None;
                     }
                 }
@@ -953,6 +959,12 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
                         )]))
                     } else {
                         // TODO: log::warn
+                        eprintln!(
+                            "{}{}{} {msg}",
+                            "Failed to process ".bold(),
+                            fs::relativize_path(path).bold(),
+                            ":".bold()
+                        );
                         None
                     }
                 }
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -229,7 +229,7 @@ fn parse_config_file(config_file: &Option<PathBuf>) -> Result<CheckSettings> {
 }
 
 /// Get the list of active rules for this session.
-fn ruleset(args: RuleSelection, preview: &PreviewMode) -> anyhow::Result<RuleTable> {
+fn to_rule_table(args: RuleSelection, preview: &PreviewMode) -> anyhow::Result<RuleTable> {
     let preview = PreviewOptions {
         mode: *preview,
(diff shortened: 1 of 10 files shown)


fortitude/src/check.rs
```diff
@@ -228,7 +229,7 @@ fn parse_config_file(config_file: &Option<PathBuf>) -> Result<CheckSettings> {
 }
 
 /// Get the list of active rules for this session.
-fn ruleset(args: RuleSelection, preview: &PreviewMode) -> anyhow::Result<Vec<Rule>> {
+fn ruleset(args: RuleSelection, preview: &PreviewMode) -> anyhow::Result<RuleTable> {
```
LiamPattinson, line 232: To avoid confusion with `RuleSet`, I think this function should be renamed `rule_table`.

fortitude/src/check.rs
```diff
@@ -905,17 +911,22 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
...
 
             let source = match read_to_string(path) {
                 Ok(source) => source,
                 Err(error) => {
-                    let message = format!("Error opening file: {error}");
-                    return Diagnostics::new(vec![DiagnosticMessage::from_error(
-                        filename,
-                        Diagnostic::new(IoError { message }, TextRange::default()),
-                    )]);
+                    if rules.enabled(Rule::IoError) {
```
LiamPattinson, line 920: I'm not sure if we should allow `IoError` to be switched off. Doing so would also break the code in `write_summary_text` in `printer.rs`, which counts the number of `IoError` raised to determine the number of files which were scanned.

LiamPattinson (commented):
I've had a look through and there were just a couple of things that jumped out at me:

fortitude/src/check.rs
ZedThree: Works fine for me:

```console
$ cargo run --bin fortitude -- check nope.f90 --ignore=E
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/fortitude check nope.f90 --ignore=E`
fortitude: 1 files scanned.
All checks passed!

```

The files we can't open get reported as files scanned, but this is probably fine.

Ruff issues a warning if you ignore `E902`:

```console
$ ruff check nope.py --ignore=E
warning: Failed to lint nope.py: No such file or directory (os error 2)
```

I've not done this yet as we probably want a proper logging system in place first. We could just write to stderr for now instead though.

LiamPattinson: I would argue that it isn't working fine in that case, as it's still giving misleading output. A user could quite easily switch off the error reporting by accident, and end up being told their file has no errors when in fact it hasn't been scanned at all:

```bash
cargo run --bin fortitude check --select=typing typo.f90
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/fortitude check --select=typing typo.f90`
fortitude: 1 files scanned.
All checks passed!
```

If you would rather not fix it here, we could merge for now and raise a new issue for it.

ZedThree: Yes, that's fair. I'll print a warning to stderr for now, but we should switch to a proper logger

LiamPattinson (approved):
Happy to merge if you are

Issue #252, opened 2024-12-20
Better warnings/errors for allow-comments


ZedThree:
Currently we just abort if an allow comment is not a valid rule, but we should instead raise a violation, which would allow us to include the line number etc.

We should also warn if an allow-comment isn't used and so can be removed.

PR #266, opened 2025-01-09
Warn about invalid rules in allow-comments

ZedThree:
~~Includes #265~~

Partial fix for #252 -- could still add a check for unused allow-comments


LiamPattinson (approved):
Nice one! This looks good to me.

Issue #257, opened 2025-01-02
Source files should have one module or a main program


Beliavsky:
For a code such as

```fortran
module foo
implicit none
end module foo

module bar
implicit none
end module bar

implicit none
print*,"hi"
end
```
I suggest that Fortitude warn that there are multiple modules defined in a single source file and that the source file has both a module and a main program. It could also warn that the modules are empty and not used in the program.

ZedThree:
A related rule would be to ensure the filename matches the module or program name, though some projects may have their own naming convention.

> It could also warn that the modules are empty 

Should be possible, although this could be an intentional side effect of conditional compliation.

> and not used in the program.

Not sure about this point though, modules might be part of a library rather than a program. We'd also need project-level information, and we're a ways off that yet.

Issue #163, opened 2024-11-29
Warn about include statements


Beliavsky:
Fortitude is silent about

```fortran
implicit none
include "inc.f90"
i = 3
print*,i
end
```
with inc.f90 containing
`integer :: i(2)`

but `include` statements can often be replaced by declaring variables and defining constants in a module

ZedThree:
I struggle with the categorisation sometimes too! We've sometimes wondered about whether we should instead have something more like tags than categories, but then we'd have to deal with situations like what happens if I disable `Style` but enable `Module` -- what should happen to a rule that's in both?

At some point we're going to need to have a reckoning with the current categories, and exactly where the rules fit best.

My gut says this more `Module` than `Style`, though I'm not sure I have a terribly convincing argument

LiamPattinson:
I would argue for `Module`, as `Style` should mostly just consider aesthetic issues with the code, although we already have a few rules in that category that go beyond that.

The categories are a bit of a headache, and I'm not sure how best to handle them going forward. I was aiming for something analogous to Ruff's categories, although targetting specific error classes instead of Ruff's categorisation of "which linter did we 'borrow' this from". However, we keep running into problems like this, where the existing categories are either too rigid or  a given rule could easily fit into multiple of them. An alternative would be to use something closer to Clippy's categorisation, in which rules are sorted based on how pedantic/opinionated they are, although I'm not convinced that would be any better.

PR #268, opened 2025-01-09
Add rules for multiple modules/programs in same file; Add rule for `include` statements

imciner2:
This implements three new rules in the module category:
* Warn on include statements (https://github.com/PlasmaFAIR/fortitude/issues/163)
* Warn if two or more modules are in the same file (https://github.com/PlasmaFAIR/fortitude/issues/257)
* Warn if a program and a module (or multiple) are in the same file (https://github.com/PlasmaFAIR/fortitude/issues/257)

The last two rules required a change to the check driver to also run the AST rules on the root node, since before it was only on the children of the root node. I also decided that instead of placing the violation diagnostic on the main module/program node, it would be better placed on just the opening statement's node - that way the display wouldn't try to print the entire module to the output.

One thing that could be improved still is the handling of the allow comments for the last two rules. Ideally there would be only one comment needed on one of the violations, and it would silence it for the entire file, but right now the way the allow comments are built and rules are defined, you have to use one comment on each violation.

CC @Beliavsky

(Fixes https://github.com/PlasmaFAIR/fortitude/issues/163, Fixes https://github.com/PlasmaFAIR/fortitude/issues/257 )

### DIFF
--- a/fortitude/resources/test/fixtures/modules/M031.f90
+++ b/fortitude/resources/test/fixtures/modules/M031.f90
@@ -1,0 +1,13 @@
+module test
+
+include "common_variables"
+
+contains
+
+integer function func1()
+    include "variable_declaration"
+
+    func1 = i + 3
+end function
+
+end module test
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -33,6 +33,7 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::fs::File;
 use std::io;
 use std::io::Write;
+use std::iter::once;
 use std::path::{Path, PathBuf};
 use std::process::ExitCode;
 use std::str::FromStr;
@@ -520,18 +521,8 @@ pub(crate) fn check_only_file(
         .parse(file.source_text(), None)
         .context("Failed to parse")?;
 
-    let root_node = tree.root_node();
-    if let Some(rules) = ast_entrypoints.get(root_node.kind()) {
-        for rule in rules {
-            if let Some(violation) = rule.check(settings, &root_node, file) {
-                for v in violation {
-                    violations.push(v);
-                }
-            }
-        }
-    }
-
-    for node in tree.root_node().named_descendants() {
+    let root = tree.root_node();
+    for node in once(root).chain(root.named_descendants()) {
         if let Some(rules) = ast_entrypoints.get(node.kind()) {
             for rule in rules {
                 if let Some(violation) = rule.check(settings, &node, file) {
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -606,19 +606,8 @@ pub(crate) fn check_and_fix_file<'a>(
             .parse(transformed.source_text(), None)
             .context("Failed to parse")?;
 
-        // Perform AST analysis, starting with root-level rules
-        let root_node = tree.root_node();
-        if let Some(rules) = ast_entrypoints.get(root_node.kind()) {
-            for rule in rules {
-                if let Some(violation) = rule.check(settings, &root_node, &transformed) {
-                    for v in violation {
-                        violations.push(v);
-                    }
-                }
-            }
-        }
-
-        for node in tree.root_node().named_descendants() {
+        let root = tree.root_node();
+        for node in once(root).chain(root.named_descendants()) {
             if let Some(rules) = ast_entrypoints.get(node.kind()) {
                 for rule in rules {
                     if let Some(violation) = rule.check(settings, &node, &transformed) {
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -519,6 +519,18 @@ pub(crate) fn check_only_file(
     let tree = parser
         .parse(file.source_text(), None)
         .context("Failed to parse")?;
+
+    let root_node = tree.root_node();
+    if let Some(rules) = ast_entrypoints.get(root_node.kind()) {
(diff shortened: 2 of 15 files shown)


fortitude/src/check.rs
```diff
@@ -519,6 +519,18 @@ pub(crate) fn check_only_file(
...
+        for rule in rules {
+            if let Some(violation) = rule.check(settings, &root_node, file) {
+                for v in violation {
+                    violations.push(v);
+                }
+            }
+        }
+    }
+
     for node in tree.root_node().named_descendants() {
```
ZedThree, line 534: We can simplify this to:

```suggestion
    for node in [tree.root_node()].into_iter().chain(tree.root_node().named_descendants()) {
```

Only a little ugly!

ZedThree:
> One thing that could be improved still is the handling of the allow comments for the last two rules. Ideally there would be only one comment needed on one of the violations, and it would silence it for the entire file, but right now the way the allow comments are built and rules are defined, you have to use one comment on each violation.

Maybe this would be nicer if we could include multiple snippets in one violation, so that it would look something like:

```text
./resources/test/fixtures/modules/M041.f90:9:1: M041 Multiple modules in one file, split into one module per file
   |
 9 | module mod2
   | ^^^^^^^^^^^ M041
10 |     use :: mod1, only: test1
   |

  |
1 | module mod1
  | ~~~~~~~~~~~ previous module already defined in this file
2 |
3 | contains
  |
```

This could be useful for other violations too. Not sure how much work is involved though!

Arguments could be made whether it would be the first module or the subsequent ones that would actually raise the violation.

I think the current form is probably fine for now, though it would be good to get some real world feedback!

fortitude/src/check.rs
LiamPattinson: You can get it a little prettier with `std::iter::once`:

```rust
    let root = tree.root_node();
    for node in once(root).chain(root.named_descendants()) {
```

LiamPattinson:
> Arguments could be made whether it would be the first module or the subsequent ones that would actually raise the violation.

It would simplify things if only the subsequent ones raised a violation. Could you do this by starting from `module_statement`/`program_statement` nodes and then iterating over `prev_sibling` to search for another `module`/`program`? Or if there some way modules/programs can be ancestor/descendant nodes? Doing it that way would also mean we don't need to include `translation_unit` in the core algorithm.

fortitude/resources/test/fixtures/modules/M031.f90
```diff
@@ -0,0 +1,13 @@
+module test
+
+include "common_variables"
```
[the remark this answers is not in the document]
LiamPattinson: Yeah, they probably should all compile. Maybe this could be done via a CI job?

imciner2:
> It would simplify things if only the subsequent ones raised a violation. Could you do this by starting from module_statement/program_statement nodes and then iterating over prev_sibling to search for another module/program? Or is there some way modules/programs can be ancestor/descendant nodes? Doing it that way would also mean we don't need to include translation_unit in the core algorithm.

Only raising it on the latter can work, however I think trying to access them from a module/program node is still a bit clunky compared to using the translation_unit node directly. E.g., to only put them on the later ones, we can basically just skip the first node in the iterator in the current rule.

imciner2:
> Maybe this would be nicer if we could include multiple snippets in one violation, so that it would look something like:
> 
> ```
> ./resources/test/fixtures/modules/M041.f90:9:1: M041 Multiple modules in one file, split into one module per file
>    |
>  9 | module mod2
>    | ^^^^^^^^^^^ M041
> 10 |     use :: mod1, only: test1
>    |
> 
>   |
> 1 | module mod1
>   | ~~~~~~~~~~~ previous module already defined in this file
> 2 |
> 3 | contains
>   |
> ```
> 
> This could be useful for other violations too. Not sure how much work is involved though!

A quick look into Ruff's crates, and I don't think there is any support for multiple locations in their system right now, and based on the comment here https://github.com/astral-sh/ruff/issues/14900#issuecomment-2541849223, it sounds like it might not be something they are keen to implement. I don't know if there might be a way to put something together downstream or not though.

fortitude/src/rules/modules/file_contents.rs
```diff
@@ -0,0 +1,81 @@
...
+        format!("Program and module in one file, split into their own files")
+    }
+}
+
+impl AstRule for ProgramWithModule {
+    fn check(_settings: &Settings, node: &Node, _src: &SourceFile) -> Option<Vec<Diagnostic>> {
+        // Mark the violation on the second, and subsequent, occurences
+        let violations: Vec<Diagnostic> = node
+            .children(&mut node.walk())
+            .filter(|node| node.kind() == "module" || node.kind() == "program")
```
LiamPattinson, line 67: This will erroneously trigger in there are two modules in one file. It could be fixed by having `program_statement` as the entrypoint, going to the node parent (`translation_unit`), and then scanning for any modules

imciner2: Long-ish chains sometimes do my head in, so I was trying to avoid those (hence starting at the `translation_unit` root to begin with instead of going up the tree and back down). I add a check for `any` with the condition to ensure there is a program statement in the file.

ZedThree (approved):
Thanks for your work on these rules @imciner2! I think there's perhaps still some debate to be had over how best to report these, but that would require changes elsewhere, so I'm happy to get these in now

PR #274, opened 2025-01-10
Add logging framework and `--verbose/quiet/silent` flags

ZedThree:
Logging framework using [log](https://docs.rs/log/latest/log/index.html) and [fern](https://docs.rs/fern/latest/fern/) crates, and setup stolen from ruff, as is traditional.

Also add `--verbose/quiet/silent` flags. As the logging isn't used for much, this doesn't change very much, although we can skip the summary with `--quiet`, and skip printing absolutely everything but still return the exit code with `--silent`.

We can now sprinkle `debug!()` through the code to get some useful insight into internal state, for example some basic profiling:

```
[2025-01-10][17:53:49][fortitude::check][DEBUG] Identified files to lint in: 66.510789ms
[2025-01-10][17:53:50][fortitude::check][DEBUG] Checked 236 files in: 935.848477ms
```

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -23,6 +23,7 @@ use colored::Colorize;
 use indicatif::{ParallelProgressIterator, ProgressStyle};
 use itertools::Itertools;
 use lazy_regex::{regex, regex_captures};
+use log::warn;
 use rayon::prelude::*;
 use ruff_diagnostics::Diagnostic;
 use ruff_source_file::{Locator, SourceFile, SourceFileBuilder};
@@ -1000,8 +1001,7 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
                             Diagnostic::new(IoError { message }, TextRange::default()),
                         )]));
                     } else {
-                        // TODO: log::warn
-                        eprintln!(
+                        warn!(
                             "{}{}{} {error}",
                             "Error opening file ".bold(),
                             fs::relativize_path(path).bold(),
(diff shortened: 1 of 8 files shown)


fortitude/src/logging.rs
```diff
@@ -0,0 +1,167 @@
...
+                    record.target(),
+                    record.level(),
+                    message
+                ));
+            }
+        })
+        .level(level.level_filter())
+        .level_for("globset", log::LevelFilter::Warn)
+        .level_for("red_knot_python_semantic", log::LevelFilter::Warn)
+        .level_for("salsa", log::LevelFilter::Warn)
```
LiamPattinson, line 149: Can we remove these?

fortitude/src/check.rs
```diff
@@ -1048,14 +1052,23 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
             }
         });
 
-    let mut all_diagnostics = diagnostics_per_file
-        .fold(Diagnostics::default, |all_diagnostics, file_diagnostics| {
-            all_diagnostics + file_diagnostics
-        })
-        .reduce(Diagnostics::default, |a, b| a + b);
+    let (mut all_diagnostics, checked_files) = diagnostics_per_file
```
LiamPattinson, line 1055: Am I right in thinking that `checked_files` includes only files that were successfully scanned? If so, could we propagate this to `write_summary_text` so we can get the number of checked files right when `IoError` is being ignored? It would also be nice if we could propagate the duration so the message could say "checked x files in y seconds", but that could wait for a later issue/PR.

LiamPattinson (commented):
Very nice! Perhaps we should go through and add a bunch of extra debug logs at some point. Only spotted a couple of things to improve.

fortitude/src/logging.rs
ZedThree: I knew I'd forgotten something, thanks!

fortitude/src/check.rs
ZedThree: No, this is essentially a duplication of the same calculation in `write_summary_text`, it's just the same as what ruff reports at this stage (that is, the file was scanned, and the error was "couldn't be read")

LiamPattinson (approved):

PR #285, opened 2025-01-13
Respect `.gitignore`

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/231

I spent a while trying to get Ruff's implementation of this working, but it required pulling in a lot of additional infrastructure that I didn't fully understand, so this seemed the safer option.

The use of the `WalkBuilder` seems messy, but when I tried building and using it from a function chain I got all kinds of lifetime issues. Making it `mut` and modifying it one line at a time seemed to do the trick.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -14,12 +14,13 @@ use crate::rules::error::allow_comments::InvalidRuleCodeOrName;
 use crate::rules::Rule;
 use crate::rules::{error::ioerror::IoError, AstRuleEnum, PathRuleEnum, TextRuleEnum};
 use crate::settings::{
-    ExcludeMode, FilePattern, FilePatternSet, FixMode, OutputFormat, PatternPrefixPair,
-    PreviewMode, ProgressBar, Settings, UnsafeFixes, DEFAULT_SELECTORS,
+    ExcludeMode, FilePattern, FilePatternSet, FixMode, GitignoreMode, OutputFormat,
+    PatternPrefixPair, PreviewMode, ProgressBar, Settings, UnsafeFixes, DEFAULT_SELECTORS,
 };
 
 use anyhow::{anyhow, Context, Result};
 use colored::Colorize;
+use ignore::WalkBuilder;
 use indicatif::{ParallelProgressIterator, ProgressStyle};
 use itertools::Itertools;
 use lazy_regex::{regex, regex_captures};
@@ -40,7 +41,6 @@ use std::str::FromStr;
 use strum::IntoEnumIterator;
 use toml::Table;
 use tree_sitter::{Node, Parser};
-use walkdir::WalkDir;
 
 /// Default extensions to check
 const FORTRAN_EXTS: &[&str] = &[
@@ -185,6 +185,7 @@ pub struct CheckSettings {
     pub exclude: Option<Vec<FilePattern>>,
     pub extend_exclude: Vec<FilePattern>,
     pub exclude_mode: ExcludeMode,
+    pub gitignore_mode: GitignoreMode,
 }
 
 impl Default for CheckSettings {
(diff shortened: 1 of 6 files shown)


fortitude/src/check.rs
```diff
@@ -323,17 +328,21 @@ fn get_files<P: AsRef<Path>, S: AsRef<str>>(
     extensions: &[S],
     excludes: &FilePatternSet,
     exclude_mode: ExcludeMode,
+    gitignore_mode: GitignoreMode,
 ) -> Vec<PathBuf> {
     paths
```
ZedThree, line 333: Is it not possible to build the walker at this level? Then we can avoid building a new walker for each individual directory. It might also make it easier to switch to the parallel walker in the future

fortitude/src/check.rs
```diff
@@ -323,17 +328,21 @@ fn get_files<P: AsRef<Path>, S: AsRef<str>>(
...
-                    .filter_entry(|e| !excludes.matches(e.path()))
+                let excludes = excludes.clone(); // Needed to get around lifetime issues
+                let mut builder = WalkBuilder::new(path);
+                builder.standard_filters(gitignore_mode.into());
+                builder.hidden(false);
+                builder.filter_entry(move |e| !excludes.matches(e.path()));
+                builder
+                    .build()
                     .filter_map(|p| p.ok()) // skip dirs if user doesn't have permission
                     .filter(|p| is_valid_extension(p.path(), extensions))
```
ZedThree, line 347: Could use [`TypeBuilder`](https://docs.rs/ignore/0.4.23/ignore/types/struct.TypesBuilder.html) and [`WalkBuilder.types`](https://docs.rs/ignore/0.4.23/ignore/struct.WalkBuilder.html#method.types) to just walk over the requested extensions?

fortitude/src/check.rs
LiamPattinson: I tried to do it this way initially but ran into a bunch of problems getting it to work properly. I can give it another go.

fortitude/src/check.rs
LiamPattinson: I wasn't aware of that feature, I'll take a look.

fortitude/src/check.rs
LiamPattinson: https://github.com/PlasmaFAIR/fortitude/pull/285/commits/7869a739436198483392a181ff349aa812cc3884 implements this. It makes the function a fair bit messier, as invalid paths need to be handled separately. `WalkBuilder` will raise an error if it's given a path that doesn't exist, but we need them to persist (either that or we rewrite the whole way we're handling missing files, but as there are already outstanding bugs related to the reporting of missing files I'd rather keep things the way they are in this PR if we can).

fortitude/src/check.rs
LiamPattinson: Added in https://github.com/PlasmaFAIR/fortitude/pull/285/commits/5362fa00e9a1aac348af1b4dbf1c2fe329c903db

LiamPattinson:
I've made the requested changes. I think we may need to rethink how to handle invalid file paths and missing files, as the current method is to deliberately allow these to propagate out of `get_files` and then catch them here:

https://github.com/PlasmaFAIR/fortitude/blob/e03b00aeae0074b82ee1979384ee68bf1aad5ff9/fortitude/src/check.rs#L986-L1002

Perhaps it would be better to instead return `Vec<Result<PathBuf, _>>` from `get_files` and propagate the error directly. I think I'd prefer to handle that in a different issue/PR though.

ZedThree (approved):

ZedThree:
Could be an opportunity to add some `debug!()` messages for why files are included/excluded

LiamPattinson:
> Could be an opportunity to add some `debug!()` messages for why files are included/excluded

There are a couple of bug reports I've been meaning to raise related to file searching, and I think a more general refactor of this part of the code will be needed, so the PRs to fix those would be a better place to add them.

PR #290, opened 2025-01-14
Handle broken pipes

ZedThree:
Previously, piping the output of fortitude into `head`, for example,
would result in output like:

```console
$ fortitude check | head -n1
fortitude/resources/test/fixtures/bugprone/B001.f90:10:1: S101 [*] trailing whitespace
Error: Broken pipe (os error 32)
```

It seems the rust stdlib doesn't handle broken pipes very well, so
some manual handling must be done.

Ruff additionally allows reporting of multiple errors, but that isn't
done here, as we probably need to switch to bubbling up their `ExitStatus`

### DIFF
--- a/fortitude/src/main.rs
+++ b/fortitude/src/main.rs
@@ -1,7 +1,9 @@
+use std::io::Write;
 use std::process::ExitCode;
 
 use anyhow::Result;
 use clap::Parser;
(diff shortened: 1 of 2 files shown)


LiamPattinson (approved):
Looks good, I'd have never caught this!

Issue #288, opened 2025-01-14
`--force-exclude` doesn't work with directories


LiamPattinson:
Say you have the directory `.venv/lib/python3.10/site-packages/`, containing `myfile.f90`:

```f90
module m
end module m
```

and `myfile.py`:

```python
import numpy
```

Both Fortitude and Ruff will exclude anything in `.venv` by default, but calling the following will report that `numpy` is unused:

```bash
ruff check .venv/lib/python3.10/site-packages/myfile.py
```

The command line arg `--force-exclude` will make Ruff ignore this file even though it was explicitly provided:

```bash
ruff check --force-exclude .venv/lib/python3.10/site-packages/myfile.py
```

Unfortunately, `--force-exclude` doesn't behave the same in Fortitude, with the following both scanning the file:

```bash
# Scans file
fortitude check .venv/lib/python3.10/site-packages/myfile.f90
# Also scans file, but it shouldn't...
fortitude check --force-exclude .venv/lib/python3.10/site-packages/myfile.f90
```

This is fixed when adding the file name to the excludes list:

```bash
# Scans file
fortitude check --exclude=myfile.f90 .venv/lib/python3.10/site-packages/myfile.f90
# Returns nothing
fortitude check --exclude=myfile.f90 --force-exclude .venv/lib/python3.10/site-packages/myfile.f90
```

To fix this, we'll need to check `path.ancestors()` against the excludes list alongside the path itself. Ruff uses the following function to do so:

https://github.com/astral-sh/ruff/blob/5ed7b55b15b6b0ce52aee1e34dec221ee845007e/crates/ruff_workspace/src/resolver.rs#L692

ZedThree:
Looks like this is just way more complicated than we first thought!

PR #296, opened 2025-01-15
Bugfix: fix `get_files` force excluding with directories

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/288

Also tidied a lot of file-handling things into `fs.rs` and added some debugging logs.

TODO: Ruff will stop checking path ancestors on reaching the 'project root', but Fortitude keeps going. This means `--exclude=home --force-exclude` will prevent any files being scanned. I'll need to add a similar concept to Fortitude to get this working in full.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -109,6 +109,17 @@ pub fn find_settings_toml<P: AsRef<Path>>(path: P) -> Result<Option<PathBuf>> {
     Ok(None)
 }
 
+/// Find the path to the project root, which contains the `fpm.toml` or `fortitude.toml` file.
+/// If no such file exists, return the current working directory.
+fn project_root<P: AsRef<Path>>(path: P) -> Result<PathBuf> {
+    find_settings_toml(&path)?.map_or(Ok(fs::normalize_path(&path)), |settings| {
+        fs::normalize_path(settings)
+            .parent()
+            .context("Settings file has no parent")
+            .map(PathBuf::from)
+    })
+}
+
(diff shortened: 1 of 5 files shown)


LiamPattinson:
I _think_ I've replicated Ruff's behaviour here, but there are a lot of things to keep track of and I'm struggling to wrap my head around all of it. I can't quite understand what everything in `crates/ruff_workspace/src/resolver.rs` is doing, and they have a lot of extra machinery that we don't (yet?) need.

I thought I'd found an edge case where a combination of `--force-exclude`, `cd`-ing into an excluded directory, and running `check` on a nested directory would lead to different behaviours, but I'm struggling to replicate it and I'm now wondering if I just dreamed the whole thing. Let me know if you find any divergences between the two.

fortitude/src/check.rs
```diff
@@ -988,8 +917,10 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
     let ast_entrypoints = ast_entrypoint_map(&rules);
 
     let start = Instant::now();
+
     let files = get_files(
         files,
+        project_root(path_absolutize::path_dedot::CWD.as_path())?,
```
ZedThree, line 923: I guess this should actually be the directory containing the parent `fortitude.toml`/`fpm.toml` file? We do search up the directory structure to find one, so we should probably use that here

ZedThree: Ignore that, I was too focussed on the `CWD` bit I missed this was a function call!

ZedThree (approved):

PR #299, opened 2025-01-15
Add rule to test for uninitialized pointers in derived types

imciner2:
Unlike normal variables in procedures, derived types can have non-saved default initializer values, and it is safest to ensure that pointers are associated, so that they can then be tested using intrinsics.

This is just the first type of variable to check in the derived types. Further checks can suggest default initializers on intrinsic typed variables.

### DIFF
--- a/fortitude/src/rules/typing/derived_default_init.rs
+++ b/fortitude/src/rules/typing/derived_default_init.rs
@@ -61,7 +61,7 @@ impl Violation for MissingDefaultPointerInitalisation {
     #[derive_message_formats]
     fn message(&self) -> String {
         let Self { var } = self;
-        format!("'{var}' does not have a default initialiser")
+        format!("pointer component '{var}' does not have a default initialiser")
     }
 
     fn fix_title(&self) -> Option<String> {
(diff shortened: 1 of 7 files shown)


fortitude/src/rules/typing/derived_default_init.rs
```diff
@@ -0,0 +1,117 @@
...
+            .named_descendants()
+            .filter(|node| node.kind() == "identifier")
+            .filter(|node| match node.parent() {
+                None => false,
+                Some(parent) => parent.kind() == "variable_declaration",
+            })
+            .map(|node| {
+                let var_name = node.to_text(src.source_text()).unwrap_or("").to_string();
+
+                let init_var = format!("{var_name} => null()");
```
ZedThree, line 103: This duplicates the variable name, see snapshot below:

```diff
20    |-        integer, pointer :: pI1, pI2
   20 |+        integer, pointer :: pI1pI1 => null(), pI2
```

```suggestion
                let init_var = format!(" => null()");
```

fortitude/src/rules/typing/derived_default_init.rs
```diff
@@ -0,0 +1,117 @@
...
+#[violation]
+pub struct MissingDefaultPointerInitalisation {
+    var: String,
+}
+
+impl Violation for MissingDefaultPointerInitalisation {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let Self { var } = self;
+        format!("'{var}' does not have a default initialiser")
```
ZedThree, line 64: Maybe something more explicit:

```suggestion
        format!("pointer component '{var}' does not have a default initialiser")
```

or is this too long-winded?

ZedThree (approved):

PR #300, opened 2025-01-16
Add guide for adding new rules to `README.dev.md`

LiamPattinson:
Based loosely on [Ruff's guide for adding new rules](https://docs.astral.sh/ruff/contributing/#example-adding-a-new-lint-rule). Let me know if I've missed any steps or if something could be better explained. @imciner2 I'd be grateful if you could let me know of any additional barriers or points of confusion that you ran into while trying to add new rules.

Aside: Should the contents of `README.dev.md` be moved to the fancy docs, or perhaps duplicated there? Either way, I think that would be a matter for another PR.

### DIFF
--- a/README.dev.md
+++ b/README.dev.md
@@ -45,13 +45,71 @@ git config --local core.hooksPath .githooks
 
 ## Adding Rules
 
-Similarly to how [Ruff](https://docs.astral.sh/ruff/) names rules, the rules in
-Fortitude should be categorised appropriately and their name should describe the pattern
-the rule is intended to fix. Words such as 'forbid' should be omitted. For example, the
-name for the rule that warns of overly long lines is `LineTooLong`, and not something
-like `AvoidLineTooLong` or `KeepLinesShort`.
+We're always open to new rule suggestions, and would be very grateful for any
+assistance in implementing them. Before raising an issue to suggest a new rule,
+please check to see if it has already been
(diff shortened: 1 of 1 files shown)


imciner2:
Thanks! I think that covers everything that I found was needed to add a rule. The main learning curve was definitely to find where all the files are.

One other thing that might be good to add though is about including fixes, and the criteria for safe/unsafe fix that should be followed.

imciner2 (approved):

README.dev.md
```diff
@@ -45,13 +45,86 @@ git config --local core.hooksPath .githooks
...
+Rules should also be categorised appropriately. For example, if a rule is
+intended to discourage the use of outdated features, it may go under
+`Obsolescent`, or perhaps a more specific category such as `Modules` or
+`Typing`. If the rule only affects code readability, it should go under `Style`.
+
+The boundaries between categories are not always clear, so the exact name and
+category of a rule is often determined following a discussion after a pull
+request has been raised.
+
+#### Rule Testing: Fixtures and Snapshots
```
ZedThree, line 127: Very minor, but I think this is one level too deep? It looks like it's under `Naming and Categorising Rules`

ZedThree (approved):

README.dev.md
LiamPattinson: Whoops! Well caught, that was just a typo

PR #304, opened 2025-01-23
Add rule for deleted feature `pause` statements

### LINKED ISSUE #279: `pause` statement has been removed
Fortitude could warn about

```fortran
implicit none
print*,"hello"
pause
print*,"bye"
end
```
and suggest that  a `read` statement be used instead of `pause`.

ZedThree:
Closes #279

### DIFF
--- a/fortitude/src/rules/obsolescent/pause_statement.rs
+++ b/fortitude/src/rules/obsolescent/pause_statement.rs
@@ -26,6 +26,9 @@ impl Violation for PauseStatement {
     fn message(&self) -> String {
         format!("`pause` statements are a deleted feature")
     }
+    fn fix_title(&self) -> Option<String> {
+        Some("Use 'read(*, *)' instead".into())
+    }
 }
 
 impl AstRule for PauseStatement {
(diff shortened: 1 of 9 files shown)


fortitude/src/rules/obsolescent/pause_statement.rs
```diff
@@ -0,0 +1,44 @@
...
+///   'Obsolescent and Deleted Features'
+#[violation]
+pub struct PauseStatement {}
+
+impl Violation for PauseStatement {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        format!("`pause` statements are a deleted feature")
+    }
+}
```
LiamPattinson, line 29: Could we add a suggestion here?

```suggestion
    fn fix_title(&self) -> Option<String> {
        Some("Use 'read(*, *)' instead".into())
    }
}
```

LiamPattinson (approved):
Cheers, looks good to me! Sorry if I forgot to rebuild the docs, I should really put it in my precommit hook and add it to the CI.

fortitude/src/rules/obsolescent/pause_statement.rs
ZedThree: Oh yes, good point!

Issue #297, opened 2025-01-15
Override missing --preview with --select


imciner2:
Currently, if a rule is configured as being in preview, the flag `--preview` must be provided even if the rule is explicitly selected using the `--select` argument. This is counter-intuitive though, because you are already explicitly requesting the rule.

LiamPattinson:
I've been caught out by this before. Ruff doesn't override preview mode, but it does print a warning:

```bash
❯ uvx ruff check --select=W391 --preview myfile.py
myfile.py:2:1: W391 [*] Extra newline at end of file
  |
1 | print("hello world")
2 | 
  | ^ W391
  |
  = help: Remove trailing newline

Found 1 error.
[*] 1 fixable with the `--fix` option.
❯ uvx ruff check --select=W391 myfile.py
warning: Selection `W391` has no effect because preview is not enabled.
All checks passed!
```

imciner2:
I agree, a preview rule should not be included if the user does a more generic selection, it should only be included without `--preview` if it is explicitly requested by number.

I'm not a fan of the warning and not applying the rule when it was requested that ruff does there, I actually think applying the rule and printing a warning that the rule is in preview would be better.

ZedThree:
My view is that `--preview` doesn't just enable those rules, it's also something like a safety switch, with the user saying "I know what I'm doing and I agree to enabling this". If the rule later undergoes some tweaking before becoming stable, then users might apply a wrong fix that needs to be undone. Requiring the opt-in switch, even for a single rule, makes this a bit more explicit.

I suspect it might be a little bit more annoying at the minute when we have a whole bunch of new rules in preview :)

We definitely should have the warning, but I do also think we should still require `--preview` even if a single rule is selected.

ZedThree:
Closed via #305

PR #305, opened 2025-01-24
Warn when selected rules are ignored for some reason

ZedThree:
So far this only applies to preview rules selected without `--preview`, but will also apply to any future redirected, deprecated, or removed rules.

There's a lot more tests we could ~~steal~~ borrow from ruff related to this, but they all read from stdin so it's a little annoying to convert them. I'll look into how easy it is to get reading from stdin separately

Related to #297

### DIFF
--- a/fortitude/Cargo.toml
+++ b/fortitude/Cargo.toml
@@ -74,6 +74,8 @@ predicates = "3.1.2"
 pretty_assertions = "1.4.1"
 tempfile = "3.13.0"
 test-case = "3.3.1"
+# This is a bit weird, but this enables the test rules when we run the tests
+fortitude = { workspace = true, features = ["test-rules"] }
 
 [features]
 # Enables rules for internal integration tests
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -43,7 +43,7 @@ use std::str::FromStr;
 use std::time::Instant;
 use strum::IntoEnumIterator;
 use toml::Table;
-use tree_sitter::{Node, Parser};
+use tree_sitter::{Node, Parser, Tree};
 
 // These are just helper structs to let us quickly work out if there's
 // a fortitude section in an fpm.toml file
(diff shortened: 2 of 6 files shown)


fortitude/Cargo.toml
```diff
@@ -74,6 +74,8 @@ predicates = "3.1.2"
 pretty_assertions = "1.4.1"
 tempfile = "3.13.0"
 test-case = "3.3.1"
+# This is a bit weird, but this enables the test rules when we run the tests
+fortitude = { workspace = true, features = ["test-rules"] }
```
LiamPattinson, line 78: Can't say I've seen this before. How come we can't just use `#[cfg(test)]`?

fortitude/src/check.rs
```diff
@@ -492,6 +635,44 @@ pub(crate) fn check_only_file(
...
+    Ok(violations
+        .into_iter()
+        .map(|v| DiagnosticMessage::from_ruff(file, v))
+        .collect_vec())
+}
+
+/// Check an already parsed file. This actually does all the checking,
+/// `check_only_file`/`check_and_fix_file` wrap this
+#[allow(clippy::too_many_arguments)]
+pub(crate) fn check_path(
```
LiamPattinson, line 666: Glad to see that the repeat code has been cleaned up!

fortitude/src/check.rs
```diff
@@ -818,6 +994,54 @@ pub struct RuleSelection {
     pub select: Option<Vec<RuleSelector>>,
     pub ignore: Vec<RuleSelector>,
     pub extend_select: Vec<RuleSelector>,
+    pub fixable: Option<Vec<RuleSelector>>,
+    pub unfixable: Vec<RuleSelector>,
+    pub extend_fixable: Vec<RuleSelector>,
```
LiamPattinson, line 999: We haven't implemented these options. Is that something we should get done before the next release?

LiamPattinson (approved):
Nice one! I would argue that this resolves https://github.com/PlasmaFAIR/fortitude/issues/297, although not with the requested solution.

The comments I've left are purely for my own interest, I'd be happy to merge this if you are.

fortitude/src/check.rs
ZedThree: I think that would be good but not essential. I added them here to be able to just use the whole section from ruff

PR #307, opened 2025-01-24
Add ability to read from stdin

### LINKED ISSUE #136: Enable reading from stdin
This would be an easy workaround to check files that need preprocessing

ZedThree:
Closes #136

Also adds `--output-file` argument as a bonus

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -19,7 +19,8 @@ use crate::settings::{
     ExcludeMode, FixMode, GitignoreMode, OutputFormat, PatternPrefixPair, PreviewMode, ProgressBar,
     Settings, UnsafeFixes, DEFAULT_SELECTORS,
 };
-use crate::{fs, warn_user_once_by_id, warn_user_once_by_message};
+use crate::stdin::read_from_stdin;
+use crate::{fs, warn_user_once, warn_user_once_by_id, warn_user_once_by_message};
 
 use anyhow::{anyhow, Context, Result};
 use colored::Colorize;
@@ -36,8 +37,8 @@ use serde::Deserialize;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
 use std::fs::File;
-use std::io;
 use std::io::Write;
+use std::io::{self, BufWriter};
 use std::iter::once;
 use std::path::{Path, PathBuf};
 use std::process::ExitCode;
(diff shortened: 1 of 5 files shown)


ZedThree:
Couple of things to note:

- `check.rs` is really getting quite monstrous now!
- this doesn't obey `--force-exclude` if `--stdin-filename` is given, which ruff does. I've not done it here, because it would involve passing more arguments around and I was being lazy. I think ruff's `Resolver` class might be what we want here, not entirely sure yet

fortitude/src/check.rs
```diff
@@ -445,6 +449,39 @@ fn to_rule_table(args: RuleSelection, preview: &PreviewMode) -> anyhow::Result<R
...
+
+    let [file] = files else {
+        return false;
+    };
+    // If the user provided exactly `-`, read from standard input.
+    file == Path::new("-")
+}
+
+/// Returns the default set of files if none are provided, otherwise returns `None`.
+fn resolve_default_files(files: Vec<PathBuf>, is_stdin: bool) -> Vec<PathBuf> {
```
LiamPattinson, line 473: The doc-comment here is incorrect, it always returns a `Vec<PathBuf>`

LiamPattinson (approved):
I'm a little confused by some of what's going on here. Please could you post an example of how this would be used?

LiamPattinson:
> I think ruff's `Resolver` class might be what we want here, not entirely sure yet

It might be needed eventually, but that's a really complicated part of Ruff. I'm a little wary of importing it wholesale at this stage, as I can barely understand all of its moving parts and it would make a big chunk of this project near-unmaintainable from my perspective. On the other hand, Fortitude is already getting sufficiently complicated that it's hard to figure out the control flow at times, so migrating to a simplified version of `Resolver` may be unavoidable at some stage soon.

ZedThree:
This allows you to pipe preprocessor output into fortitude:

```console
$ gfortran -E -DFOO=1 bar.F90 | fortitude check --stdin-filename=bar.f90 -
```

or fypp, pfunit, etc.

There's also some editors and language servers that can use stdin (see the [efm language server bit](https://docs.astral.sh/ruff/editors/setup/#neovim))

> migrating to a simplified version of Resolver may be unavoidable at some stage soon.

Oh yes, I don't think we need the whole thing, but something that wraps up the exclude/file extensions stuff would probably be useful as it would shift some of the logic out of `check.rs`

Issue #308, opened 2025-01-27
Distributing binaries more directly


ZedThree:
Currently we really only support installing binaries via `pip`/`PyPI`. Ruff also supports downloading the binaries directly from their releases on GitHub, which they manage using [`dist`](https://opensource.axo.dev/cargo-dist/) ([GitHub](https://github.com/axodotdev/cargo-dist) in case their site is broken).

This looks fairly easy to use, and would give people a non-python related way of installing fortitude

PR #310, opened 2025-01-28
Add some tools for automating release workflow

ZedThree:
Two related things here:

- Use `dist` (see #308) to make a `release` workflow that uploads binaries to GitHub releases, along with an installer script so users don't need a Python environment
- Add `scripts/release.sh` that uses [`rooster`](https://github.com/zanieb/rooster/tree/main) to automate bumping version number in multiple files, and generating `CHANGELOG.md`

To make the changelog nicely structured, it's best if we label PRs -- I've gone through and labelled pretty much everything (I think), using labels inspired by ruff, obviously. I've used `rooster backfill` to then generate a changelog for previous versions.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -282,20 +282,6 @@ jobs:
           args: --release --locked --out dist
         env:
           CFLAGS: -std=c99
-      - uses: uraimo/run-on-arch-action@v2
-        if: matrix.platform.arch != 'ppc64'
-        name: Test wheel
-        with:
-          arch: ${{ matrix.platform.arch == 'arm' && 'armv6' || matrix.platform.arch }}
-          distro: ${{ matrix.platform.arch == 'arm' && 'bullseye' || 'ubuntu20.04' }}
-          githubToken: ${{ github.token }}
-          install: |
-            curl -LsSf https://astral.sh/uv/install.sh | sh
-          run: |
-            ~/.local/bin/uv venv --python 3.11
-            source .venv/bin/activate
-            ~/.local/bin/uv pip install ${{ env.PACKAGE_NAME }} --no-index --find-links dist/ --force-reinstall
-            fortitude --help
       - name: "Upload wheels"
         uses: actions/upload-artifact@v4
         with:


dist-workspace.toml
```diff
@@ -0,0 +1,17 @@
...
+# Config for 'dist'
+[dist]
+# The preferred dist version to use in CI (Cargo.toml SemVer syntax)
+cargo-dist-version = "0.28.0"
+# CI backends to support
+ci = "github"
+# The installers to generate for each app
+installers = ["shell"]
+# Target platforms to build apps for (Rust target-triple syntax)
+targets = ["aarch64-apple-darwin", "aarch64-unknown-linux-gnu", "x86_64-apple-darwin", "x86_64-unknown-linux-gnu", "x86_64-pc-windows-msvc"]
```
LiamPattinson, line 13: We're currently building for a lot more targets in the `publish.yml` workflow. Do we need to add these here too?

scripts/release.sh
```diff
@@ -0,0 +1,24 @@
...
+set -eu
+
+export UV_PREVIEW=1
+
+script_root="$(realpath "$(dirname "$0")")"
+project_root="$(dirname "$script_root")"
+
+echo "Updating metadata with rooster..."
+cd "$project_root"
+uv tool run --from 'rooster-blue>=0.0.7' --python 3.12 --isolated -- \
```
LiamPattinson, line 14: Doesn't change anything, just a nice alias that clarifies that it's something similar to `pipx`:

```suggestion
uvx --from 'rooster-blue>=0.0.7' --python 3.12 --isolated -- \
```

scripts/release.sh
```diff
@@ -0,0 +1,24 @@
...
+    rooster release "$@"
+
+echo "Updating lockfile..."
+cargo update -p fortitude
+
+echo "Generating contributors list..."
+echo ""
+echo ""
+uv tool run --from 'rooster-blue>=0.0.7' --python 3.12 --isolated -- \
```
LiamPattinson, line 23:
```suggestion
uvx --from 'rooster-blue>=0.0.7' --python 3.12 --isolated -- \
```

pyproject.toml
```diff
@@ -41,3 +41,32 @@ python-source = "python"
...
+changelog_ignore_labels = ["internal", "ci", "tests"]
+
+changelog_sections.breaking = "Breaking changes"
+changelog_sections.preview = "Preview features"
+changelog_sections.rule = "Rule changes"
+changelog_sections.diagnostics = "Rule changes"
+changelog_sections.docstring = "Rule changes"
+changelog_sections.fixes = "Rule changes"
+changelog_sections.formatter = "Formatter"
+changelog_sections.isort = "Rule changes"
```
LiamPattinson, line 58: I think this can be removed?
```suggestion
```

pyproject.toml
```diff
@@ -41,3 +41,32 @@ python-source = "python"
 python-packages = ["fortitude"]
 manifest-path = "fortitude/Cargo.toml"
 strip = true
+
+[tool.rooster]
+major_labels = []  # Ruff never uses the major version number
```
LiamPattinson, line 46: Tool agnostic comments:
```suggestion
major_labels = []  # Major version number never used
```

LiamPattinson (approved):
Thanks for this, especially for adding a detailed tutorial in `README.dev.md`! Just have a few comments and minor suggestions

LiamPattinson:
Another thing I forgot in my review: do we need the `publish.yml` workflow anymore? If that's still needed for the Python version, maybe it should be renamed `publish_python.yml` or something.

dist-workspace.toml
ZedThree: Ah, yes, but also we need to get the workflows working together to avoid duplicating building everything. I'll see how ruff does it

.github/workflows/release.yml
```diff
@@ -0,0 +1,291 @@
...
+on:
+  pull_request:
+  push:
+    tags:
+      - '**[0-9]+.[0-9]+.[0-9]+*'
+
+jobs:
+  # Run 'dist plan' (or host) to determine what tasks we need to do
+  plan:
+    runs-on: "ubuntu-20.04"
```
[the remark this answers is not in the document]
ZedThree: This file is autogenerated, so I think we just leave it alone for now

ZedThree:
There's now a `build-binaries` workflow for building the binaries once; then when a tag is pushed, `release` will make a GitHub release and add them to that, and then call `publish-pypi` to upload them to PyPI.

ZedThree:
Ok, the failing `linux-cross` builds now are due to something upstream -- astropy is having exactly the same issue, see https://github.com/uraimo/run-on-arch-action/issues/160

Seeing as it appears to be an issue with python, I'll try to use uv in the action, but if that doesn't work, I'll just skip testing the package on `linux-cross`

ZedThree:
Ok, well, we just won't test the wheels on those platforms then

LiamPattinson:
> Ok, well, we just won't test the wheels on those platforms then

We were only testing on some variant of `x86_64_manylinux` before, and I doubt we have many users on some of the less common platforms at this stage, so I'm happy to leave this for now.

PR #311, opened 2025-01-28
Add rule to flag trailing backslash

imciner2:
This is a rule to detect a simple case - a backslash is the terminating character on the line. This can be problematic if the Fortran source is passed through a C preprocessor to do preprocessing (like gfortran does), because then this will be treated as a line continuation character and the preprocessor will join the lines together. This could be problematic, especially if the backslash is in a comment line, as was shown here: https://fortran-lang.discourse.group/t/mistakes-that-will-surprise-you-the-end-of-line-backslash/8056.


imciner2:
I guess one additional change that could be made is to see if the line starts with a `#`, because then this would be valid (as well as any continued lines after this).

ZedThree:
If we parsed the AST first, and passed it into these checks, then we could check if the node at the given point was a preprocessor define. That would require a bit of restructuring so maybe best done separately

ZedThree (approved):

ZedThree:
We could also just check that we're in a comment, although again that would be easier if we had the AST already.

I ran this over some projects, and it turns out some people do use `\` to escape new lines instead of using a line continuation. I guess they must be preprocessing everything

LiamPattinson:
> I ran this over some projects, and it turns out some people do use \ to escape new lines instead of using a line continuation. I guess they must be preprocessing everything

How does `tree-sitter-fortran` handle that? If it's possible to catch it in the AST, we could auto-fix it to `&` in code and delete it in comments.

ZedThree:
> How does `tree-sitter-fortran` handle that? If it's possible to catch it in the AST, we could auto-fix it to `&` in code and delete it in comments.

Oh it doesn't like it at all

ZedThree:
I think the check for if the backslash is in a macro might be quite tricky to make robust without using the AST -- should we merge this for now, with a TODO to fix that? Or wait to get it more robust?

ZedThree:
Good point, I think that's actually the only worrying place for them too

LiamPattinson (approved):
Thanks for the update, all looks good to me

imciner2:
Circling back to this (since I was away for FOSDEM and then catching up on stuff), if the only worrying area is in the macros, since they must be on their own line and start with a `#`, then wouldn't checking for the first non-whitespace character being a `#` be a signal to ignore the line, and then if the continuation character is found also ignore the following lines (for as many are found)? I don't think that is actually too difficult to implement without the AST information actually.

LiamPattinson:
@imciner2 Apologies, when I gave approval to this PR I didn't realise that it was @ZedThree who had made the changes -- if I had been paying more attention I would have asked you first!

If you're still up for it, I would be happy for you to have another go at it as a text rule. I can't immediately see any reason why the method you described wouldn't work.

ZedThree:
> Circling back to this (since I was away for FOSDEM and then catching up on stuff), if the only worrying area is in the macros, since they must be on their own line and start with a `#`, then wouldn't checking for the first non-whitespace character being a `#` be a signal to ignore the line, and then if the continuation character is found also ignore the following lines (for as many are found)? I don't think that is actually too difficult to implement without the AST information actually.

Yes that's certainly possible, but I don't think there's any particular benefit from having it as a text rule, especially as the AST implementation ends up being very simple anyway.

LiamPattinson:
I thought the argument in favour of a text rule is that it's possible to use backslashes in place of ampersands for line continuations, and that we should forbid it everywhere except macros. As tree-sitter can't handle them, a text rule would be more robust in this case.

imciner2:
> I thought the argument in favour of a text rule is that it's possible to use backslashes in place of ampersands for line continuations, and that we should forbid it everywhere except macros. As tree-sitter can't handle them, a text rule would be more robust in this case.

Yes, in some cases it can be used instead of ampersands in some compilers. I just did an experiment with gfortran and ifort on my machine using the following snippet:

```
program t
   implicit none
   real :: A    \
   integer :: i

   ! Just a comment \
   A = 2.0     
   print *, A

end
```

When passing it through the ifort and ifx preprocessors (`ifort -fpp -E slash_end.f90`), it appears the `\` at the end of the declaration line is silently converted to an `&` continuation character:
```
# 1 "slash_end.f90"
program t
   implicit none
   real :: A    &
   integer :: i

! Just a comment \
   A = 2.0     
   print *, A

end
```

On the other hand, when passing it through the gfortran preprocessor (`gfortran -cpp -E slash_end.f90`), the lines are just concatenated together:
```
# 1 "slash_end.f90"
# 1 "<built-in>"
# 1 "<command-line>"
# 1 "slash_end.f90"
program t
   implicit none
   real :: A       integer :: i


   ! Just a comment    A = 2.0     

   print *, A

end
```

While this example won't compile with a line continuation there, it does show that it can lead to silent differences between compilers - namely under `ifort` it will behave like a standard Fortran line continuation, but gfortran could end up exceeding line limits. (also ifort/ifx isn't affected by the continuation character in a comment, but gfortran is).

PR #312, opened 2025-01-31
Decouple CLI and config file

### LINKED ISSUE #235: CLI and config file should be decoupled

ZedThree:
Closes #235 

Add `Options` struct for the config file settings, which is converted to `CheckSettings` as an intermediate "known good" config (as previously). `Options` uses the `OptionsMetadata` ruff trait to aid generating docs. This PR doesn't actually go as far as I'd like, we still do all the merging of config file and CLI settings in `check.rs`, when we could pull that out into a "final" settings struct, but I think that's a job for another day.

I've tried to keep the config file pretty much the same, but there are some keys that have been removed: for example, there's no need for `no-fix` and the other similar override switches.
`ruff` has a bunch of keys at the top-level instead of under `lint` (our `check`), for example `fix` is top-level, so are most of the file discovery settings. `exclude` _also_ appears under `lint`, and extends the top-level `exclude` for `ruff check`.

We're going to want to decide:

a) how much do we rearrange the config file settings?
b) do we retain old keys but deprecate them?
c) do we provide a config file upgrader?

I've also fiddled with the `cargo dev generate-all` command a little so that a `--mode=check` argument works more consistently and we can use it in CI to make sure we're running it more often!


.github/workflows/test.yml
```diff
@@ -22,6 +22,8 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --all-targets --all-features --verbose
+    - name: Check docs
+      run: cargo dev generate-all --mode=check
```
LiamPattinson, line 26: Thanks for this, it's something I've been meaning to do myself for ages

fortitude/src/options.rs
```diff
@@ -0,0 +1,287 @@
...
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct CheckOptions {
+    /// A list of file patterns to include when linting.
+    ///
+    /// Inclusion are based on globs, and should be single-path patterns, like
+    /// `*.pyw`, to include any file with the `.pyw` extension. `pyproject.toml` is
+    /// included here not for configuration but because we lint whether e.g. the
+    /// `[project]` matches the schema.
+    ///
+    /// Notebook files (`.ipynb` extension) are included by default on Fortitude 0.6.0+.
```
LiamPattinson, line 36: This doc-comment needs to be changed

fortitude/src/options.rs
```diff
@@ -0,0 +1,287 @@
...
+    // TODO: fix default and example when default rules decided!
+    /// A list of rule codes or prefixes to enable. Prefixes can specify exact
+    /// rules (like `T003` or `superfluous-implicit-none`), entire categories
+    /// (like `T` or `typing`), or anything in between.
+    ///
+    /// When breaking ties between enabled and disabled rules (via `select` and
+    /// `ignore`, respectively), more specific prefixes override less
+    /// specific prefixes.
+    #[option(
+        default = r#"["E4", "E7", "E9", "F"]"#,
```
LiamPattinson, line 144: I see a TODO comment here to address that this needs to be updated, but I would have thought this would break things.

fortitude/src/options.rs
```diff
@@ -0,0 +1,287 @@
...
+        example = r#"["f90", "fpp"]"#
+    )]
+    pub file_extensions: Option<Vec<String>>,
+
+    /// A list of file patterns to exclude from formatting and linting.
+    ///
+    /// Exclusions are based on globs, and can be either:
+    ///
+    /// - Single-path patterns, like `.mypy_cache` (to exclude any directory
+    ///   named `.mypy_cache` in the tree), `foo.f90` (to exclude any file named
```
LiamPattinson, line 180: Should we use something other than `.mypy_cache` here? Maybe something like `generated`, which might contain Fortran code generated at compile time?

fortitude/src/options.rs
```diff
@@ -0,0 +1,287 @@
...
+    ///   file) or `directory/*.f90` (to exclude any Fortran files in
+    ///   `directory`). Note that these paths are relative to the project root
+    ///   (e.g., the directory containing your `fpm.toml`).
+    ///
+    /// For more information on the glob syntax, refer to the [`globset` documentation](https://docs.rs/globset/latest/globset/#syntax).
+    ///
+    /// Note that you'll typically want to use
+    /// [`extend-exclude`](#extend-exclude) to modify the excluded paths.
+    #[option(
+        default = r#"[".git", ".git-rewrite", ".hg", ".svn", "venv", ".venv", "pyenv", ".pyenv", ".eggs", "site-packages", ".vscode", "build", "_build", "dist", "_dist"]"#,
```
LiamPattinson, line 192: Is this default needed? This list of default excluded paths is mirrored in `fs.rs`, but there it goes one further, as they're treated as a default that can't be overridden. Setting `exclude` here wouldn't override the stated default behaviour as it would for other options.

fortitude/src/options.rs
```diff
@@ -0,0 +1,287 @@
...
+        value_type = "bool",
+        example = r#"
+            respect-gitignore = false
+        "#
+    )]
+    pub respect_gitignore: Option<bool>,
+
+    // Global Formatting options
+    /// The line length to use when enforcing long-lines violations (like `E501`)
+    /// and at which `isort` and the formatter prefers to wrap lines.
```
LiamPattinson, line 256: This doc-comment references Ruff-specific features

fortitude/src/options.rs
```diff
@@ -0,0 +1,287 @@
...
+    ///
+    /// The length is determined by the number of characters per line, except for lines containing East Asian characters or emojis.
+    /// For these lines, the [unicode width](https://unicode.org/reports/tr11/) of each character is added up to determine the length.
```
LiamPattinson, line 259: I think we do actually count characters in `line-too-long`.

fortitude/src/options_base.rs
LiamPattinson, line 1: I have no idea what this file does, but I trust that it's doing a good job :+1:

fortitude/src/check.rs
LiamPattinson, line 1: Glad to see so much moving out of `check.rs`!

LiamPattinson (commented):
Amazing! Looks like we're now very close to our next release.

I've left some comments for things that need fixing, but it's almost entirely doc-comments copied from Ruff that need updating.

fortitude/src/options.rs
ZedThree: I'm like 95% certain the default here is _just_ for the docs. The `option` macro requires a `default` value, so we do have to supply something.

LiamPattinson: Ah, I see. I think my brain had automatically applied the filter that `#[option(...)]` was a `serde` thing.

fortitude/src/options.rs
LiamPattinson: Resolved in another comment...

fortitude/src/options_base.rs
ZedThree: This is all the stuff for options metadata -- only used by `generate-options` to pull out the various bits from the big `Options` struct and convert to markdown

ZedThree:
:facepalm: Well, now we know the "check you've rebuilt the docs" test is definitely working

LiamPattinson (approved):

PR #315, opened 2025-01-31
Fix some issues with newlines in fixes on Windows

ZedThree:
When I deduplicated building the binaries, it also stopped us building on different platforms.

And it looks like we were never actually testing on those platforms!

Windows failures seem pretty simple: paths are using `\` instead of `/` as expected in the snapshots. Can probably do something simple to fix those in the tests.

Mac failures more puzzling, looks like `per-file-ignores` not working quite as expected?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,11 +4,19 @@ on:
   push:
     paths:
       - '**.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - '.github/workflows/test.yml'
   pull_request:
     paths:
       - '**.rs'
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      - '.github/workflows/test.yml'
+  workflow_dispatch:
 
 env:
+  CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always
 
 jobs:
@@ -16,6 +24,26 @@ jobs:
   build:
     name: Build and Test
     runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --all-targets --all-features --verbose
+
+  build-windows:
+    name: Build and Test (windows)
+    runs-on: windows-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build
+      run: cargo build --verbose
+    - name: Run tests
+      run: cargo test --all-targets --all-features --verbose
+
+  build-macos:
+    name: Build and Test (macos)
+    runs-on: macos-latest
     steps:
     - uses: actions/checkout@v4
     - name: Build
--- a/fortitude/src/ast.rs
+++ b/fortitude/src/ast.rs
@@ -212,7 +212,7 @@ impl FortitudeNode<'_> for Node<'_> {
 
 /// Strip line breaks from a string of Fortran code.
 pub fn strip_line_breaks(src: &str) -> String {
-    src.replace('&', "").replace('\n', " ")
+    src.replace('&', "").replace('\n', " ").replace('\r', "")
 }
 
 /// Returns true if the type passed to it is number-like, and of a kind that can be modified using
--- a/fortitude/src/rules/bugprone/mod.rs
+++ b/fortitude/src/rules/bugprone/mod.rs
@@ -9,6 +9,7 @@ mod tests {
     use insta::assert_snapshot;
     use test_case::test_case;
 
+    use crate::apply_common_filters;
     use crate::registry::Rule;
     use crate::settings::Settings;
     use crate::test::test_path;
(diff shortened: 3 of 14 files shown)


LiamPattinson:
For the Mac problem, perhaps it's this line:

https://github.com/PlasmaFAIR/fortitude/blob/ea2d6b8c944f5b9b0b08c6e8cbe2d0f520bea0ed/fortitude/tests/check.rs#L699

We're using `.current_dir(path)` and also `.arg(path)`, which might be confusing things. I think removing `.arg(path)` might fix it, but without a Mac in my hands I can't really check.

ZedThree:
> We're using `.current_dir(path)` and also `.arg(path)`, which might be confusing things. I think removing `.arg(path)` might fix it, but without a Mac in my hands I can't really check.

Nice, good catch!

Failures on Windows now are due to line-ending handling -- I think these are actual bugs, not just issues in the tests

ZedThree:
Hmm remaining failure is this:

```
---- rules::style::tests::rules::rule_incorrectspacebeforecomment_path_new_s102_f90_expects stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot file: fortitude\src\rules\style\snapshots\fortitude__rules__style__tests__incorrect-space-before-comment_S102.f90.snap
Snapshot: incorrect-space-before-comment_S102.f90
Source: D:\a\fortitude\fortitude:42
───────────────────────────────────────────────────────────────────────────────
Expression: diagnostics
───────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬──────────────────────────────────────────────────────────────────
   11    11 │ ℹ Safe fix␊
   12    12 │ 1 1 | ! This is fine␊
   13    13 │ 2 2 |  ! This isn't but we'll let it pass␊
   14    14 │ 3   |-module mymod! This should be given two extra spaces␊
   15       │-  3 |+module mymod  ! This should be given two extra spaces␊
   16       │-4 4 |   implicit none ! This should be given one extra space␊
   17       │-5 5 |   private  ! This is fine␊
   18       │-6 6 |   ! This is fine␊
         15 │+  3 |+module mymod  ! This should be given two extra spaces␍
         16 │+  4 |+␊
         17 │+4 5 |   implicit none ! This should be given one extra space␊
         18 │+5 6 |   private  ! This is fine␊
         19 │+6 7 |   ! This is fine␊
   19    20 │ ␊
   20    21 │ ./resources/test/fixtures/style/S102.f90:4:17: S102 [*] need at least 2 spaces before inline comment␊
   21    22 │   |␊
   22    23 │ 2 |  ! This isn't but we'll let it pass␊
┈┈┈┈┈┈┈┈┈┈┈┈┼┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
   32    33 │ 1 1 | ! This is fine␊
   33    34 │ 2 2 |  ! This isn't but we'll let it pass␊
   34    35 │ 3 3 | module mymod! This should be given two extra spaces␊
   35    36 │ 4   |-  implicit none ! This should be given one extra space␊
   36       │-  4 |+  implicit none  ! This should be given one extra space␊
   37       │-5 5 |   private  ! This is fine␊
   38       │-6 6 |   ! This is fine␊
   39       │-7 7 | ! As is this
         37 │+  4 |+  implicit none  ! This should be given one extra space␍
         38 │+  5 |+␊
         39 │+5 6 |   private  ! This is fine␊
         40 │+6 7 |   ! This is fine␊
         41 │+7 8 | ! As is this
────────────┴──────────────────────────────────────────────────────────────────
```

and I cannot see anything obvious. I would've guessed that this could happen if we were writing a newline, but I'm reasonably sure we're explicitly _not_.

It might be quite tricky to debug this without a Windows env!

LiamPattinson:
That's really strange. Am I reading it correctly in that it it's adding an extra newline in Windows?

ZedThree:
Sort of -- it's changing the existing line ending from `\n` (unix) to `\r\n` (DOS line ending). The characters in the output are really small, but you can see `CR`/`LF`.

My understanding is that we're replacing the comment node upto but not including the line ending, so we shouldn't be writing one back out. I don't think we explicitly write new lines _anywhere_?

LiamPattinson:
I think this might be similar to an issue we saw with `end module` etc, where the node included a newline character. Oddly, I think tree-sitter was adding a trailing `\r` via `utf8_text`. I wasn't able to completely figure out what was going wrong and why, but I was able to fix it up by simplifying the edit. I also fixed up a couple other Windows issues, and I'll raise a PR once it's all finalised.

ZedThree:
Nice, thanks for the extra fixes! The insertion is much simpler.

I've investigated further, and the tree-sitter grammar consumes the `\r` character into comments -- that might be something we have to watch out for in future. Or work out how to fix it in the grammar.

LiamPattinson (approved):

PR #319, opened 2025-02-05
Add `Settings` struct for final resolved settings

ZedThree:
We now have several types for the input settings, they start at the bottom with:

- `Cli` and `Options` which represent the inputs
- `Configuration` as an intermediate step, representing the config file after reading and applying some defaults
- `Settings` representing the finalised settings, after overriding using the cli

This mirrors ruff, although their `Configuration` is more complicated and can be built up out of many files.

The point of this is to both a) tidy up `check.rs` even further, and b) give us one struct we can pass around, allowing us to remove a few repeated function arguments.

We _should_ be able to get rid of `rules` as a separate variable as well, except we currently still need to split it up into `path_rules` etc. We could probably rejig that now, but that's a separate PR.

The last two commits add `--show-settings` and `--show-files`, they should maybe go in separate PRs. They are sort of related, in that they're made much easier to implement by having this new one-true-struct.

### DIFF
--- a/fortitude/src/cli.rs
+++ b/fortitude/src/cli.rs
@@ -165,6 +165,10 @@ pub struct CheckArgs {
     #[arg(long, value_enum)]
     pub progress_bar: Option<ProgressBar>,
 
+    /// See the settings fortitude will use to check a given Fortran file.
+    #[arg(long)]
+    pub show_settings: bool,
+
     // Rule selection
     /// Comma-separated list of rules to ignore.
     #[arg(
--- a/fortitude/src/cli.rs
+++ b/fortitude/src/cli.rs
@@ -166,8 +166,19 @@ pub struct CheckArgs {
     pub progress_bar: Option<ProgressBar>,
 
     /// See the settings fortitude will use to check a given Fortran file.
-    #[arg(long)]
+    #[arg(long,
+        // Fake subcommands.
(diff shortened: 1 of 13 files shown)


fortitude/src/cli.rs
```diff
@@ -165,6 +165,21 @@ pub struct CheckArgs {
     #[arg(long, value_enum)]
     pub progress_bar: Option<ProgressBar>,
 
+    /// See the settings fortitude will use to check a given Fortran file.
+    #[arg(long,
+        // Fake subcommands.
+        conflicts_with = "show_files",
+        // conflicts_with = "show_settings",
+    )]
+    pub show_settings: bool,
```
LiamPattinson, line 174: I'm a bit confused with the commented out lines here.

LiamPattinson (approved):
All looks good to me. Just had one comment.

fortitude/src/cli.rs
LiamPattinson: There are others for the `show_files` setting too.

LiamPattinson: It's your call. I'd be happy to merge them here, but it might be better for the changelog to add them separately.

ZedThree: I'm just going to merge this now, and we'll have to remember to manually add those flags to the changelog!

PR #323, opened 2025-02-14
Proposal: Hardcode list of default rules

ZedThree:
Here's another proposal for updating the default rules. Here, I've done two things:

1. hardcode the list of rules (and some prefixes)
2. gone more inclusive with what's turned on by default -- anything not "too" pedantic (probably some of the ones I've missed could also be default)

My issues with this version are:

1. tricky to communicate what's on by default
2. actually really quite ugly

The reason for 2 is that `DEFAULT_SELECTORS` is a `const` list of `RuleSelector`, and there's no straightforward conversion from `Rule` to `RuleSelector`, hence the ugliness. This could be improved if we used `lazy_static` and then we could just use `RuleSelector::from_str("rule-name")`, which would be both a _lot_ simpler and also much easier to read.

@LiamPattinson previously suggested having something in the `map_codes` macro to determine which rules were default, and I'd pushed back on the basis I thought this method was going to be fairly simple. As it's not, we could add something like `Default`/`Pedantic` to `map_codes`, which we could use to 

1. build this list
2. put something indicating whether or not it's on by default into the docs and `explain`. The config docstring would then just say "see the docs for which rules are on by default"

### DIFF
--- a/fortitude/src/settings.rs
+++ b/fortitude/src/settings.rs
@@ -12,10 +12,12 @@ use ruff_diagnostics::Applicability;
 use ruff_macros::CacheKey;
 use serde::{de, Deserialize, Deserializer, Serialize};
 
-use crate::display_settings;
 use crate::fs::{FilePatternSet, EXCLUDE_BUILTINS, FORTRAN_EXTS};
+use crate::registry::Category;
 use crate::rule_selector::{CompiledPerFileIgnoreList, PreviewOptions, RuleSelector};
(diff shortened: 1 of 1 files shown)


LiamPattinson:
Yeah, I think I'm still leaning towards the `map_codes` change, largely just so that we can have a column in the docs saying "Default: :white_check_mark:", although I realise that it will require quite a lot of fiddling around with macros.

PR #324, opened 2025-02-14
Mark rules as `Default` or `Optional` in `map_codes`

### LINKED ISSUE #306: Default rule list needed

ZedThree:
Closes #306 

This solution adds a marker `Default` or `Pedantic` to the `map_codes` macro, and a method `Rule::is_default()` which is used to lazily create `DEFAULT_SELECTORS` -- we could probably do this statically if I put in some more effort, but it's Friday afternoon.

The individual rule docs only say if they're turned on by default, and not if they're turned off.

Now that I've implemented it, I've come around to thinking that this is maybe actually the easiest method to just have a few rules be too pedantic to be on by default. Exactly which rules those are is almost a separate discussion, of course.


fortitude/src/rules/mod.rs
```diff
@@ -74,92 +74,92 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Filesystem, "001") => (RuleGroup::Stable, Path, Default,filesystem::extensions::NonStandardFileExtension),
+
+        (Style, "001") => (RuleGroup::Stable, Text, Default, style::line_length::LineTooLong),
+        (Style, "021") => (RuleGroup::Stable, Ast, Default, style::exit_labels::MissingExitOrCycleLabel),
+        (Style, "041") => (RuleGroup::Stable, Ast, Default, style::old_style_array_literal::OldStyleArrayLiteral),
+        (Style, "051") => (RuleGroup::Stable, Ast, Default, style::relational_operators::DeprecatedRelationalOperator),
+        (Style, "061") => (RuleGroup::Stable, Ast, Default, style::end_statements::UnnamedEndStatement),
+        (Style, "071") => (RuleGroup::Stable, Ast, Default, style::double_colon_in_decl::MissingDoubleColon),
+        (Style, "081") => (RuleGroup::Preview, Ast, Default, style::semicolons::SuperfluousSemicolon),
+        (Style, "082") => (RuleGroup::Preview, Ast, Default, style::semicolons::MultipleStatementsPerLine),
```
LiamPattinson, line 90: I'm not sure if this one should be on by default. `superfluous-semicolon` should definitely be on though.

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +74,92 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Style, "101") => (RuleGroup::Stable, Text, Default, style::whitespace::TrailingWhitespace),
+        (Style, "102") => (RuleGroup::Stable, Ast, Pedantic, style::whitespace::IncorrectSpaceBeforeComment),
+
+        (Typing, "001") => (RuleGroup::Stable, Ast, Default, typing::implicit_typing::ImplicitTyping),
+        (Typing, "002") => (RuleGroup::Stable, Ast, Default, typing::implicit_typing::InterfaceImplicitTyping),
+        (Typing, "003") => (RuleGroup::Stable, Ast, Default, typing::implicit_typing::SuperfluousImplicitNone),
+        (Typing, "004") => (RuleGroup::Preview, Ast, Default, typing::implicit_typing::ImplicitExternalProcedures),
```
LiamPattinson, line 97: I'm not sure about this one. I think it's a good idea to do this, but I think it's rare to see people do it in practice. Maybe it's okay as long as it's kept in preview?

fortitude/src/settings.rs
```diff
@@ -177,8 +180,19 @@ impl fmt::Display for PreviewMode {
     }
 }
 
-/// Default rule selection
-pub const DEFAULT_SELECTORS: &[RuleSelector] = &[RuleSelector::All];
+lazy_static! {
+    /// Default rule selection
+    pub static ref DEFAULT_SELECTORS: Vec<RuleSelector> = {
```
LiamPattinson, line 185: I think it's fine to do this lazily for now. With fewer than 50 default rules, I imagine the performance implications would barely even be measurable.

LiamPattinson (approved):
Cheers for this! I would lean towards renaming `Pedantic` to `Optional`, as the former feels like it has negative connotations, but otherwise the only comments I have are some minor disagreements regarding what should or shouldn't be on by default (and even for these I don't feel particularly strongly one way or the other).

fortitude/src/rules/mod.rs
ZedThree: I think it's rare to see because people don't know about it -- which I think is a good argument for us having it on by default! It's reasonably widely supported, basically just nvfortran and amd-flang/aocc that don't. Sensible to keep it in preview for now though

fortitude/src/settings.rs
ZedThree: Just timed this: `Built default selectors in: 107.602µs`. Probably fast enough!

ZedThree:
@imciner2 @Beliavsky You've both made significant contributions, do you have any strong feelings about which rules should or should not be on by default?

PR #325, opened 2025-02-14
Bump ruff packages to 0.9.6

ZedThree:
Bumping all the ruff packages to the latest ruff so that we don't drift too far from them

The only breaking changes are:

1. `Locator` is no longer in `ruff_source_file` but `ruff_linter`, which we don't want to use, so we need to vendor it completely. We actually don't use most of it currently, but there's definitely a few bits we will want to use (the index stuff is useful for LSP it seems)
2. `#[violation]` is now `#[derive(ViolationMetadata)]`
3. `Violation::mesage` now really hates `format!()` that doesn't actually format any arguments, so needs to be `"".to_string()` instead

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -824,9 +824,9 @@ dependencies = [
 
 [[package]]
 name = "itertools"
-version = "0.13.0"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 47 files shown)


LiamPattinson (approved):
Thanks for this! It's probably good for us to have our own version of `Locator`, as we might have to make some changes when we add preprocessor support (although keeping it up to date with Ruff would then be harder).

PR #329, opened 2025-02-19
Recategorise rules

LiamPattinson:
Builds on https://github.com/PlasmaFAIR/fortitude/pull/324

Implementing discussion in https://github.com/PlasmaFAIR/fortitude/issues/269#issuecomment-2659142255.

Aiming to move all rules into new categories, while also introducing new redirect logic to ensure old settings files won't break. This isn't the prettiest, so I'm hoping we can remove this in version 0.8 or 0.9.

To begin, only renamed `bugprone` to `correctness`. I'll be churning through the other rules one deprecated category at a time.

One issue is the redirecting of partial rule codes, e.g. `B0` or `B01`. Should we just accept that those will break, or should they also be included in the new redirects?


LiamPattinson:
I'm not sure why it's saying there's a merge conflict, as if I try merging locally it works okay.

LiamPattinson:
I think I'm finally there. Let me know if you think I've moved something into the wrong category, as there's still room to move things around.

One thing I'm tempted to do is play around with the numbers in the `correctness` category, as we currently have the `implicit none` rules at `C05`. As they're are often used as example rules, it would be nicer if we could have those at `C00`.

fortitude/src/rule_selector.rs
```diff
@@ -129,13 +152,17 @@ pub enum ParseError {
 }
 
 impl RuleSelector {
-    pub fn prefix_and_code(&self) -> (&'static str, &'static str) {
+    pub fn prefix_and_code(&self) -> (String, String) {
```
ZedThree, line 155: Not sure I understand the reason for this change?

ZedThree (approved):
I broadly agree with this recategorisation, with some quibbles on some individual rules, although I could probably live with them!

One other big internal change we might consider: just put all the individual rule tests into a single directory, and name the `.f90` files after the rule name rather than code? Maybe we can also rejig the snapshots to be along side them too. Possibly for another PR?

ZedThree:
I also don't understand where the merge conflict has come from. Maybe we need to merge #324 first?

@imciner2 @Beliavsky Any opinions on the new categories?

fortitude/src/rule_selector.rs
LiamPattinson: The prefixes and codes it returns are macro-generated via `map_codes` and the `RuleNamespace` trait, so they're defined at compile time. To support the deprecated categories, we need to instead return whatever string the user provided.

There's probably a better way to do it by diving into the macros, but I thought that might be overkill when we really just want to redirect everything.

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +70,89 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        
+        (Portability, "001") => (RuleGroup::Preview, Ast, Optional, portability::magic_io_unit::NonPortableIoUnit),
+        (Portability, "011") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKind),
+        (Portability, "012") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKindSuffix),
+        (Portability, "021") => (RuleGroup::Stable, Ast, Default, portability::star_kinds::StarKind),
+
+        (Style, "001") => (RuleGroup::Stable, Text, Default, style::line_length::LineTooLong),
+        (Style, "021") => (RuleGroup::Stable, Ast, Default, style::exit_labels::MissingExitOrCycleLabel),
+        (Style, "041") => (RuleGroup::Stable, Ast, Default, style::old_style_array_literal::OldStyleArrayLiteral),
+        (Style, "051") => (RuleGroup::Stable, Ast, Default, style::relational_operators::DeprecatedRelationalOperator),
```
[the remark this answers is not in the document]
LiamPattinson: Good point, I didn't think to move anything out of the style category.

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +70,89 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Correctness, "091") => (RuleGroup::Stable, Ast, Default, correctness::external::ExternalProcedure),
+        (Correctness, "101") => (RuleGroup::Preview, Ast, Default, correctness::derived_default_init::MissingDefaultPointerInitalisation),
+        (Correctness, "111") => (RuleGroup::Stable, Ast, Default, correctness::external_functions::ProcedureNotInModule),
+        (Correctness, "121") => (RuleGroup::Stable, Ast, Default, correctness::use_statements::UseAll),
+        (Correctness, "122") => (RuleGroup::Preview, Ast, Default, correctness::use_statements::MissingIntrinsic),
+        (Correctness, "131") => (RuleGroup::Preview, Ast, Default, correctness::accessibility_statements::MissingAccessibilityStatement),
+        (Correctness, "132") => (RuleGroup::Preview, Ast, Optional, correctness::accessibility_statements::DefaultPublicAccessibility),
+        (Correctness, "141") => (RuleGroup::Preview, Ast, Optional, correctness::include_statement::IncludeStatement),
+        
+        (Modernization, "001") => (RuleGroup::Stable, Ast, Optional, modernization::double_precision::DoublePrecision),
```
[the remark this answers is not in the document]
LiamPattinson: I did feel a small part of my soul die when I pressed that 'z' key.

I assumed that the US spelling would be considered the international standard, but apparently British English is more widely taught. I'll fix it up.

LiamPattinson:
> One other big internal change we might consider: just put all the individual rule tests into a single directory, and name the `.f90` files after the rule name rather than code? Maybe we can also rejig the snapshots to be along side them too. Possibly for another PR?

I would support this, but perhaps that can follow in its own PR.

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +70,89 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Correctness, "081") => (RuleGroup::Stable, Ast, Default, correctness::init_decls::InitialisationInDeclaration),
+        (Correctness, "091") => (RuleGroup::Stable, Ast, Default, correctness::external::ExternalProcedure),
+        (Correctness, "101") => (RuleGroup::Preview, Ast, Default, correctness::derived_default_init::MissingDefaultPointerInitalisation),
+        (Correctness, "111") => (RuleGroup::Stable, Ast, Default, correctness::external_functions::ProcedureNotInModule),
+        (Correctness, "121") => (RuleGroup::Stable, Ast, Default, correctness::use_statements::UseAll),
+        (Correctness, "122") => (RuleGroup::Preview, Ast, Default, correctness::use_statements::MissingIntrinsic),
+        (Correctness, "131") => (RuleGroup::Preview, Ast, Default, correctness::accessibility_statements::MissingAccessibilityStatement),
+        (Correctness, "132") => (RuleGroup::Preview, Ast, Optional, correctness::accessibility_statements::DefaultPublicAccessibility),
+        (Correctness, "141") => (RuleGroup::Preview, Ast, Optional, correctness::include_statement::IncludeStatement),
```
imciner2, line 98: Perhaps better under modernization? MFE considers this more of a "deprecated" feature.

(I do see the argument for how it can be a correctness issue, but it does feel more like updating to use newer features of the language compared to just being a bug-prone thing.)

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +70,89 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Style, "051") => (RuleGroup::Stable, Ast, Default, style::relational_operators::DeprecatedRelationalOperator),
+        (Style, "061") => (RuleGroup::Stable, Ast, Default, style::end_statements::UnnamedEndStatement),
+        (Style, "071") => (RuleGroup::Stable, Ast, Default, style::double_colon_in_decl::MissingDoubleColon),
+        (Style, "081") => (RuleGroup::Preview, Ast, Default, style::semicolons::SuperfluousSemicolon),
+        (Style, "082") => (RuleGroup::Preview, Ast, Optional, style::semicolons::MultipleStatementsPerLine),
+        (Style, "091") => (RuleGroup::Stable, Path, Default, style::file_extensions::NonStandardFileExtension),
+        // There are likely to be many whitespace rules at some point, reserve S1xx for them
+        (Style, "101") => (RuleGroup::Stable, Text, Default, style::whitespace::TrailingWhitespace),
+        (Style, "102") => (RuleGroup::Stable, Ast, Optional, style::whitespace::IncorrectSpaceBeforeComment),
+        (Style, "201") => (RuleGroup::Stable, Ast, Default, style::implicit_none::SuperfluousImplicitNone),
```
imciner2, line 119: I don't know if this is as useful as a default rule. The missing implicit none rule is definitely one to be on by default, but a superfluous statement doesn't seem like it is bad enough to justify being on by default.

ZedThree: It is a `style` rule though, and as long as we're upfront that these are opinionated, that's maybe the justification? I'm mildly in favour of keeping it on my default, but not opposed to off by default.

LiamPattinson: I wouldn't really mind either way.

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +70,89 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Portability, "011") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKind),
+        (Portability, "012") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKindSuffix),
+        (Portability, "021") => (RuleGroup::Stable, Ast, Default, portability::star_kinds::StarKind),
+
+        (Style, "001") => (RuleGroup::Stable, Text, Default, style::line_length::LineTooLong),
+        (Style, "021") => (RuleGroup::Stable, Ast, Default, style::exit_labels::MissingExitOrCycleLabel),
+        (Style, "041") => (RuleGroup::Stable, Ast, Default, style::old_style_array_literal::OldStyleArrayLiteral),
+        (Style, "051") => (RuleGroup::Stable, Ast, Default, style::relational_operators::DeprecatedRelationalOperator),
+        (Style, "061") => (RuleGroup::Stable, Ast, Default, style::end_statements::UnnamedEndStatement),
+        (Style, "071") => (RuleGroup::Stable, Ast, Default, style::double_colon_in_decl::MissingDoubleColon),
```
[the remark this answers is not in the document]
LiamPattinson: Do you mean `double-colon-in-decl`? I would argue that one fits better in `style` personally, as I see it more about maintaining consistency than about doing it the modern way.

fortitude/src/rules/mod.rs
```diff
@@ -74,92 +70,89 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
...
+        (Portability, "001") => (RuleGroup::Preview, Ast, Optional, portability::magic_io_unit::NonPortableIoUnit),
+        (Portability, "011") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKind),
+        (Portability, "012") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKindSuffix),
+        (Portability, "021") => (RuleGroup::Stable, Ast, Default, portability::star_kinds::StarKind),
+
+        (Style, "001") => (RuleGroup::Stable, Text, Default, style::line_length::LineTooLong),
+        (Style, "021") => (RuleGroup::Stable, Ast, Default, style::exit_labels::MissingExitOrCycleLabel),
+        (Style, "041") => (RuleGroup::Stable, Ast, Default, style::old_style_array_literal::OldStyleArrayLiteral),
+        (Style, "051") => (RuleGroup::Stable, Ast, Default, style::relational_operators::DeprecatedRelationalOperator),
+        (Style, "061") => (RuleGroup::Stable, Ast, Default, style::end_statements::UnnamedEndStatement),
```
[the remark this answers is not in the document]
LiamPattinson: Is there a case where an end statement might be ambiguous? I think I see this as more of a style thing.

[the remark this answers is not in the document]
imciner2: I was wondering that, and I don't think there is a problem for code correctness really, the problem would be for understanding of the code if it is badly formatted. But I agree, style seems more appropriate than correctness for this rule.

fortitude/src/rules/mod.rs
LiamPattinson: I had a think about it and decided to set it to optional. I don't see it as a particularly severe style violation, and I know some style guides recommend putting it everywhere (I think it's to make sure you're protected in case you ever decide to move a function into a module that deliberately isn't using `implicit none` for whatever reason).

LiamPattinson:
I'm now getting really messy merge/rebase conflicts when I try to catch up to main, but I can't figure out why. All of my commits sit on top of the latest one, so I can't understand how there could possibly be a conflict.

ZedThree:
Ah, it's because this doesn't include #325

LiamPattinson:
Yeah, just came to the same conclusion. This is going to take me a while...

LiamPattinson:
@ZedThree Thanks for helping out with the rebase! Is there anything else we need to do for this?

I shuffled around some of the numbers in the `correctness` category, as I wanted `C001` to be `implicit-none` (it's a really good rule for explaining what Fortitude does, so I think it makes sense to have it at the top of the list with an easily memorable number). Any other re-numbering I should do while I have the chance? We probably shouldn't touch established `style` or `obsolescent` rules, but the rest can be moved around freely at this stage.

ZedThree:
`ExternalProcedure` is somewhat related to `ProcedureNotInModule` in that they both result in procedures without explicit interfaces, so they could share a prefix?

Other than that, I think this is great and good to go!

LiamPattinson:
> ExternalProcedure is somewhat related to ProcedureNotInModule in that they both result in procedures without explicit interfaces, so they could share a prefix?

They were also defined in files with very similar names, so I've numbered them to be both under `C09` and moved them both into `external.f90`.

fortitude/src/rules/style/implicit_none.rs
```diff
@@ -0,0 +1,71 @@
...
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for unnecessary `implicit none` in module procedures
+///
+/// ## Why is this bad?
+/// If a module has 'implicit none' set, it is not necessary to set it in contained
+/// functions and subroutines (except when using interfaces).
+#[derive(ViolationMetadata)]
+pub struct SuperfluousImplicitNone {
```
imciner2, line 20: I just noticed, the code for `SuperfluousImplicitNone` also still exists in the correctness `implicit_typing` file on the current main branch after this PR has been merged.

LiamPattinson: Nice catch! I'm surprised that didn't cause any issues. Feel free to remove it in a new PR if you have the time.

PR #334, opened 2025-02-26
Add rules for unused/duplicated/redirected allow comments

### LINKED ISSUE #252: Better warnings/errors for allow-comments

ZedThree:
Closes #252 

Unfortunately, these changes do mean we lose the ability to `allow` whole categories or prefixes. I do think it's wise to not actually do that in practice, but maybe it's useful occasionally -- if we manage to reimplement that, then we probably also want to add a corresponding warning.

Moves all the allow-comment related rules to a new `Fortitude/FORT` category, as well as the test rules.

Also adds `--ignore-allow-comments` to the CLI to temporarily ignore them completely.


fortitude/src/rules/fortitude/allow_comments.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+/// ## What it does
+/// Checks for allow comments that aren't applicable.
+///
+/// ## Why is this bad?
+/// Probably a mistake
+#[derive(ViolationMetadata)]
+pub(crate) struct DuplicatedAllowComment {
```
LiamPattinson, line 87: Doc-comment is the same as the one above
```suggestion
/// ## What it does
/// Checks for duplicated allow comments.
///
/// ## Why is this bad?
/// Probably a mistake
#[derive(ViolationMetadata)]
pub(crate) struct DuplicatedAllowComment {
```

fortitude/src/rules/fortitude/allow_comments.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+/// ## What it does
+/// Checks for allow comments that aren't applicable.
+///
+/// ## Why is this bad?
+/// Probably a mistake
+#[derive(ViolationMetadata)]
+pub(crate) struct DisabledAllowComment {
```
LiamPattinson, line 109: Same again
```suggestion
/// ## What it does
/// Checks for allow comments that refer to disabled rules.
///
/// ## Why is this bad?
/// Probably a mistake
#[derive(ViolationMetadata)]
pub(crate) struct DisabledAllowComment {
```

fortitude/src/rules/fortitude/allow_comments.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    fn fix_title(&self) -> String {
+        let Self {
+            new_code, new_name, ..
+        } = self;
+        format!("Replace with `{new_code}` or `{new_name}`")
+    }
+}
+
+// fn check_redirected_codes(diagnostics: &mut Vec<Diagnostics>, codes:
```
LiamPattinson, line 57: Left some commented-out code in

LiamPattinson (approved):
All looks good except for some minor docs/comments issues!

Issue #333, opened 2025-02-25
Don't raise or apply fixes if syntax errors


ZedThree:

LiamPattinson:
I'm having a go at this, but I'm unsure what the desired behaviour should be. We know that syntax errors can result in false positives elsewhere and so fixes should definitely be disabled, but what about raising other violations?

My current strategy is to only raise AST violations if the user provides the option `--allow-syntax-errors` (or something like that, I'm not sure about the name).
- By default, if the `syntax-error` rule is enabled and syntax errors are detected, then only `syntax-error` violations will be raised.
- If `--allow-syntax-errors` is activated, also raise the other AST violations.
- If the `syntax-error` rule is disabled and `--allow-syntax-errors` in on, then raise AST violations as usual.
- If the `syntax-error` rule is disabled and `--allow-syntax-errors` is off, then raise no AST violations and instead print a warning.
- Regardless of whether `--allow-syntax-errors` is on, always forbid applying fixes.

Is that sensible? Some other options are:

- Always raise AST rule violations even when syntax errors are present, but forbid fixes. Provide no option to configure this behaviour.
- Do the same as I've suggested above, and also add an option to allow fixes when syntax errors are present.

LiamPattinson:
Gave it some further thought, and maybe the `--allow-syntax-errors` option isn't needed, as `--exclude=syntax-error` could do the same job.

ZedThree:
I think keeping it just depending on the `syntax-error` rule is best.

The simplest, dumbest thing I had thought of doing was:

```diff
modified   fortitude/src/check.rs
@@ -236,6 +236,9 @@ pub(crate) fn check_path(
         if let Some(allow_rules) = gather_allow_comments(&node, file) {
             allow_comments.push(allow_rules);
         };
+        if node.is_error() && rules.enabled(Rule::SyntaxError) {
+            break;
+        }
     }
```

basically, raise the first `syntax-error` and then bail on any more checks. That's maybe reasonable, except it breaks the semicolon tests because of this:

```f90
program foo
   ;  ! <- syntax error
implicit none;
```

Turns out I'd only allowed solitary `;` in the executable statements section in the grammar. Oops.

It also doesn't account for `allow` comments.

> By default, if the syntax-error rule is enabled and syntax errors are detected, then only syntax-error violations will be raised.

[...]

PR #336, opened 2025-02-27
Don't raise linter violations or apply fixes when syntax errors are detected

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/333

As suggested by @ZedThree, this could be altered so that linter violations are reported up to and including the first `syntax-error`. I'm currently playing around to see if there are any counter-examples that means a rule could be checking beyond a syntax error and return bad data as a result.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -293,17 +293,28 @@ pub(crate) fn check_path(
     // after it, as they may be false positives.
     if rules.enabled(Rule::SyntaxError) && root.has_error() {
         warn_user_once_by_message!(
-            "Syntax errors detected in file: {}. Discarding subsequent violations.",
+            "Syntax errors detected in file: {}. Discarding subsequent violations from the AST.",
             path.to_string_lossy()
         );
         // Sort by byte-offset in the file
         violations.sort_by_key(|diagnostic| diagnostic.range.start());
         // Retain all violations up to the first syntax error, inclusive.
+        // Text and path rules can be safely retained.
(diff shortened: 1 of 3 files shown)


LiamPattinson:
I've updated it to only remove violations after the first syntax-error. I'm worried that this could cause unexpected behaviour, as any rule that tries to look ahead could still report a faulty violation if there's a syntax error between the entrypoint and the node from which a diagnostic is raised. Still looking to see if there are any counter-examples.

ZedThree:
We might need to sort the violations by range first, as I see things like:

```
warning: Syntax errors detected in file: /home/peter/Tools/tree-sitter/fortran_examples/WRF/var/obsproc/src/setup.F90. Discarding subsequent violations.
var/obsproc/src/setup.F90:1:7: C092 subroutine not contained within (sub)module or program
var/obsproc/src/setup.F90:2:59: E001 Syntax error
var/obsproc/src/setup.F90:97:16: C061 subroutine argument 'domain_check_h' missing 'intent' attribute
```

Here the last violation is after the syntax error.

We also only _really_ need to remove the AST rules, as the text ones will still be valid, but maybe it's more consistent to remove everything after the first syntax error.

LiamPattinson:
Can we be confident that the presence of a syntax error will only ever cause AST errors at later points in the file?

LiamPattinson:
Turns out the required macro edits really weren't too difficult, so I now have it discarding all subsequent AST violations by byte offset, but retaining all violations of other types.

PR #335, opened 2025-02-26
Bump version to 0.7.0

ZedThree:

PR #337, opened 2025-02-27
Move rules from preview

LiamPattinson:
In preparation for https://github.com/PlasmaFAIR/fortitude/pull/335, moves the following rules out of `--preview` mode:

- `missing-accessibility-statement`
- `default-public-accessibility`
- `implicit-external-procedures`

`implicit-external-procedures` has also been made an optional rule, mostly because it's a Fortran 2018 feature that some users won't be able to use (although I'd hope by now that most people would have migrated to newer compilers).

Entirely unrelated, but I also deleted the very old `test.f90` file sitting in the root directory, as I haven't needed to use it for months and it's cluttering up the repo.


ZedThree:
I think these were all added since the last release, so maybe best keep them in preview for now? Or do you think they've been tested enough already?

I'd also quite like to keep `implicit-external-procedures` as on by default, the majority of compilers support it, including gcc since 5.1 (2015!)

LiamPattinson:
I picked those rules because they were the three preview rules added for v0.6.0. I thought we should move at least a few rules out of preview, as otherwise the overall effect of this release for the average user would just be to take away rules that were previously on by default.

I've toggled `implicit-external-procedures` to default, but I'm still a bit uneasy about it. I guess we'll see if anybody raises any issues about it.

ZedThree:
Ah, sorry, I thought they were added _after_ 0.6.0 was released, apologies! In that case, I definitely support moving them out of preview

ZedThree (approved):

Issue #338, opened 2025-02-28
False positives on statement functions


ZedThree:
Because the statement function syntax is completely ambiguous with an ordinary assignment statement, it's not actually possible to tell them apart during parsing. Instead, one has to work out the types and attributes of the identifiers afterwards. Here's a couple of examples from lfortran:

```f90
    ! Example of statement functions:
    integer :: A
    A(i,j) = i*j
    ! implicit typing on
    A(i,j) = i*j
    ! Examples of non statement functions:
    integer :: A(3, 5)
    A(i,j) = i*j
```

Here's an example of tree-sitter flip-flopping:

```f90
  ! Too many assignment expressions and the first one gets parsed as a
  ! statement function
  SUBROUTINE mpi_error_handler(comm, error_code)
    CHARACTER(LEN=29) :: errstring(0:MPI_ERR_LASTCODE)

    ! this gets parsed as `statement_function`, unless just one line is removed
    errstring(MPI_SUCCESS                  ) = 'MPI_SUCCESS                  '
    errstring(MPI_ERR_BUFFER               ) = 'MPI_ERR_BUFFER               '
    errstring(MPI_ERR_COUNT                ) = 'MPI_ERR_COUNT                '
    errstring(MPI_ERR_TYPE                 ) = 'MPI_ERR_TYPE                 '
    errstring(MPI_ERR_TAG                  ) = 'MPI_ERR_TAG                  '
    errstring(MPI_ERR_COMM                 ) = 'MPI_ERR_COMM                 '
    errstring(MPI_ERR_RANK                 ) = 'MPI_ERR_RANK                 '
    errstring(MPI_ERR_REQUEST              ) = 'MPI_ERR_REQUEST              '

    print*, x

  END SUBROUTINE mpi_error_handler
```

This is sort of a bug in the tree-sitter grammar, but also a problem for us because we just can't detect them without a lot more work. Every real statement function I've seen has its type declared locally, so probably we could work this out just from the AST -- lfortran has `is_statement_function` which I think just operates on the AST, so we could use this. If we remain on tree-sitter for now, then I think the logic is something like:

1. check if LHS has `call_expression` and all the arguments are identifiers and not literals or expressions
2. check if `implicit_none` is in scope (but this could be a compiler flag 🙃 )
3. check if LHS identifier is local variable
4. check that it's scalar

If all the above are true, then it's a statement function.

PR #339, opened 2025-02-28
Temporarily remove `statement-function`

ZedThree:
See #338


LiamPattinson (approved):
Looks good! I'm suddenly glad we ported over all of the deprecated/removed rule stuff from Ruff, as it makes this sort of situation really easy to resolve.

PR #340, opened 2025-02-28
Add command to generate shell completion scripts

ZedThree:
Nice little quality of life thing, very easy to use

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,6 +194,36 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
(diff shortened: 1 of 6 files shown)


LiamPattinson:
Nice, surprised at how little extra code this requires!

> very easy to use

For sensible people perhaps, but it took me a good 10 minutes to figure out where to squeeze in the generated script between the layers of bash/zsh/oh-my-zsh/antigen/powerlevel10k that make up my terminal. I should probably set fire to the whole thing and start over at some point.

Issue #76, opened 2024-10-08
Avoid `character*N`


LiamPattinson:
PR https://github.com/PlasmaFAIR/fortitude/pull/75 prevents the use of `character*(*)`, but `character*N` is still allowed. This also isn't picked up by the rule that prevents `real*N`, as in that case `N` denotes a kind instead of a length.

Should the rule that prevents `character*(*)` be modified to also catch `character*N`? I think the real problem is the use of `character*` rather than the use of an assumed size, so I'd argue that the rule should be moved into `star_kinds.rs` instead of `assumed_size.rs`.

PR #354, opened 2025-03-07
Expand rule for `character*(*)` to handle `character*N`, `character*(:)`, and `character*(expression)`

LiamPattinson:
Solves https://github.com/PlasmaFAIR/fortitude/issues/76

Something that's been in the backlog for a while.

It would be nice to retain upper/lowercase and report the variable name in the diagnostics.

I also don't have my copy of Modern Fortran Explained at hand, so I'm not sure if `modernisation` is the correct place for this!


fortitude/src/rules/modernisation/old_style_character_syntax.rs
```diff
@@ -0,0 +1,67 @@
...
+        source_file: &SourceFile,
+    ) -> Option<Vec<Diagnostic>> {
+        let src = source_file.source_text();
+
+        // Only applies to `character`
+        if node.child(0)?.kind().to_lowercase() != "character" {
+            return None;
+        }
+
+        if let Some(kind) = node.child_by_field_name("kind") {
```
ZedThree, line 47: You can use:

```suggestion
        let kind = node.child_by_field_name("kind")?;
```

to remove the `if` and reduce indentation

ZedThree:
I think this is actually just a variation on `DeprecatedAssumedSizeCharacter` and should be in there. gfortran actually only warns that this form is obsolescent, and not `character*(*)` (or the even weirder `character a(N)*N`)

fortitude/src/rules/modernisation/old_style_character_syntax.rs
LiamPattinson: Good point, there were actually two levels of indentation I could remove here. Done in https://github.com/PlasmaFAIR/fortitude/pull/354/commits/db8879e415df8c8560874c81ff09587465afafcd.

ZedThree (approved):
LGTM, thanks! One extra thing might be to add the `character*(*)` form to this test, and `character*N` to the other rule, to check we're not getting duplicate warnings?

LiamPattinson:
Good thing you asked me to do that, as it turns out there are a bunch of false positives between the two rules!

LiamPattinson:
I've also found that the test `OB061`/`deprecated-assumed-size-character` doesn't compile, as while tree-sitter seems happy to accept syntax like `character*(len=*)`, this isn't actually valid Fortran.

LiamPattinson:
Oh wow, that's absolutely awful... Just checked and the following compiles and runs just fine:

```f90
program p
  implicit none

  character*((5)) foo
  foo = "hello"
  call my_print(foo)

contains

  subroutine my_print(foo)
    character*(*) foo, bar*5
    bar = "world"
    print *, foo, bar
  end subroutine my_print

end program p
```

ZedThree:
Made a post on fortran-lang to get some outside opinion: https://fortran-lang.discourse.group/t/is-character-foo-n-n-obsolescent/9325

LiamPattinson:
I've merged the two rules. It now targets the `intrinsic_type` endpoint rather than the `assumed_size` endpoint, and only reports once if there are multiple declarations on the same line. Let me know what you think.

Due to the rule name change and the altered behaviour, this will count as a breaking change, albeit a minor one.

fortitude/src/rules/obsolescent/deprecated_character_syntax.rs
```diff
@@ -0,0 +1,85 @@
...
+        // If 'kind' field doesn't match the regex, exit early
+        let (_, length, star) = regex_captures!(
+            r#"^\*(?:[\s\(]*([[:word:]]+)[\s\)]*|\s*\(\s*(\*)\s*\))$"#,
+            kind_text
+        )?;
+        // Only one of length or star should be present. The other will be empty.
+        let length = if length.is_empty() { star } else { length };
+
+        let original = node.to_text(src)?;
+        let replacement = format!("character(len={})", length);
```
imciner2, line 70: There is the function `match_original_case` (https://github.com/PlasmaFAIR/fortitude/blob/45ac087ee67855f932b7bca247c6651530e17e3f/fortitude/src/rules/utilities.rs#L5) that will match the cases using simple heuristics (at this point, just matching the case of the first character), so it would at least preserve all uppercase and all lowercase keywords. (anything more complicated probably needs to be implemented in each rule).

LiamPattinson: Cheers, but I just figured out that I was already checking `dtype.lowercase() == "character"`, and so with `dtype.to_text(src)` I could copy it across directly. At some point we should probably add a `mixed-case` rule to prevent people writing things like `Character` or `ChArActEr`.

fortitude/src/rule_redirects.rs
```diff
@@ -57,6 +57,10 @@ static REDIRECTS: LazyLock<HashMap<&'static str, &'static str>> = LazyLock::new(
         ("S041", "MOD011"),
         ("S051", "MOD021"),
         ("S021", "C141"),
+        (
+            "deprecated-assumed-size-character",
+            "deprecated-character-syntax",
```
ZedThree, line 62: Nice! I'm glad we can do this!

fortitude/src/rules/obsolescent/deprecated_character_syntax.rs
```diff
@@ -0,0 +1,115 @@
...
+                } else {
+                    panic!("This should not happen");
+                }
+            }
+        };
+
+        let original = node.to_text(src)?.to_string();
+        let dtype = dtype.to_text(src)?.to_string();
+        let replacement = format!("{}(len={})", dtype, length);
+        let fix = Fix::safe_edit(node.edit_replacement(source_file, replacement));
```
ZedThree, line 100: If this is a safe edit, should it be an `AlwaysFixableViolation`?

fortitude/src/rules/obsolescent/deprecated_character_syntax.rs
```diff
@@ -0,0 +1,115 @@
...
+            _ => {
+                // If there is no math_expression node, this may be an integer
+                // literal or '(*)', or it may also be something like '(N)', or
+                // '((((4))))'.
+                // An arbitrary amount of whitespace is permitted.
+                // Strangely, plain '*N' is not allowed.
+                let (_, length, star, expression) = regex_captures!(
+                    r#"^\*(?:\s*(\d+)|\s*\(\s*(\*)\s*\)|\([\s\(]*([[:word:]]+)[\s\)]*\))$"#,
+                    kind_text
+                )?;
```
ZedThree, line 83: Sort of surprised we have to go for a regex here. Is the tree very complicated at this point?

ZedThree (approved):

fortitude/src/rule_redirects.rs
LiamPattinson: Oddly, it doesn't raise a redirect warning. I didn't look into it deeply enough to figure out why. Should I try to get that fixed before the merge?

ZedThree: Hmm, a little suspicious. Does it actually work?

ZedThree: I mean, the fix might be best in another PR, but it would be good to check if it's a bug

fortitude/src/rules/obsolescent/deprecated_character_syntax.rs
LiamPattinson: Looking into it, I don't know if we strictly need to. The tree does look quite strange, with `character*(expression)` having an `assumed_size` node even if expression is not `*`. I'll see if I can simplify things.

LiamPattinson: Neat, managed to get it down to just a few lines in https://github.com/PlasmaFAIR/fortitude/pull/354/commits/e101e5bd1e012de0c0426ba3b69167af0b778956.

This isn't at all urgent or necessary, but could the `tree-sitter` grammar be amended so that the first `assumed_size` node is instead something like `character_length_operator`?

ZedThree:
How breaking is this change, actually? With the name redirect, users should just get a warning, right?

LiamPattinson:
The name change doesn't generate a warning in this case because the short code hasn't changed. I could fix that, but we would need to tweak `RuleSelector::from_str`.

I guess it could be considered a breaking change because an existing 'stable' rule is now behaving differently. I'd personally be tempted to just call this a bugfix and release it in v0.7.1 after we've cleaned up a few other outstanding issues.

imciner2:
Since the name changed, I guess that would break any allow comments in existing code? Would there be a way to alias them somehow?

LiamPattinson:
> Since the name changed, I guess that would break any allow comments in existing code? Would there be a way to alias them somehow?

We have the rule `FORT003` to warn about redirected allow comments, and they should just work. For example:

```f90
! allow(T004)
implicit none
```

Results in:

```bash
test.f90:2:11: FORT003 [*] `T004` is a redirect to `C003`
  |
1 | program p
2 |   ! allow(T004)
  |           ^^^^ FORT003
3 |   implicit none
  |
  = help: Replace with `C003` or `implicit-external-procedures`
```

 In this case however, it's causing Fortitude to panic somewhere. I guess I'll need to fix this up sooner rather than later.

ZedThree:
Allow comments do work with the old name. Testing on the `OB061.F90` test file:

```f90
  ! allow(deprecated-assumed-size-character)
  subroutine char_input(a, b, c, d, e, f, g, h, i, j, k, l, m, n)
```

does indeed allow the violation. However for some reason this only works if I explicitly pass `--select=OB`, even though the rule is allowed?

```
thread 'main' panicked at fortitude/src/allow_comments.rs:138:62:
called `Result::unwrap()` on an `Err` value: Unknown
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/panicking.rs:76:14
   2: core::result::unwrap_failed
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/library/core/src/result.rs:1699:5
   3: core::result::Result<T,E>::unwrap
             at /home/peter/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1104:23
   4: fortitude::allow_comments::check_allow_comments
             at ./fortitude/src/allow_comments.rs:138:36
   5: fortitude::check::check_path
             at ./fortitude/src/check.rs:284:23
   6: fortitude::check::check_only_file
             at ./fortitude/src/check.rs:179:22
   7: fortitude::check::check_file
             at ./fortitude/src/check.rs:121:22
   8: fortitude::check::check_files::{{closure}}
             at ./fortitude/src/check.rs:870:19
   9: <rayon::iter::map::MapFolder<C,F> as rayon::iter::plumbing::Folder<T>>::consume
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/map.rs:237:27
  10: <indicatif::rayon::ProgressFolder<C> as rayon::iter::plumbing::Folder<T>>::consume
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/indicatif-0.17.9/src/rayon.rs:186:19
  11: rayon::iter::plumbing::Folder::consume_iter
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/plumbing/mod.rs:178:20
  12: rayon::iter::plumbing::Producer::fold_with
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/plumbing/mod.rs:109:9
  13: rayon::iter::plumbing::bridge_producer_consumer::helper
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/plumbing/mod.rs:437:13
  14: rayon::iter::plumbing::bridge_producer_consumer
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/plumbing/mod.rs:396:12
  15: <rayon::iter::plumbing::bridge::Callback<C> as rayon::iter::plumbing::ProducerCallback<I>>::callback
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/plumbing/mod.rs:372:13
  16: <rayon::slice::Iter<T> as rayon::iter::IndexedParallelIterator>::with_producer
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/slice/mod.rs:826:9
  17: rayon::iter::plumbing::bridge
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/plumbing/mod.rs:356:12
  18: <rayon::slice::Iter<T> as rayon::iter::ParallelIterator>::drive_unindexed
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/slice/mod.rs:802:9
  19: indicatif::rayon::<impl rayon::iter::ParallelIterator for indicatif::iter::ProgressBarIter<T>>::drive_unindexed
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/indicatif-0.17.9/src/rayon.rs:205:9
  20: <rayon::iter::map::Map<I,F> as rayon::iter::ParallelIterator>::drive_unindexed
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/map.rs:49:9
  21: <rayon::iter::fold::Fold<I,ID,F> as rayon::iter::ParallelIterator>::drive_unindexed
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/fold.rs:59:9
  22: rayon::iter::reduce::reduce
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/reduce.rs:15:5
  23: rayon::iter::ParallelIterator::reduce
             at /home/peter/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.10.0/src/iter/mod.rs:998:9
  24: fortitude::check::check_files
             at ./fortitude/src/check.rs:839:23
  25: fortitude::check::check
             at ./fortitude/src/check.rs:735:9
  26: fortitude::main
             at ./fortitude/src/main.rs:18:43
  27: core::ops::function::FnOnce::call_once
             at /home/peter/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
```

Ah, the problem is `redirected-allow-comment (FORT003)`. Turning that off avoids the panic. So there's a bug in there -- probably we're only expecting codes and not names!

LiamPattinson:
@ZedThree @imciner2 I think I have it fixed now. The bug went deeper than just `FORT003` itself, so I've overhauled how rule long names are handled. Previously, `RuleSelector::from_str` would test if a rule could be parsed from the long name as a string (`line-too-long` -> `LinetooLong`), convert to the short code (`S001`), and recursively call `from_str` again from the short code. Now, `Category::parse_code` can handle the long name directly. This means redirect warnings are now raised as expected:

I've also rewritten a bit of `FORT003`, as it was repeating the long name of `deprecated-character-syntax` twice in the help text, and I've updated the redirect warnings to ensure long-name redirects are reported correctly.

Now that I've figured out a way to handle long names with the macros, it might be possible to push this functionality back into Ruff.

fortitude/src/rules/fortitude/allow_comments.rs
```diff
@@ -58,10 +58,8 @@ pub(crate) struct RedirectedAllowComment {
 impl AlwaysFixableViolation for RedirectedAllowComment {
     #[derive_message_formats]
     fn message(&self) -> String {
-        let Self {
-            original, new_code, ..
-        } = self;
-        format!("`{original}` is a redirect to `{new_code}`")
+        let Self { original, .. } = self;
+        format!("`{original}` has been redirected")
```
ZedThree, line 62: Have we dropped this because we don't know (at this point, at least) if it was a code or a name that's triggered the redirect?

The fix will include the correct one, so this isn't so important, but how difficult would it be to get the redirect here too?

LiamPattinson: Whoops, I changed that on my first pass at fixing this up and forgot to come back to it. It'll be easy enough to fix.

LiamPattinson: Fixed in https://github.com/PlasmaFAIR/fortitude/pull/354/commits/e5ece1e2745b174b902c4452ff69c26a3a86a171

ZedThree (approved):

ZedThree:
Although you've also fixed a bug here, I don't think it was possible for it to manifest previously, so we probably don't need a separate changelog entry for that?

LiamPattinson:
I don't think it could, as previously all redirects were short code -> short code, so I don't think it needs flagging up as a bug-fix.

Issue #355, opened 2025-03-07
Linting using stdin is inconsistent


hougesen:
I noticed that when linting stdin input an output is only printed when the file has changed (unless I am doing something wrong). 

I would expect that an output, either the original input or the fixed, to always be printed when fixing is enabled. 

Not always printing makes it impossible to run the tool directly from neovim since the file is replaced with nothing  when it finds no errors (`%! fortitude check --fix --quiet --stdin-filename hello.f90`).

LiamPattinson:
Hi @hougesen, thanks for raising this! I hadn't considered this use-case myself, but I can replicate the behaviour. @ZedThree I'm having some trouble locating where in the code we decide to write out the fixed file to stdout.

ZedThree:
This bit: https://github.com/PlasmaFAIR/fortitude/blob/45ac087ee67855f932b7bca247c6651530e17e3f/fortitude/src/check.rs#L957-L958

Ruff also has this wrapper that we don't have: https://github.com/astral-sh/ruff/blob/main/crates/ruff/src/commands/check_stdin.rs

LiamPattinson:
Digging deeper into how Ruff does things, it seems they always output the input file if no fixes were found or fixing failed:

https://github.com/astral-sh/ruff/blob/36d12cea4769cf36c83c9a1bc92399ea6afa5952/crates/ruff/src/diagnostics.rs#L419-L423

https://github.com/astral-sh/ruff/blob/36d12cea4769cf36c83c9a1bc92399ea6afa5952/crates/ruff/src/diagnostics.rs#L443-L457

I'll see if I can get this working. I'll add something to the docs too, as users might find it useful to bind this to a custom vim command.

PR #358, opened 2025-03-11
Write to stdout even when no fixes are applied in stdin mode

LiamPattinson:
Solves https://github.com/PlasmaFAIR/fortitude/issues/355

Also updates the docs (and fixes some minor docs issues I've been aware of in the process). @ZedThree Could you let me know if there's an equivalent Emacs command? I'll try to figure out how to add this to VSCode's command palette.

While working on this I also uncovered and fixed some bugs in the behaviour of `--fix-only` mode. I'll raise a separate PR to keep things nicer in the changelog.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,12 @@ Written in Rust :crab: and installable with Python :snake:.
 - :zap: Blazingly fast, up to hundreds of times faster than other open-source Fortran
   linters.
 - :wrench: Automatically fixes linter warnings.
-- :chart_with_upwards_trend: 30+ rules, with many more planned.
+- :chart_with_upwards_trend: 50+ rules, with many more planned.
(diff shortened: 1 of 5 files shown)


ZedThree (approved):
For Emacs, there's a couple of options. Built-in, you can select the whole buffer and run `C-u M-|` (`shell-command-on-region` with prefix argument to replace the contents). This is a little awkward, but does work out of the box.

There's also a nice [reformatter](https://github.com/purcell/emacs-reformatter) package that makes this easier:

```elisp
(reformatter-define fortitude-lint
  :program "/path/to/fortitude"
  :args '("check" "--fix-only" "--silent" "-"))
```

which defines `fortitude-lint-buffer` and `fortitude-lint-on-save-mode` commands. The latter can be used in your config to automatically run in Fortran buffers when they're saved.

PR #359, opened 2025-03-11
Don't print diagnostics in `--fix-only` mode

LiamPattinson:
Builds on https://github.com/PlasmaFAIR/fortitude/pull/358

I discovered that `--fix-only` mode wasn't working at all! This prevents printing the full summary text or any diagnostics when running in this mode.


LiamPattinson:
Between the previous PR and this one, I switched the tests from running in `--silent` mode to `--quiet` mode, so the stdin tests are now testing this indirectly. Let me know if you'd like something more explicit.

ZedThree (approved):

PR #361, opened 2025-03-12
Use `pre-commit` in place of custom `.githooks` scripts

LiamPattinson:
Something I've been meaning to update for a while, as my custom pre-commit script won't work on Windows. I'm also just generally quite impressed with `pre-commit`.


ZedThree (approved):

stigh:
@LiamPattinson, sorry for commenting in a closed PR. I also use pre-commit in my work. It would be nice if Fortitude could be used as a pre-commit hook. I assume there's some tricks to make this happen, for example running a local hook... But it would be really nice if it was possible to add something like this to my `.pre-commit-config.yaml`:

```
  - repo: https://github.com/PlasmaFAIR/fortitude
    rev: v0.7.0
    hooks:
      - id: fortitude
```

Is it appropriate to create a separate issue on this?

LiamPattinson:
> Is it appropriate to create a separate issue on this?

@stigh Of course, go for it! It's something I'd thought about but wasn't sure how to go about it. Looking at [ruff-pre-commit](https://github.com/astral-sh/ruff-pre-commit), it actually doesn't seem too hard to achieve.

PR #364, opened 2025-03-13
Add rule for `exit`/`cycle` in unlabelled loop

### LINKED ISSUE #56: Nested loops with `exit` / `cycle` should be labelled
Bad:

```fortran
do i = 1, n
  do j = 1, n
     if (i == j) cycle
  end do
end do
```

Good:

```fortran
outer: do i = 1, n
  inner: do j = 1, n
     if (i == j) cycle inner
  end do inner
end do outer
```

### LINKED ISSUE #280: Only `exit` or `cycle` a named `do` loop
At https://fortran-lang.discourse.group/t/ending-a-do-loop/2329/6

John Collins, whose [company](http://www.simconglobal.com/) sells Fortran static analysis tools, wrote, 

"I strongly advise never to use an un-tagged EXIT statement. We write tools to detect software errors. The number of errors associated with un-tagged EXIT and LEAVE statements is alarming. Tag the loop and tag the EXIT and the danger is removed."

Fortitude could warn about code such as

```fortran
implicit none
integer :: i,j,k
do i=1,3
   do j=1,2
      k = i**j
      if (k > 5) exit
      print*,k
   end do
end do
end
```

which could be written as

```fortran
implicit none
integer :: i,j,k
do i=1,3
   inner: do j=1,2
      k = i**j
      if (k > 5) exit inner
      print*,k
   end do inner
end do
end
```

to avoid the warning.

ZedThree:
Closes #280
Closes #56

### DIFF
--- a/docs/rules/exit-or-cycle-in-unlabelled-loop.md
+++ b/docs/rules/exit-or-cycle-in-unlabelled-loop.md
@@ -5,9 +5,9 @@ This rule is unstable and in [preview](../preview.md). The `--preview` flag is r
 Checks for `exit` or `cycle` in unnamed `do` loops
 
 ## Why is this bad?
-Using loop labels with `exit` and `cycle` statements prevents bugs from
-exiting the wrong loop. The danger is particularly enhanced when code is
-refactored to add further loops.
+Using loop labels with `exit` and `cycle` statements prevents bugs when exiting the
+wrong loop, and helps readability in deeply nested or long loops. The danger is
+particularly enhanced when code is refactored to add further loops.
 
 ## Settings
-See [nested-loops-only](../settings.md#check_exit-labelled-loops_nested-loops-only)
+See [nested-loops-only](../settings.md#check_exit-unlabelled-loops_allow-unnested-loops)
--- a/docs/rules/exit-or-cycle-in-unlabelled-loop.md
+++ b/docs/rules/exit-or-cycle-in-unlabelled-loop.md
@@ -10,4 +10,4 @@ wrong loop, and helps readability in deeply nested or long loops. The danger is
 particularly enhanced when code is refactored to add further loops.
 
 ## Settings
-See [nested-loops-only](../settings.md#check_exit-unlabelled-loops_allow-unnested-loops)
+See [allow-unnested-loops](../settings.md#check_exit-unlabelled-loops_allow-unnested-loops)
--- a/docs/rules/exit-or-cycle-in-unlabelled-loop.md
+++ b/docs/rules/exit-or-cycle-in-unlabelled-loop.md
@@ -8,3 +8,6 @@ Checks for `exit` or `cycle` in unnamed `do` loops
 Using loop labels with `exit` and `cycle` statements prevents bugs from
 exiting the wrong loop. The danger is particularly enhanced when code is
 refactored to add further loops.
+
+## Settings
+See [nested-loops-only](../settings.md#check_exit-labelled-loops_nested-loops-only)
--- a/docs/rules/exit-or-cycle-in-unlabelled-loop.md
+++ b/docs/rules/exit-or-cycle-in-unlabelled-loop.md
@@ -1,0 +1,10 @@
+# exit-or-cycle-in-unlabelled-loop (C142)
+This rule is unstable and in [preview](../preview.md). The `--preview` flag is required for use.
+
+## What does it do?
+Checks for `exit` or `cycle` in unnamed `do` loops
+
+## Why is this bad?
+Using loop labels with `exit` and `cycle` statements prevents bugs from
+exiting the wrong loop. The danger is particularly enhanced when code is
+refactored to add further loops.
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -571,15 +571,22 @@ If set to false, the hint will be hidden.
 
 ---
 
-### `check.exit-labelled-loops`
+### `check.exit-unlabelled-loops`
 
 Options for the `exit-or-cycle-in-unlabelled-loops` rule
 
-#### [`nested-loops-only`](#check_exit-labelled-loops_nested-loops-only) {: #check_exit-labelled-loops_nested-loops-only }
-<span id="nested-loops-only"></span>
+#### [`allow-unnested-loops`](#check_exit-unlabelled-loops_allow-unnested-loops) {: #check_exit-unlabelled-loops_allow-unnested-loops }
+<span id="allow-unnested-loops"></span>
 
 Whether to check for `exit`/`cycle` in unlabelled loops only if the loop has at
-least one level of nesting
+least one level of nesting. With this setting off (default), the following will
+raise a warning, and with it on, it won't:
+
+```f90
+do i = 1, 100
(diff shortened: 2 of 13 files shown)


ZedThree:
There's also #56 which is a more specific variation on this. We could have a setting `nested_only` (for example) that only activate the rule on nested loops.

This rule and C141 could probably be done with a single function as well, once we move to that structure

LiamPattinson (approved):
Cheers, all looks good! Definitely a strong case for the AST scanning refactor.

ZedThree:
I've add a setting to control whether this rule is triggered by any loops or nested ones only. As it's a setting just for an individual rule, I've tried out the `option_group` functionality from ruff, which adds it to a `exit-unlabelled-loops` group. This little bit of nesting is nice to make it clear it's not a general setting... but it does require a bunch of machinery which might get a little onerous if we only have one or two settings for a rule. It's less bad in ruff because they group settings by plugin, and the boilerplate overhead is much less. It also means they can automate linking from the list of rules to the relevant settings.

I think I'm worrying over nothing. Looking through ruff's settings, very few of their plugins actually have more than two settings. We're likely to end up in a similar situation, with most rules not needing any settings, and a few related rules we can put in a group for the purposes of settings.

fortitude/src/rules/correctness/exit_labels.rs
```diff
@@ -107,21 +110,29 @@ impl Violation for ExitOrCycleInUnlabelledLoop {
...
+
         node.ancestors()
             .filter(|ancestor| ancestor.kind() == "do_loop_statement")
             .filter(|do_loop| {
                 do_loop
                     .child_with_name("block_label_start_expression")
                     .is_none()
             })
-            .nth(0)?;
+            .nth(nth)?;
```
LiamPattinson, line 135: If I'm understand right, when `nth == 1`, this will report no violation if only one of the inner and output loop are labelled. Just to double-check, is this the intent?

docs/settings.md
```diff
@@ -571,3 +571,34 @@ If set to false, the hint will be hidden.
...
+
+    ```toml
+    [fpm.extra.check.exit-labelled-loops]
+    nested-loops-only = true
+    ```
+=== "fortitude.toml"
+
+    ```toml
+    [check.exit-labelled-loops]
+    nested-loops-only = true
```
LiamPattinson, line 600: I'm not sure about the name of the option group or the option itself. I feel like the option group should name the thing the rule is trying to prevent, and the option itself should be clearer about what is being switched on or off. Maybe something like:

```toml
[check.exit-unlabelled-loops]
allow-non-nested-loops = true
```

Not 100% sure about this though.

fortitude/src/options.rs
```diff
@@ -279,4 +280,32 @@ pub struct CheckOptions {
...
+/// Options for the `exit-or-cycle-in-unlabelled-loops` rule
+#[derive(
+    Clone, Debug, PartialEq, Eq, Default, OptionsMetadata, CombineOptions, Serialize, Deserialize,
+)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct ExitLabelledLoopOptions {
+    /// Whether to check for `exit`/`cycle` in unlabelled loops only if the loop has at
+    /// least one level of nesting
+    #[option(
+        default = "false",
```
LiamPattinson, line 298: Generally speaking, should we default to the stricter interpretation of the rule and require the user to tell us to be nicer, or would the opposite be preferable? Personally, I think I would always want this to be `true`.

LiamPattinson (commented):
Nice one! This also lays the groundwork for my plans to add a `[check.real_kinds]` group down the line. Just a few questions in the comments.

fortitude/src/rules/correctness/exit_labels.rs
ZedThree: Ah, good catch. This needs to be a bit more complicated. I'll have to think a bit more what the "nested-only" version should trigger on. I'm less certain of the utility of this setting now!

fortitude/src/options.rs
ZedThree: Currently the default is for the stricter version, triggering on `exit`/`cycle` in just one level of unlabelled loop. Turning this on makes it need an extra level of nested loop.

docs/settings.md
ZedThree: Sure, I think this is clear. I'll also make the wording the docs clearer too

fortitude/src/rules/correctness/exit_labels.rs
LiamPattinson: I think it could still work:

- If the exit/cycle is labelled, exit early. I assume the compiler would pick up if the label doesn't refer to anything (plus this could be a different rule, `unmatched-exit-label` or something).
- If it's unlabelled, count the number of loops in the ancestors.
- When running in `nested-only` mode, count the number of loops in the ancestors (labelled or unlabelled), and if the count is only 1, exit early.
- Otherwise, iterate over the ancestors again until finding the first unlabelled loop.
- If you find anything, raise the violation.

Perhaps the `fix_title` help text could point to the nesting level which is unlabelled? This is a case where I think it would be nice if we could highlight two blocks of code in the diagnostics -- first the offending `exit`/`cycle`, and second the offending `do` loop(s).

ZedThree: Think I've got this sorted now and covering the edge cases I could think of. Basically we only really care if the inner most loop is labelled or not. Then we might care if there's any other parent loops, and we don't care if they are labelled.

> Perhaps the fix_title help text could point to the nesting level which is unlabelled? This is a case where I think it would be nice if we could highlight two blocks of code in the diagnostics -- first the offending exit/cycle, and second the offending do loop(s).

Yes, agreed!

ZedThree:
Bah, I see why ruff doesn't tell you all the possible keys in the toml file now.

docs/rules/exit-or-cycle-in-unlabelled-loop.md
```diff
@@ -0,0 +1,13 @@
...
+## What does it do?
+Checks for `exit` or `cycle` in unnamed `do` loops
+
+## Why is this bad?
+Using loop labels with `exit` and `cycle` statements prevents bugs when exiting the
+wrong loop, and helps readability in deeply nested or long loops. The danger is
+particularly enhanced when code is refactored to add further loops.
+
+## Settings
+See [nested-loops-only](../settings.md#check_exit-unlabelled-loops_allow-unnested-loops)
```
LiamPattinson, line 13:
```suggestion
See [allow-unnested-loops](../settings.md#check_exit-unlabelled-loops_allow-unnested-loops)
```

LiamPattinson (changes requested):
Cheers, all looks good except for a link in the docs

docs/rules/exit-or-cycle-in-unlabelled-loop.md
ZedThree: Oops! :facepalm: Checked the link actually worked, didn't check this!

LiamPattinson (approved):

PR #368, opened 2025-03-14
Fix `--statistics` reporting for unsafe fixes

### LINKED ISSUE #367: Unsafe fix is flagged as fixable in `--statistics`
## Description

Unsafe fixes are flagged and explicitly mentioned as fixable when running `fortitude check --statistics`. Adding `--fix` returns nothing.

```sh
$ fortitude check --statistics --select C003 --fix
2	C003	[*] implicit-external-procedures
[*] fixable with `fortitude check --fix`
  # empty return - implied success but no changes applied
```

Removing `--statistics` does show the correct issue status:

```sh
$ fortitude check --select C003
src/particle_types.f90:3:3: C003 'implicit none' missing 'external'
test/check.f90:2:4: C003 'implicit none' missing 'external'
No fixes available (2 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

ZedThree:
Fixes #367

Previously, unsafe fixes were counted as "fixable" in
`Printer::write_statistics`, in contrast to the behaviour in
`Printer::write_once`. This changes the behaviour to align with
`write_once`, including them only if `--unsafe-fixes` is set.

We now also reuse `Printer::write_summary` to avoid duplicating the
logic for whether or not to report if there are hidden fixes.

### DIFF
--- a/fortitude/src/check.rs
+++ b/fortitude/src/check.rs
@@ -770,7 +770,7 @@ pub fn check(args: CheckArgs, global_options: &GlobalConfigArgs) -> Result<ExitC
     );
 
     if args.statistics {
-        printer.write_statistics(&results.diagnostics, &mut summary_writer)?;
+        printer.write_statistics(&results, &mut summary_writer)?;
     } else {
         printer.write_once(&results, &mut summary_writer)?;
     }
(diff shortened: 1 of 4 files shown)


LiamPattinson (approved):
Nice, all looks good! I still think we should maybe trim down the summary text at some point though. Maybe something for v0.8?

PR #369, opened 2025-03-14
Add script to add boilerplate for new (AST) rule

ZedThree:
Obviously stolen from ruff.

Should make it easier to new folks to get started.


LiamPattinson (approved):
Cheers, this will be really helpful if we ever get to run a workshop or hackathon to get a load of new rules in.

LiamPattinson (approved):
Thanks for that, happy to merge

Issue #375, opened 2025-03-17
Provide pre-commit hooks


LiamPattinson:
@LiamPattinson, sorry for commenting in a closed PR. I also use pre-commit in my work. It would be nice if Fortitude could be used as a pre-commit hook. I assume there's some tricks to make this happen, for example running a local hook... But it would be really nice if it was possible to add something like this to my `.pre-commit-config.yaml`:

```
  - repo: https://github.com/PlasmaFAIR/fortitude
    rev: v0.7.0
    hooks:
      - id: fortitude
```

Is it appropriate to create a separate issue on this?

_Originally posted by @stigh in https://github.com/PlasmaFAIR/fortitude/issues/361#issuecomment-2729298531_

stigh:
My mistake! You do use git tagging 😉

LiamPattinson:
Pre-commit hooks are now provided via https://github.com/PlasmaFAIR/fortitude-pre-commit. Please let me know over on that repo if you run into any trouble with them.

PR #376, opened 2025-03-17
Add pre-commit hooks

LiamPattinson:
Solves https://github.com/PlasmaFAIR/fortitude/issues/375

Adds pre-commit hooks so that the following with work:

```
# .pre-commit-config.yaml
repos:
  - repo: https://github.com/PlasmaFAIR/fortitude
    rev: feature/pre-commit-hooks
    hooks:
    - id: fortitude
```

After running `pre-commit install`, this will check all staged files. You can also add `args: ["--preview", "--fix", "--unsafe-fixes"]` to apply fixes if you're feeling really confident.

_However_, I don't actually think this should be merged. The problem is that `pre-commit` will clone the repo and try to install the project from scratch with `pip install`. This takes about a minute on my machine, and it will only work if the user has a working Rust environment (tested using a clean Docker container).

One option is to set `language: rust` in `.pre-commit-hooks.yaml`, as this bootstraps rust for the user, but I couldn't get this to work as it doesn't provide options for installing a particular binary from a workspace package.

Ruff get around this by having a secondary project [`ruff-pre-commit`](https://github.com/astral-sh/ruff-pre-commit) (once again, I couldn't understand why they'd done it this way until I tried doing it a more 'clever' way myself, and now it makes perfect sense). I suggest we copy this instead.


ZedThree (approved):
Do we need to keep the versions in the docs up-to-date with the latest release? What does ruff do?

LiamPattinson:
I've just been trying to figure that out. I can't tell where they're doing it in `ruff_dev`, if they're doing it there at all.

LiamPattinson:
I've now created the repo [fortitude-pre-commit](https://github.com/PlasmaFAIR/fortitude-pre-commit), based heavily on the similar ruff-pre-commit. I've tested its auto-version-update features and it looks good to me.

LiamPattinson:
I think `scripts/release.sh` handles the version update, or at least that's the case according to Ruff's `CONTRIBUTING.md`

PR #379, opened 2025-03-19
Bump version to 0.7.1

### LINKED ISSUE #372: Version 0.7.1
We've fixed a few bugs and added a few rules/features. Would now be a good time for a patch release? Please leave a comment if you think there are any quick and easy tasks we should do beforehand.

ZedThree:
Closes #372

We probably want to do something about #375 before merging?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,35 @@
 # Changelog
 
+## 0.7.1
+
+This is a bug-fix release, with one new rule. Thanks to our new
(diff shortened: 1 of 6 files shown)


LiamPattinson:
> We probably want to do something about https://github.com/PlasmaFAIR/fortitude/issues/375 before merging?

Yes, but we'll only need to update the docs to point people towards the new repo.

LiamPattinson (approved):

PR #380, opened 2025-03-19
docs: Remove readthedocs version menu from nav bar

ZedThree:
Previously this was getting duplicated on every page navigation due to
some incompatibility with mkdoc-material's "instant loading"
feature. It's probably better to have a speedy website than a fancy one!


LiamPattinson (approved):
That's really strange, was it always doing that?

Happy to merge, but it would nice if it could be brought back at some point.

LiamPattinson:
I hadn't spotted the other version selector -- in that case, I'd say it's not a big problem to lose the one at the top.

PR #382, opened 2025-03-24
Fix testcase extension in generated rule template

imciner2:
The testcase file is created as a `.f90` extension, but the script called it a `.py` file when adding the testcase to the module.


ZedThree (approved):
Thanks @imciner2 !

PR #386, opened 2025-03-25
Add rule to detect missing function result clause

imciner2:
This is probably a bit of an opinionated style rule (so I put it in Style), but it seems like something that could be useful.

This adds a rule to detect if the `result()` clause is not used in a function declaration, e.g., it will flag the following:

```fortran
function distance_real64(a, b)
    real(kind=real64), intent(in) :: a, b
    real(kind=real64)             :: distance_real64
    distance_real64 = abs(a-b)
end function distance_real
```


fortitude/src/rules/style/functions.rs
```diff
@@ -0,0 +1,103 @@
...
+        _settings: &Settings,
+        node: &'a Node,
+        _src: &'a SourceFile,
+    ) -> Option<Vec<Diagnostic>> {
+        let mut node_cursor = node.walk();
+
+        // Just need to check for the presence of the function_result node
+        if node
+            .named_children(&mut node_cursor)
+            .any(|node| node.kind() == "function_result")
```
LiamPattinson, line 92: If you include `crate::ast::FortitudeNode`, you can remove the node cursor and replace this with:
```suggestion
        if node.child_with_name("function_result").is_some()
```

fortitude/src/rules/style/functions.rs
```diff
@@ -0,0 +1,103 @@
...
+/// ## References
+/// - Clerman, N. Spector, W., 2012, _Modern Fortran: Style and Usage_, Cambridge
+///   University Press, Rule 129, p. 178.
+#[derive(ViolationMetadata)]
+pub(crate) struct FunctionMissingResult;
+
+impl Violation for FunctionMissingResult {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Function missing result() specifier".to_string()
```
LiamPattinson, line 77: Would it be possible to include the function name in the help text? `rules/style/end_statements.rs` has a similar example to this.

fortitude/src/rules/mod.rs
```diff
@@ -125,6 +125,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Style, "201") => (RuleGroup::Stable, Ast, Optional, style::implicit_none::SuperfluousImplicitNone),
         (Style, "211") => (RuleGroup::Preview, Ast, Optional, style::file_contents::MultipleModules),
         (Style, "212") => (RuleGroup::Preview, Ast, Optional, style::file_contents::ProgramWithModule),
+        (Style, "220") => (RuleGroup::Preview, Ast, Optional, style::functions::FunctionMissingResult),
```
LiamPattinson, line 128: A super pedantic thing: Could this be rule 221 instead? The reason we never end with a zero is because we don't want to have rule codes like `S000`, and for consistency with each of the other groupings such as `S01x`, `S21x`, etc I've been avoiding ending any rule codes with a 0. The only exception is `E000`, which is only raised if a file can't be accessed for some reason.

I thought this was something I'd picked up from Ruff, but on double checking it seems they're all over the place with rule codes, so I'm now not sure where I got this from.

ZedThree (approved):
LGTM, thanks @imciner2!

Maybe the code should be `221` for consistency? I think we might have let our numbering evolve organically, but we seem to start most sub-groups with `xx1`

LiamPattinson (commented):
Thanks @imciner2, I think this is a really good idea! It's something I'll likely use for my own work.

I have a few comments, but there's nothing major to address. I tried looking for edge cases that might result in false positives/negatives but it seems pretty solid to me. My biggest concern was that it might not handle nested functions properly, but as far as I can tell it has no trouble with them whatsoever.

fortitude/src/rules/mod.rs
imciner2: Yea, I just forgot the standard was starting at 1 - changed now.

fortitude/src/rules/style/functions.rs
LiamPattinson: The benefit of putting the function name in the title is that it gives a little more info when users are using `--output-format=concise` (or similar), but, as you've pointed out with `unnamed-end-statement`, we haven't always been very consistent about this (sorry, I didn't spot that it was just in the fix title!).

I'd be happy to merge with things as they are now. We're planning to do a big overhaul of how AST rules are handled soon in order to make it easier to keep track of scoping rules and other useful information, so a lot of rules will likely get minor updates like this as part of that process.

LiamPattinson (approved):

Issue #383, opened 2025-03-24
`fpm.toml` discrepancy in the docs


LiamPattinson:
On the 'Configuration' page, the docs say to list settings under the heading `[extra.fortitude.check]`, while on the 'Settings' page, the examples say to list settings under `[fpm.extra.check]`. Only the former works.

PR #387, opened 2025-03-25
Fix `fpm.toml` examples in `docs/settings.md`

LiamPattinson:
Solves https://github.com/PlasmaFAIR/fortitude/issues/383

### DIFF
--- a/fortitude_dev/src/generate_options.rs
+++ b/fortitude_dev/src/generate_options.rs
@@ -205,7 +205,7 @@ fn format_tab(tab_name: &str, header: &str, content: &str) -> String {
 
 /// Format the TOML header for the example usage for a given option.
 ///
-/// For example: `[extra.foritutde.check]` in `fpm.toml`` or `[check]` in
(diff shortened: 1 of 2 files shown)


fortitude_dev/src/generate_options.rs
```diff
@@ -205,10 +205,11 @@ fn format_tab(tab_name: &str, header: &str, content: &str) -> String {
 
 /// Format the TOML header for the example usage for a given option.
 ///
-/// For example: `[fpm.extra.format]` or `[fpm.extra.lint.isort]`.
+/// For example: `[extra.foritutde.check]` in `fpm.toml`` or `[check]` in
```
ZedThree, line 208:
```suggestion
/// For example: `[extra.fortitude.check]` in `fpm.toml`` or `[check]` in
```

ZedThree (approved):

fortitude_dev/src/generate_options.rs
LiamPattinson: Whoops, cheers for that

Issue #66, opened 2024-10-04
Prefer `else if` over `elseif` and `end if` over `endif`


LiamPattinson:
Or should it be the other way around? Either way, we should encourage consistency.

blaylockbk:
I'm really excited to have just discovered this project.

I believe fprettify by default auto formats to having a space. It would be nice if the auto formatter and linter are consistent.

imciner2:
This should probably be generalized to include all the keywords where the separating blanks are optional, i.e.,
![image](https://github.com/user-attachments/assets/23926f51-60e7-4a77-906b-7edccd3ea7e4)

The only thing about this is, I can see there wanting to be different rules for various keywords on that list (e.g., you can technically put a space inside `intent(in out)` according to that table, but I personally might keep them together while wanting the spaces in other keywords)... but I think one rule per keyword is probably too excessive - so I am not sure what the best tradeoff is here.

ZedThree:
Easiest way I can think of is to be opinionated and just default to no whitespace for `intent(inout)` but with an option to include it, something like: `missing-keyword-whitespace.whitespace-intent-inout = true`. No options for anything else

PR #393, opened 2025-03-27
Add rules preferring multi-word keywords to include spaces

LiamPattinson:
Closes https://github.com/PlasmaFAIR/fortitude/issues/66

I've started this off with a rule that only targets `elseif`/`elsewhere`, but I thought I would seek feedback before proceeding.

- Should there be one rule for all cases? This would include `end x`, `select x`, `double precision`, `go to`, and `in out`. Or one rule for each of those?
- I'm not sure how to justify this one for the docs.


imciner2:
Given the number of keywords that can have optional spaces (shown in the table in https://github.com/PlasmaFAIR/fortitude/issues/66#issuecomment-2576111816) I don't know if a rule-per-keyword is going to be good. I think having a general rule that searches for any of those optional keywords missing spaces, and then being opinionated about only including a few settings to allow ignoring certain keywords to let them have missing spaces in certain keywords (such as `inout` and `goto`) would be better.

The main justification I can think of is that the spacing can improve readability of the code and make it obvious what the second portion of a compound statement is when there are multiple options for that second portion.

LiamPattinson:
I was originally thinking about having rules like `else-clause-missing-space`, `end-clause-missing-space`, etc that would give some granularity without requiring different rules for e.g. `end module` and `end program`. I'm now instead coming around to the idea of just having one rule for all of them, although it likely won't be very pretty code. I'll give it a go, I don't think I'll bother for obsolescent things like `block data` at this stage.

ZedThree:
Yep, I think just one rule, `missing-whitespace-in-keyword` or something is best. I thought I'd quickly check if there's a clear winner between `intent(inout)` and `intent(in out)` across the ~34M lines in my corpus:

```
ag -io "intent *\( *in *out *\)" -G".*\.f.*" --nonumbers --nofilename | sort | uniq -c | sort -hr
```

| count  | |
|--------|-|
|  90416 | `intent(inout)` |
|  24722 | `INTENT(INOUT)` |
|   5410 | `INTENT(inout)` |
|   5117 | `INTENT( INOUT )` |
|   1149 | `intent (inout)` |
|   1109 | `intent(INOUT)` |
|    574 | `intent(in out)` |
|    441 | `intent (in out)` |
|    310 | `Intent(inout)` |
|    212 | `Intent(InOut)` |
|    209 | `INTENT (INOUT)` |
|    205 | `INTENT(inOUT)` |
|    203 | `INTENT(INOUT )` |
|    189 | `Intent(INOUT)` |
|    141 | `INTENT(INout)` |
|    113 | `intent(inOut)` |
|    110 | `intent( inout )` |
|     71 | `INTENT(IN OUT)` |
|     58 | `INTENT(INOUT   )` |
|     55 | `INTENT (inout)` |
|     41 | `Intent (InOut)` |
|     24 | `Intent( InOut )` |
|     24 | `intent (inout  )` |
|     17 | `INTENT (INout)` |
|     17 | `intent(inOUT)` |
|     16 | `INTENT ( INOUT )` |
|     12 | `INTENT( inout )` |
|     10 | `INTENT(INOUT  )` |
|     10 | `intent(inout   )` |
|      8 | `intent (INOUT)` |
|      8 | `intent (inout   )` |
|      5 | `Intent (INOUT)` |
|      4 | `INTENT(INOUT      )` |
|      4 | `INTENT (IN OUT)` |
|      4 | `INTENT(InOut)` |
|      4 | `intent(inout  )` |
|      3 | `intent (inout )` |
|      3 | `intent ( inout)` |
|      2 | `INTENT( INOUT)` |
|      2 | `INTENT( INOUT  )` |
|      2 | `INTENT ( inout)` |
|      1 | `INTENT(  INOUT  )` |
|      1 | `INTENT (INOUT   )` |
|      1 | `INTENT(INout )` |
|      1 | `INTENT(INout   )` |
|      1 | `INTENT(inout  )` |
|      1 | `Intent(InOUt)` |
|      1 | `Intent(Inout)` |
|      1 | `intent(InOut)` |
|      1 | `intent( InOut )` |
|      1 | `intent(Inout)` |
|      1 | `intent(inout )` |

That's 52 different ways of writing it, although there is a pretty clear winner -- I think we can get away without providing an option, and in fact we should probably recommend exactly `intent(inout)`, `intent(in)`, `intent(out)`

imciner2:
I'd be curious about statistics on `goto`. I know I would prefer one word, but that is because I come from more of a C background where there is no space.

LiamPattinson:
I've had a go at generalising the rule for all of the other multi-keywords (except block data). Just need to add more endpoints and test them all one-by-one...

ZedThree:
> I'd be curious about statistics on `goto`. I know I would prefer one word, but that is because I come from more of a C background where there is no space.

This took me a couple of attempts -- plain text search also picks up comments, and tree-sitter query is quite slow at this scale -- but it seems there's a slight preference for `go to`.

| count | |
|-------|-|
|  18559 | `GO TO` |
|  18531 | `goto` |
|   8335 | `go to` |
|   4583 | `GOTO` |
|    312 | `Goto` |
|      3 | `Go To` |
|      2 | `go  to` |

I think it's probably safe to recommend `goto` -- although of course we should also recommend to not use it at all!

LiamPattinson:
Oh wow, those are some disturbing stats when taken in combination: for every 3-4 times somebody has written `intent(inout)`, there's a `goto` lurking out there somewhere.

LiamPattinson:
Amended the last commit to include more entrypoints and a now very complicated test (I've never used things like `select type` so I'm learning a lot here!). I still have 8 things to check by my count.

LiamPattinson:
I think I'm finally there!

The implementation for `keyword-has-whitespace` is pretty ugly, largely because `inout` doesn't behave like any of the other double keywords, and removing the whitespace between two keywords is much harder to do than simply adding a space. I spent a little too long trying to account for a very rare edge case that we're unlikely to ever see anyway:

```f90
integer, intent(in & ! Comments are allowed here...
  out) :: x  ! Totally valid Fortran
```

fortitude/src/rules/style/keywords.rs
```diff
@@ -0,0 +1,276 @@
...
+use ruff_macros::{derive_message_formats, ViolationMetadata};
+use ruff_source_file::SourceFile;
+use ruff_text_size::{TextRange, TextSize};
+use std::str::FromStr;
+use tree_sitter::Node;
+
+// TODO Support for `endfile`/`end file`
+
+#[derive(strum_macros::EnumString, strum_macros::Display)]
+#[strum(serialize_all = "lowercase", ascii_case_insensitive)]
```
ZedThree, line 15: `ascii_case_insensitive` is the magic thing that preserves existing case here?

ZedThree (approved):
Looks good, a bit of tricky one to implement! Thanks @LiamPattinson!

fortitude/src/rules/style/keywords.rs
LiamPattinson: No, in this case it just means `DoubleKeyword::from_str(text)` is case insensitive, and the combination with `serialize_all = "lowercase"` means `double_keyword.to_string()` is always lowercase. To preserve case in the fixes, I do it by just inserting a space/removing whitespace at the relevant offset. The error messages and fix titles both use lowercase regardless of what the user wrote. That could be corrected fairly easily though.

fortitude/src/rules/style/end_statements.rs
```diff
@@ -101,6 +101,7 @@ impl AstRule for UnnamedEndStatement {
     }
 
     fn entrypoints() -> Vec<&'static str> {
+        // TODO: Can also do end associate, end block, end do, etc, but these are optionally named
```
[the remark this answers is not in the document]
LiamPattinson: Yes, for that reason I don't think it's a very high priority. I can remove the TODO note if it's just adding unnecessary clutter.

fortitude/src/rules/style/keywords.rs
ZedThree: Ah, yes, I see. No, that's fine

fortitude/src/rules/style/end_statements.rs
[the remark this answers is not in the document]
ZedThree: Yeah, I think remove the TODO and make an issue

PR #394, opened 2025-03-28
Raise `SyntaxError` for missing nodes

### LINKED ISSUE #384: Deal with tree-sitter `MISSING` nodes
Currently, we don't handle these, but `root.has_error()` does show them, so we print a useless warning about throw away other violations, even though we don't.

ZedThree:
Closes #384

Along with error nodes, tree-sitter will also insert "missing" nodes
to recover from "certain kinds" of syntax errors: see
[Node::is_missing](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.is_missing)
and [The MISSING
node](https://tree-sitter.github.io/tree-sitter/using-parsers/queries/1-syntax.html#the-missing-node).

Unlike `ERROR` nodes, missing nodes don't appear as `MISSING` and have
to be manually checked.

We _could_ also report the inserted node, but from experience this
isn't necessarily very useful information, and may be quite
misleading.

### DIFF
--- a/fortitude/resources/test/fixtures/error/E001.f90
+++ b/fortitude/resources/test/fixtures/error/E001.f90
@@ -1,0 +1,7 @@
+program foo
+  implicit none
+contains
+  subroutine bar(x, y, )
+    integer, intent(in) :: x, y
+  end subroutine bar
+end program foo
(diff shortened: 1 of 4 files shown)


LiamPattinson:
Thanks for this!

Do we know if missing nodes can lead to a corrupted AST? If everything after a missing node is okay, then we could adjust the bit that throws away subsequent violations and forbids fixes. That might require making up a new `Error` type -- maybe `something-missing` or something like that?

ZedThree:
I'm not sure, but I think we should err on the side of caution for now.

I did wonder about making this a separate rule, but I worry that might be confusing to have two kinds of syntax error. Might be worth trying it, if only to easily find some existing examples and if they affect the rest of the AST

PR #400, opened 2025-04-02
docs: Include changelog on website

### LINKED ISSUE #397: Release notes on website
Might be as simple as linking the changelog, but we should have all the release notes on the website

ZedThree:
Closes #397


LiamPattinson (approved):
Nice! Love it when something complicated turns out to have an easy implementation like this.

ZedThree:
Yep, might do the same for ruff if they want it!

PR #404, opened 2025-04-03
WIP: add formatting rules and associated tests

TomMelt:
**Note: This PR is WIP**

### Summary
I added new formatting rules and associated tests.

We would need to decide if we are happy with the new formatting rules as applied in the snapshots.

Few things to note:
- The rules do need refactoring / tidying
- AFAIK it is not possible to add multiple new lines e.g., if we want 2 new lines before each subroutine...
- The treatment of end statements e.g., `endprogram` is problematic :disappointed: due to `end program` being treated as one node when it is written as one word. Again AFAIK it is not possible to add a space to the middle of a node.

### To Do
- [x] extend formatting rules in `fortran.scm`
- [x] add associated tests (using `cargo insta` framework)
- [ ] check all snapshots are satisfactory


LiamPattinson:
@TomMelt Amazing, this is looking great!

> AFAIK it is not possible to add multiple new lines e.g., if we want 2 new lines before each subroutine...

I'll try to find time tomorrow to play around with it and see if I can find a solution to this. If it's not possible, I imagine it wouldn't be too hard to fix it up in post-processing.

> The treatment of end statements e.g., `endprogram` is problematic 😞 due to end program being treated as one node when it is written as one word. Again AFAIK it is not possible to add a space to the middle of a node.

I just added a [pair of style rules](https://github.com/PlasmaFAIR/fortitude/blob/main/fortitude/src/rules/style/keywords.rs) that adds a space to `endprogram`, `endif`, etc, and removes the spacing in `in out` and `go to`. It may be possible to reuse some of that functionality in a post-processing step. I imagine that quite a few style rules may end up being reused here, e.g. the one to remove pointless semicolons.

fortitude/src/snapshots/fortitude__format__test__use-operator-and-assignment.f90.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: fortitude/src/format.rs
+expression: string
+---
+module test
+  useother, only:operator(), assignment( = )public :: assignment( = )
+  privateoperator()
+end module
```
LiamPattinson, line 8: Looks like there's something strange going on in this one. Maybe `public_statement` and `private_statement` need an `@append_space`, and `use_statement` needs an `@append_hardline`?

fortitude/src/snapshots/fortitude__format__test__functions-simple.f90.snap
```diff
@@ -0,0 +1,22 @@
+---
+source: fortitude/src/format.rs
+expression: string
+---
+real(8) function current_time()
+  ! returns the wall clock time in seconds.
+  usempicurrent_time = mpi_wtime()
+end function
+type(object)function new_object
```
LiamPattinson, line 9: I think `derived_type` needs an `@append_space`

TomMelt:
Thanks @LiamPattinson for the initial suggestions. I should say it's not finished yet. I still have a few more rules to implement. But I just wanted to update you and @ZedThree on my progress so far

fortitude/src/snapshots/fortitude__format__test__functions-simple.f90.snap
TomMelt: thanks done in d01c192

fortitude/src/snapshots/fortitude__format__test__semicolon-in-interface.f90.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: fortitude/src/format.rs
+expression: string
+---
+interfaceoperator();
```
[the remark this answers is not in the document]
TomMelt: hmmm this example makes me wonder if we just want not allow semicolons :grimacing: 

For now I will fix the `interface` part but not sure how to treat semicolons...

See here for more info https://gcc.gnu.org/onlinedocs/gcc-3.0.4/g77_12.html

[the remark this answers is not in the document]
ZedThree: We have a couple of rules for dealing with superfluous semicolons, we could run the fixes as part of the preprocessing for the formatting

TomMelt:
@ZedThree @LiamPattinson I have done a final pass for now. The rules need generalizing and tidying up still but I am happy with most of the formatting now. Lmk what you think

fortitude/src/snapshots/fortitude__format__test__functions-simple.f90.snap
```diff
@@ -0,0 +1,23 @@
...
+end function
+type(object) function new_object
+  ! returns the current date string
+  TYPE(object) :: obj
+  obj%counter = 0
+  new_object = obj
+end function
+integer(8) function exported_fun(arg1, arg2) bind(c)
+end function
+complex(8) function exported_fun2() bind(c, name = "f90_func")
```
ZedThree, line 18: Could we remove the space from the keyword argument?

```suggestion
complex(8) function exported_fun2() bind(c, name="f90_func")
```

fortitude/src/snapshots/fortitude__format__test__submodule-definition-simple.f90.snap
```diff
@@ -0,0 +1,12 @@
+---
+source: fortitude/src/format.rs
+expression: string
+---
+SUBMODULE(BoundingBox_Method) ConstructorMethods
```
ZedThree, line 5: Looks like no one agrees if there should be whitespace after `submodule`:

| count | |
|-|-|
|    651| `submodule (`|
|    604| `submodule(`|
|     11| `submodule  (`|
|      5| `subMODULE(`|
|      2| `SUBMODULE(`|
|      2| `Submodule(`|

Do we want it or not? I can never remember the syntax, but I think `submodule(parent)` is actually clearer and easier to remember

fortitude/resources/format/fortran.scm
```diff
@@ -1,29 +1,127 @@
+; Don't format strings or comments
 [
- (program_statement)
- (module_statement)
- (variable_declaration)
- (assignment_statement)
- (if_statement)
- (do_loop_statement)
- (elseif_clause)
+ (string_literal)
```
ZedThree, line 3: Should `(comment)` be in here too?

ZedThree (changes requested):
Fantastic, thanks @TomMelt! Could you have a stab at the keyword argument whitespace?

As this is into `format-command`, we can go ahead and merge it into that, and continue development there

fortitude/src/snapshots/fortitude__format__test__functions-simple.f90.snap
TomMelt: sure I will take a look. Shouldn't be too tricky :crossed_fingers:

fortitude/resources/format/fortran.scm
TomMelt: yes, it should. We may want to consider a post-processing rule for enforcing a space between the exclamation point "!" and the comment text though. AFAIK we cannot do this in topiary.

Added `(comment)` in 2e421ad

fortitude/src/snapshots/fortitude__format__test__submodule-definition-simple.f90.snap
TomMelt: IMO I agree that `submodule(parent)` is cleanest, but happy to change it.

LiamPattinson: I personally prefer `submodule(parent)`, but then I also prefer `allocate (...)`, `write (...)`, `if (...)`, etc, so I'm not terribly consistent on this point.

imciner2: I'm in the camp of prefering no space before the opening parenthesis, but a space after it, e.g., `submodule( parent )`. Just goes to show how subjective all of this really is.

ZedThree (approved):
Thanks @TomMelt! I'm going to merge this so we can keep the `format-command` branch up-to-date

PR #407, opened 2025-04-04
Add rule for `missing-end-label`

### LINKED ISSUE #401: `unnamed-end-statement` can be expanded
The rule currently handles constructs that must be named but may be unnamed in their end statement, e.g:

```f90
function my_function(...)
  ...
end function
! instead of 'end function myfunction'
```

It doesn't handle constructs that can be optionally named:

```f90
named_do: do i=1,N
  ...
end do
! should be: 'end do named_do'
```

The latter should be a compile error, so this isn't a strong priority, but it should be possible for us to handle it. tree-sitter can parse it without issue.

ZedThree:
Closes #401

This is a compile-time error, but we can catch it, so why not


fortitude/src/rules/correctness/exit_labels.rs
```diff
@@ -149,6 +149,100 @@ impl AstRule for ExitOrCycleInUnlabelledLoop {
...
+    ) -> Option<Vec<Diagnostic>> {
+        let src = src.source_text();
+        // Skip unlabelled loops
+        let label = node
+            .child_with_name("block_label_start_expression")?
+            .to_text(src)?
+            .trim_end_matches(':');
+
+        let end = match node.kind() {
+            "select_case_statement" | "select_type_statement" | "select_rank_statement" => {
```
LiamPattinson, line 195: I missed `select_rank_statement` in `keywords-missing-space`, I'll have to raise an issue for that

fortitude/src/rules/mod.rs
```diff
@@ -100,6 +100,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Correctness, "132") => (RuleGroup::Stable, Ast, Optional, correctness::accessibility_statements::DefaultPublicAccessibility),
         (Correctness, "141") => (RuleGroup::Stable, Ast, Default, correctness::exit_labels::MissingExitOrCycleLabel),
         (Correctness, "142") => (RuleGroup::Preview, Ast, Optional, correctness::exit_labels::ExitOrCycleInUnlabelledLoop),
+        (Correctness, "143") => (RuleGroup::Preview, Ast, Optional, correctness::exit_labels::MissingBlockLabel),
```
LiamPattinson, line 103: Could this be switched on by default? I can't imagine this one would be offensive to anybody.

LiamPattinson (commented):
Cheers for this! Just a couple of comments, and one quite annoying request: could it be renamed `missing-end-label`? To me, `missing-block-label` suggests that rule is about catching unlabelled `block` than unlabelled `end block/associate/select`. The name would be confusingly similar to `unnamed-end-statement`, but I think that might be unavoidable.

fortitude/src/rules/correctness/exit_labels.rs
LiamPattinson: Actually, it seems `selectrank` isn't valid, but tree-sitter parses it as if it were.

LiamPattinson (approved):
Cheers, all looks good

Issue #422, opened 2025-04-28
Prefer strings with double quotes


LiamPattinson:
A fairly simple one that should be easy to auto-fix. Prefer strings of the form `"..."`, unless the string contains double quotes, in which case single quotes are required.

PR #423, opened 2025-04-29
Add rule `single-quote-string`

LiamPattinson:
Solve https://github.com/PlasmaFAIR/fortitude/issues/422

Ignores strings containing double quotes like `'hello "world"'`, and doesn't allow fixes when single quote strings contain escaped single quotes (`'hello ''world'''`). It would be possible to go through and reduce all of them to un-escaped quotes, but I think this would be better left to a separate rule (e.g. https://docs.astral.sh/ruff/rules/avoidable-escaped-quote/).

One case I wasn't able to test for was multiline strings with comments in the middle of them, as tree-sitter-fortran doesn't support them:

```f90
print *, 'Hello, &
  ! Helpful comment
          & World!'
```

I'm not 100% sure if this is supported by the standard, but gfortran doesn't complain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,9 +146,9 @@ checksum = "79296716171880943b8470b5f8d03aa55eb2e645a4874bdbb28adb49162e012c"
 
 [[package]]
 name = "cc"
-version = "1.1.31"
+version = "1.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


ZedThree:
Unfortunately that is in the standard, and unfortunately I have found a couple of real world uses. I did think I'd got that working in tree-sitter which is a little upsetting

ZedThree (approved):

LiamPattinson:
Yes, I thought it was working in an earlier build too. I'm not sure when that changed.

Issue #406, opened 2025-04-04
Misleading semicolon in one-line `if`


ZedThree:
The gfortran docs have this really nice example:

> This restriction avoids the confusion that can result when reading a line such as:
> 	
>     IF (VALIDP) CALL FOO; CALL BAR
>
> Some readers might think the `CALL BAR` is executed only if `VALIDP` is .TRUE., while others might assume its execution is unconditional.

We currently have [multiple-statements-per-line](https://fortitude.readthedocs.io/en/stable/rules/multiple-statements-per-line/#multiple-statements-per-line-s082), but maybe we should also have `misleading-semicolon` as a more specific rule? This would be a strict subset of `multiple-statements-per-line`, but under `correctness`. It does mean that the snippet above would get flagged twice if both rules were on, but maybe that's fine?

LiamPattinson:
I would support this. Would we be able to add an unsafe fix to this that would rewrite it as:

```f90
if (VALIDP)
  call FOO
  call BAR
end if
```

or would the ambiguity mean we would just have to leave it to the user?

---

A related one I've seen in the wild is a misleading line continuation in an if statement:

```f90
if (condition) &
  call something(x, y, z)
  call something_else(x, y, z)  ! Very misleading indentation!
call more_code(x, y, z)
```

I'm not really sure how we would detect that in general, except by mandating the use of `if (...) then` in all cases.

ZedThree:
It's a safe fix, it's always:

```f90
if (VALIDP) then
  call FOO
end if

call BAR
```

Similar to your second example.

Yeah, I'm not sure about how to catch your second one. Maybe the logic is:

1. is this an inline `if`?
2. does it have a line continuation after the condition? (not even sure how we catch this!)
3. calculate the indentation of the next physical line
4. check the indentation of the next logical line, does it match step 3?
5. warn

LiamPattinson:
> It's a safe fix

Ah, of course, I was thinking about fixing it to what the user likely intended, not what they'd actually written!

Issue #65, opened 2024-10-04
Prefer `if ... then` over inline `if`


LiamPattinson:
For consistency and readability, prefer this:

```f90
if (.not. allocated(x)) then
  allocate (x(n))
end if
```

over this:

```f90
if (.not. allocated(x)) allocate (x(n))
```

ZedThree:
This is probably a "pedantic" warning rather than a default one!

LiamPattinson:
I agree, but I've seen this sort of thing argued in C/C++ style guides, so I think it would make sense to have it as an option

ZedThree:
I wrote something out arguing that it's different, but then I convinced myself it isn't really!

Maybe if the statement is split over multiple lines it's more dangerous than if it's all on one line. I'm not sure how easy that is to catch

Beliavsky:
I prefer the one-line if to the block and would therefore prefer that a warning about the one-line if not be turned on by default.

LiamPattinson:
For future reference, this can be used to do the fix:

```rust
/// Given an if statement node, convert it from a one-line if statement
/// to a block if statement.
/// Returns None if the node is already a block if statement.
pub fn ifthenify(node: &Node, src: &SourceFile) -> Option<String> {
    let source_text = src.source_text();
    // check that the node is an if statement without an end if statement
    if !inline_if_statement(node) {
        return None;
    }
    // Divide the if statement into everything up to the end of the condition
    // and the body.
    let condition_node = node.child_with_name("parenthesized_expression")?;
    let condition_end_byte = condition_node.end_byte();
    let mut body_start_byte = condition_end_byte;
    let source_bytes = source_text.as_bytes();
    while body_start_byte < node.end_byte() && source_bytes[body_start_byte].is_ascii_whitespace() {
        body_start_byte += 1;
    }
    // Concatenate bytes to the end of the condition, " then\n", the body, and
    // "\nend if". Take care to get the indentation right.
    let if_start_byte = node.start_byte();
    let if_end_byte = node.end_byte();
    let indentation = node.indentation(src);
    let mut correction = String::new();
    correction.push_str(&source_text[if_start_byte..condition_end_byte]);
    correction.push_str(" then\n");
    correction.push_str(&indentation);
    correction.push_str("  "); // Assume two-space indent for the if body
    correction.push_str(&source_text[body_start_byte..if_end_byte]);
    correction.push('\n');
    correction.push_str(&indentation);
    correction.push_str("end if");
[...]

PR #426, opened 2025-05-01
Add rule for misleading semicolon following if statement

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/406

I'm not sure of the name for this one.

The added `ifthenify` function could also be used to easily implement https://github.com/PlasmaFAIR/fortitude/issues/65

### DIFF
--- a/fortitude/src/rules/correctness/conditionals.rs
+++ b/fortitude/src/rules/correctness/conditionals.rs
@@ -1,7 +1,6 @@
 use crate::ast::FortitudeNode;
 use crate::settings::Settings;
 use crate::{AstRule, FromAstNode};
-use anyhow::{Context, Result};
 use ruff_diagnostics::{AlwaysFixableViolation, Diagnostic, Edit, Fix};
 use ruff_macros::{derive_message_formats, ViolationMetadata};
 use ruff_source_file::SourceFile;
@@ -47,10 +46,7 @@ impl AlwaysFixableViolation for IfStatementSemicolon {
 impl AstRule for IfStatementSemicolon {
     fn check(_settings: &Settings, node: &Node, src: &SourceFile) -> Option<Vec<Diagnostic>> {
         // If this is an `if (...) then` construct, exit early
-        if node
-            .named_descendants()
-            .any(|n| n.kind() == "end_if_statement")
-        {
+        if !inline_if_statement(node) {
             return None;
         }
 
@@ -72,7 +68,7 @@ impl AstRule for IfStatementSemicolon {
                 }
                 break;
             }
-            let if_edit = node.edit_replacement(src, ifthenify(node, src).ok()?);
+            let if_edit = node.edit_replacement(src, ifthenify(node, src)?);
             let indentation = node.indentation(src);
             // To replace the semicolon node, we should also replace all
             // trailing whitespace.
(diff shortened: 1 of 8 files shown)


fortitude/src/rules/correctness/conditionals.rs
```diff
@@ -0,0 +1,134 @@
...
+    // Concatenate bytes to the end of the condition, " then\n", the body, and
+    // "\nend if". Take care to get the indentation right.
+    let if_start_byte = node.start_byte();
+    let if_end_byte = node.end_byte();
+    let indentation = node.indentation(src);
+    let mut correction = String::new();
+    correction.push_str(&source_text[if_start_byte..condition_end_byte]);
+    correction.push_str(" then\n");
+    correction.push_str(&indentation);
+    correction.push_str("  "); // Assume two-space indent for the if body
```
LiamPattinson, line 128: I should probably adjust this to account for whitespace between the end of the condition and the start of the body.

fortitude/src/rules/correctness/conditionals.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn fix_title(&self) -> String {
+        "Convert to `if(...) then` statement".to_string()
+    }
+}
+impl AstRule for IfStatementSemicolon {
+    fn check(_settings: &Settings, node: &Node, src: &SourceFile) -> Option<Vec<Diagnostic>> {
+        // If this is an `if (...) then` construct, exit early
+        if node
+            .named_descendants()
+            .any(|n| n.kind() == "end_if_statement")
```
ZedThree, line 52: Rather than iterating over all the descendants, it might be sufficient to check `child(2)`, which should be the unnamed node `then`?

ZedThree (commented):
Just checked this on some real projects, I think it would be sensible to add a the following sorts of constructs to the tests, particularly for the fixes:

```f90
! Check triggers multiple times on same line, does the fix work ok?
if (foo) a = b; if (bar) c = d; if (zing) e = f;

! How does this mess up the indentation in the fix?
! Of course, this should also trigger the multiple-statements rule
do i = 1, 3; if (foo(i) /= i) foo(i) = i; end do
```

fortitude/src/rules/correctness/conditionals.rs
LiamPattinson: Thanks, corrected in https://github.com/PlasmaFAIR/fortitude/pull/426/commits/ba5bbd1154741dd9d626abc96311728eafda03cd

fortitude/src/rules/correctness/conditionals.rs
LiamPattinson: Corrected in https://github.com/PlasmaFAIR/fortitude/pull/426/commits/dcf48cc92ecfc386551edd3425f750a6a2cbea09

LiamPattinson:
One issue with this implementation is that it replaces the inline if with a `if then` block, and it just occurred to me that the minimum I needed to do was replace the semicolon with a newline. To preserve the user's intent, that should probably be preferable. I'm expecting things like:

```f90
if (allocated(x)) x(0) = 0.0; x(1) = 0.0; x(2) = 0.0
```

It would be better to rewrite this as:

```f90
if (allocated(x)) x(0) = 0.0
x(1) = 0.0; x(2) = 0.0  ! this is still an error, but at least it's no longer misleading
```

---

```
! Check triggers multiple times on same line, does the fix work ok?
if (foo) a = b; if (bar) c = d; if (zing) e = f;
```

I've added a new test that checks that the combined fixes work okay, and it looks good to me. 

```
! How does this mess up the indentation in the fix?
! Of course, this should also trigger the multiple-statements rule
do i = 1, 3; if (foo(i) /= i) foo(i) = i; end do
```

It looks pretty horrible, but at least it doesn't break anything!

This is a case where I would argue this rule gives a false positive, as the meaning of that line is unambiguous. However, I'm not sure how best to handle it. Maybe just a check if the next node kind starts with `"end_"`?

LiamPattinson:
Updated to leave the one-line if statement alone and simply replace the semicolon with a newline. I've also allowed the following node kind to start with `end_` to permit the use of this construct in the do_loop one-liner, though I'm unsure if that's wise.

ZedThree:
I'm not sure about allowing the `end` in the next statement, here's another real world example:

```
77 | if (allocated(nsd)) then
78 | deallocate(nsd,           stat=ios); if (ios /= 0) call error('dealloc (nsd), unable to deallocate'); end if
   |                                                                                                     ^ C151
```

I guess false negatives are better than false positives, and this will still get picked up by the multiple-statements rule.

LiamPattinson:
Okay, that's a fair counter example. I guess this might just be a case where an `! allow` comment would be warranted.

ZedThree (approved):

PR #427, opened 2025-05-01
Enable reading non-utf8 files

### LINKED ISSUE #199: Handle non-utf-8 files
Some small portion of files aren't utf-8, which currently fortitude cannot handle. We give a nice error message though

ZedThree:
Closes #199

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -372,6 +372,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a357d28ed41a50f9c765dbfe56cbc04a64e53e5fc58ba79fbc34c10ef3df831f"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75030f3c4f45dafd7586dd6780965a8c7e8e285a5ecb86713e63a79c5b2766f3"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
+name = "encoding_rs_io"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cc3c5651fb62ab8aa3103998dade57efdd028544bd300516baa31840c252a83"
+dependencies = [
+ "encoding_rs",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


LiamPattinson:
Oh wow, once again I'm surprised at how easy it is to solve issues like this in Rust.

I tried applying fixes to a UTF-16 file, and it turned it into a UTF-8 file. I personally see this as an added bonus, but how will this work if we start from an encoding that includes a non-UTF character? Would we still just get a parsing error like we currently do for UTF-16?

ZedThree:
`String::from_utf8_lossy` replaces non-utf8 characters with `U+FFFD REPLACEMENT CHARACTER` (�), so I was thinking we could just have a rule that detects presence of this character and flags it as non-utf8. Getting the original bytes will be a little tricky, we probably have to re-read the file and find the corresponding offset.

The only places Fortran supports non-ANSI characters are comments and string literals, so there _shouldn't_ be any syntax errors from this conversion -- however, we might mess up string literals if someone is using non-unicode (e.g. latin-1/ISO-8859/Windows-1252). This is likely to be pretty rare. The standard only supports ASCII and what it calls "ISO 10646" `character` kinds (the standard says ISO 10646, which is not an encoding, means UCS-4, or what people now call UTF-32), and both of these can be losslessly converted to UTF-8.

I'm not actually terribly happy about this PR, it ends up doing an unnecessary copy in 99.99% of cases. I think we probably want to try `std::fs::read_to_string`, and if the error says non-utf8, try this method instead and raise either a warning or diagnostic.

ZedThree:
I've gone off this PR. I don't like the idea of silently converting encodings. It also turns out intel doesn't support utf16 files at all, so maybe what's best is a couple of rules:

1. utf16: as this isn't completely portable, and is also completely safe to convert, we diagnose and autofix
2. non-unicode encodings: this is almost certainly due to a comment, and we can suggest a fix, but it might be wrong

ZedThree:
I've spent too much time on this already, and it's getting quite complicated. Because this happens before `check_file`, we'd have to reinvent a whole bunch of fixing machinery. Given there's only ~50 files out of 80k in the corpus that are non-utf8, probably not worth the effort until we get requests from users!

PR #428, opened 2025-05-02
Add rule `misleading-inline-if-continuation`

LiamPattinson:
Solves https://github.com/PlasmaFAIR/fortitude/issues/65

Partly.

Rather than flagging up all inline if statements, which would be a fairly controversial style rule, this solves the subset of cases where an inline if statement spills over multiple lines.
I decided to go for the reduced problem, as I would prefer not to flag up things like:

```f90
if (allocated(A)) deallocate(A)
```

However, I would argue code like the following goes beyond a style violation and becomes a correctness problem:

```f90
if (condition) &
  call do_something()
```


LiamPattinson:
I'm not sure why this is failing on Windows. It also looks like it's running the test multiple times and really slowly.

ZedThree:
My first thought would be a line-ending issue. Windows uses `\r\n` rather than just `\n`. I'm sure we've encountered this before, can't remember what the solution was though!

LiamPattinson:
It seems to be leaving in an extra line saying `"Hello world!"` in Windows, not sure why. I'll keep at it.

LiamPattinson:
That seems to have done it. It seems it's okay if we're only adding new newlines, but if we're replacing existing ones then we need to be careful not to change them.

ZedThree:
Ruff has a `Stylist` struct which detects the kind of line ending used in a file, represented as a `LineEnding` enum (which also supports the ancient macos `\r`-only). As we start implementing more multiline fixes, we probably want to get something like this in ourselves to avoid these issues!

LiamPattinson:
I've updated this to permit multi-line conditions, as I think the following actually isn't a problem:

```f90
if (long_condition_1 &
    .and. long_condition_2) call something()
```

A remaining question for me is which of the following we really care about:

```f90
! Break immediately after the condition
if (condition) &
  call something()

! Break in the middle of a statement
if (condition) call &
  something()
```

I'm leaning towards being stricter about it and only forbidding the former.

LiamPattinson:
I've tightened up the scope of the rule and the tests. Is anything else needed for this PR?

ZedThree (approved):
Nice, looks good!

PR #430, opened 2025-05-07
Don't raise `SyntaxError` for missing nodes if rule not enabled

### LINKED ISSUE #429: Can't ignore syntax errors
I don't think this was a deliberate change, but it seems no longer possible to ignore `E001`, `syntax-error`

This is a bit of a pain when checking against the corpus, unless we can also filter out files that need the preprocessor!

ZedThree:
Fixes #429


LiamPattinson (approved):
All looks good! :+1:

PR #431, opened 2025-05-07
Add rule for non-portable short circuiting use of inquiry functions

### LINKED ISSUE #67: Bugprone order of evaluation in logical expression
Fortran doesn't have short-circuiting in logical expression, so the order of evaluation of subexpressions is no guaranteed. A common mistake is checking optional arguments:

```fortran
if (present(arg) .and. arg > 42) then
...
```

This may try to evaluate `arg > 42` before `present(arg)`, which can result in a (essentially) a nullptr dereference if `arg` is not present.

Correct:

```fortran
if (present(arg)) then
  if (arg > 42) then
    ...
```

More generally, this is also an issue for multiple (non-pure) function calls in the same expression. Although that does require working out if a `call_expression` is a function or an array! Yey ambiguous syntax!

ZedThree:
Closes #67

I'm not completely sold on the name, and could be persuaded to rename
it. I'm also not sure I've written a great summary of why this is bad.

Testing on the corpus was very useful to find some edge cases I wasn't
catching properly.

I got stuck down a very painful rabbit hole trying to add a predicate
to `FortitudeNode::descendants` that would skip iterating a node _and_
its descendants. I think it might involve rewriting the
`DepthFirstIterator` quite a lot -- something I hadn't realised is
that that (deliberately?) skips the node itself, which is possibly
fine, but means it can't be rewritten in a recursive way, which might
easier for adding a predicate.

### DIFF
--- a/fortitude/src/rules/correctness/nonportable_shortcircuit_inquiry.rs
+++ b/fortitude/src/rules/correctness/nonportable_shortcircuit_inquiry.rs
@@ -9,8 +9,7 @@ use ruff_text_size::TextRange;
 use tree_sitter::Node;
 
 /// ## What it does
-/// Checks for use of optional dummy arguments in the same logical expression as
-/// `present(<arg>)`.
+/// Checks for use of a variable in the same logical expression as "definedness" inquiry.
 ///
 /// ## Why is this bad?
 /// Unlike many other languages, the Fortran standard doesn't mandate (or prohibit)
@@ -25,6 +24,9 @@ use tree_sitter::Node;
 /// languages). Unfortunately, any `else` branches may need to be duplicated or
 /// refactored to accommodate this change.
 ///
+/// This lack of short-circuiting also affects other inquiry functions such as
+/// `associated` and `allocated`.
+///
 /// ## Example
 /// Don't do this:
 /// ```f90
@@ -37,6 +39,8 @@ use tree_sitter::Node;
 ///     example = 1
 ///   end if
 /// ```
+/// The compiler may or may not evaluate `arg1 > 2` _even if_ `present(arg1)` is
+/// false. This is a runtime error, and may crash your program.
 ///
 /// Use instead, noting that we either need to duplicate the `else` branch, or refactor
 /// differently:
@@ -62,24 +66,25 @@ use tree_sitter::Node;
 ///
 ///   example = present(arg1) ? (arg1 > 2 ? arg1 * arg1 : 1) : 1
 /// ```
-/// Note that it's still not possible to use a compound logical expression here, so we
-/// still duplicate the default value.
+/// Note that although the true/false arms of the conditional-expression are lazily
+/// evaluated, it's still not possible to use a compound logical expression here, so we
+/// still must have a nested expression and duplicate the default value.
 ///
 /// ## References
 /// - <https://www.scivision.dev/fortran-short-circuit-logic/>
 #[derive(ViolationMetadata)]
-pub(crate) struct BugproneOrderOfEvaluation {
+pub(crate) struct NonportableShortcircuitInquiry {
     arg: String,
     // Useful for when we get multiple notes in DiagnosticMessages
     #[allow(dead_code)]
     present: TextRange,
 }
 
-impl Violation for BugproneOrderOfEvaluation {
+impl Violation for NonportableShortcircuitInquiry {
     #[derive_message_formats]
     fn message(&self) -> String {
         let Self { arg, .. } = self;
-        format!("use of optional `{arg}` in same logical expression as `present({arg})`")
+        format!("variable inquiry `present({arg})` and use in same logical expression")
     }
 }
 
@@ -107,7 +112,7 @@ fn present_call(expr: &Node, src: &str) -> Option<PresentCall> {
     })
 }
 
-impl AstRule for BugproneOrderOfEvaluation {
+impl AstRule for NonportableShortcircuitInquiry {
     fn check<'a>(
         _settings: &Settings,
         node: &'a Node,
--- a/fortitude/src/rules/correctness/nonportable_shortcircuit_inquiry.rs
+++ b/fortitude/src/rules/correctness/nonportable_shortcircuit_inquiry.rs
@@ -25,7 +25,7 @@ use tree_sitter::Node;
 /// refactored to accommodate this change.
 ///
 /// This lack of short-circuiting also affects other inquiry functions such as
-/// `associated` and `allocated`.
+/// `associated` and `allocated` which are used to guard invalid accesses.
 ///
 /// ## Example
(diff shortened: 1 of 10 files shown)


LiamPattinson:
Cheers, I think this all looks good so far.

I'm also not sure about the name. The order of evaluations isn't the problem so much as the combination of `present(x)` and `x` in the same conditional expression. The following code flags up both if statements:

```f90
  subroutine s(i)
    integer, optional :: i
    if (i > 2 .and. present(i)) print *, "Foo"
    if (present(i) .and. i > 2) print *, "Bar"
  end subroutine s
```

I think this should be considered the correct behaviour, as it's possible a user might think short-circuiting works right-to-left, or that Fortran compilers are smart enough to check the `present(i)` call first regardless of order.

I'm similarly struggling to think of a good name though. The closest I've got is `assumed-short-circuit`, but that might be a bit too broad.


---


> I got stuck down a very painful rabbit hole trying to add a predicate
> to `FortitudeNode::descendants` that would skip iterating a node and
> its descendants. I think it might involve rewriting the
> DepthFirstIterator quite a lot -- something I hadn't realised is
> that that (deliberately?) skips the node itself, which is possibly
> fine, but means it can't be rewritten in a recursive way, which might
> easier for adding a predicate.

There is `DepthFirstIteratorExcept` that does that based on node kind, but not based on a specific node. Maybe that could be adapted to help?

Yes, I deliberately excluded the calling node when writing `descendants`. I wrote it as a variant on `node.children()`, as early on I found myself writing a lot of recursive functions that made things quite confusing. However, for parity with other Rust iterators, I think it would make sense to include the first element. I think there are a few places where skipping the top node is assumed, so rewriting it to include the top node might break things, but it's nothing that couldn't easily be fixed with `.skip(1)`.


---

Another use case of this logic could be `allocated(...)` in place of `present(...)`. Do you think it would be worth folding them into the same rule, or should they be separate rules? I can't imagine a situation where somebody would want one rule activated but not the other.

imciner2:
> Another use case of this logic could be allocated(...) in place of present(...). Do you think it would be worth folding them into the same rule, or should they be separate rules? I can't imagine a situation where somebody would want one rule activated but not the other.

Yes, IMO this check doesn't just apply to `present`, but also `allocated` and `associated`. All of these inquiry functions ensure that the data in the variable is valid before it is used, so they shouldn't be combined with variable access in the same if condition.

And thinking about that a bit more, would this be a bad access pattern as well? I think it might be, because if the variable isn't present, then it doesn't make sense to check for it to be allocated/associated, so if the changes above are made (to expand the rule), then it would be good to add this to the examples/tests to show another incorrect case.
```
logical function bad(arg1, arg2)
    integer, optional, pointer, intent(in) :: arg1, arg2(2)

    bad = .false.

    if (present(arg1) .and. associated(arg1)) then
      bad = .true.
    end if
end function bad
```

Since this just checks for the bad access pattern, I think a rule name could be `bad-access-with-inquiry`, or is that too cryptic now? (these are termed `inquiry functions` in the standard).

ZedThree:
Indeed `associated` and `allocated` have exactly the same issue (although weirdly both gfortran and intel _do_ shortcircuit for `allocated`, but not for `present` or `associated` :shrug:. flang seems to consistently _not_ shortcircuit, and lfortran consistently _does_).

They should probably be rolled into this rule, although `associated` has a two-argument version that I suspect is actually fine:

```f90
if (associated(this, arg) .and. foo(this)) then
```
Here we're checking if `this` is associated with a particular variable, and not if it's associated at all. We'd need to do a lot more expensive static analysis to check if that expression was ok or not. For now, we could just skip inquiry calls with more than one argument.

---

As for the name, taking inspiration from you both, maybe something like:

- `nonportable-shortcircuit-inquiry`
- `bad-inquiry-conditional`
- `bad-inquiry-logical-expression` (this implies we might also expand to expressions outside of `if` statements too, which would be possible I guess, although expensive. We could check `logical_expression` instead of `if_statement`)

I think the first option is perhaps the sweet spot of short and precise while still being understandable.

The message needs some work too, and being able to highlight both pieces would be very useful.

---

> There is `DepthFirstIteratorExcept` that does that based on node kind, but not based on a specific node. Maybe that could be adapted to help?

Yes, sorry, that's what I actually started from, but I think all three versions could be rolled into one.

ZedThree:
Ok pretty happy with this now:

- Name changed to `nonportable-shortcircuit-inquiry`
- Message changed to `"variable inquiry {function}({arg}) and use in same logical expression"`
- Expanded to check `associated` and `allocated`
- Rule turned on by default

I haven't expanded this to logical expressions in general, that's maybe something for the future

fortitude/src/rules/correctness/nonportable_shortcircuit_inquiry.rs
```diff
@@ -0,0 +1,182 @@
...
+impl AstRule for NonportableShortcircuitInquiry {
+    fn check<'a>(
+        _settings: &Settings,
+        node: &'a Node,
+        src: &'a SourceFile,
+    ) -> Option<Vec<Diagnostic>> {
+        let expr = node.child(1)?;
+        let text = src.source_text();
+
+        let violations = ["present", "allocatable", "associated"]
```
LiamPattinson, line 130:
```suggestion
        let violations = ["present", "allocated", "associated"]
```

LiamPattinson: Actually they were, I only noticed because an `allocated` in the tests wasn't triggering it!

ZedThree:
If I'm being really pedantic, should this be `NonPortableShortCircuitInquiry` aka `non-portable-short-circuit-inquiry`?

LiamPattinson:
I think a five-word rule name might be getting a bit verbose, so I'd be happy to pretend it's actually three with `nonportable-shortcircuit-inquiry`. `assumed-inquiry-short-circuit` would be a little shorter and I think reads better, but either works.

LiamPattinson (approved):
All looks good, cheers for this :+1:

imciner2:
> They should probably be rolled into this rule, although `associated` has a two-argument version that I suspect is actually fine:
> ```
> if (associated(this, arg) .and. foo(this)) then
> ```
> Here we're checking if `this` is associated with a particular variable, and not if it's associated at all. We'd need to do a lot more expensive static analysis to check if that expression was ok or not. For now, we could just skip inquiry calls with more than one argument.

Thinking more about this, are we sure the 2-argument case is actually safe to do other operations with in the same statement? My reading of how the 2-argument function should behave is that it doesn't require the pointer to be associated at all before calling it, so I think it actually is doing 2 checks: 1) is the pointer associated, 2) if it is associated, then is it associated with the target. So it is plausible for `this` to be not associated in that and then potentially be an issue in `foo`.

PR #436, opened 2025-05-12
Rename `if-statment-semicolon` -> `misleading-inline-if-semicolon`

LiamPattinson:
For consistency with `misleading-inline-if-continuation`.

Unsure how this should be labelled for the changelog.


ZedThree:
Because it's not actually been in a release yet, just mark it as `internal` and manually fix the changelog to have the correct name

ZedThree (approved):

PR #442, opened 2025-05-14
Preserve existing case for `unnamed-end-statement`

### LINKED ISSUE #441: Preserve capitalization in S061 autofix
If given the following code:

```fortran
MODULE testmod
   PRIVATE
   IMPLICIT NONE(external)
CONTAINS
   SUBROUTINE sub(a,b)
      IMPLICIT NONE(type,external)
      REAL, INTENT(IN) :: a
      REAL, INTENT(OUT) :: b
      b = a
   END SUBROUTINE
END MODULE
```
Fortitude detects that the `END MODULE` and `END SUBROUTINE` statements are not named. If run with the `--fix` option, the autofix correctly adds the names but changes the `END` statements to lowercase:
```fortran
MODULE testmod
   PRIVATE
   IMPLICIT NONE(external)
CONTAINS
   SUBROUTINE sub(a,b)
      IMPLICIT NONE(type,external)
      REAL, INTENT(IN) :: a
      REAL, INTENT(OUT) :: b
      b = a
   end subroutine sub
end module testmod
```
If a user is beholden to a style guide that requires capitalization of keywords, that user now has to go change all the altered `end [x]` statements back to `END [X]`. I propose that Fortitude simply append the name to the `END` statement without changing it to lowercase.

ZedThree:
Fixes #441


ZedThree:
About 1% of codes seem to use `Title Case` for keywords, so if we wanted to be really nice, we would want a utility to check this variety, returning

```rust
enum KeywordCase {
  Lower,
  Upper,
  Title,
}
```

and using something like [heck](https://crates.io/crates/heck) to handle title case

LiamPattinson (commented):
Thanks for this, all looks good! Please could we also update the `fix_title`? I would be tempted to make a quick patch release after the merge with this.

LiamPattinson (approved):

PR #445, opened 2025-05-15
Rename `single-quote-string` to `bad-quote-string`, add option

### LINKED ISSUE #440: `single-quote-string` -> `bad-quotes`?
As pointed out in this [Discourse thread](https://fortran-lang.discourse.group/t/fortitude-0-7-0-release-a-fortran-linter/9297/9), single-quote strings were the norm in 77, so some Fortran programmers have a preference for them.

Ruff has the rule [`bad-quotes-inline-string`](https://docs.astral.sh/ruff/rules/bad-quotes-inline-string/) which permits either but enforces consistency. We could modify our `single-quote-string` rule to behave similarly. The equivalent name for Fortitude would be `bad-quotes`, but I'm sure we could come up with something better than that.

ZedThree:
Includes option to set preferred quote character

Closes #440


fortitude/src/rules/style/strings.rs
```diff
@@ -2,55 +2,81 @@ use ruff_diagnostics::{Diagnostic, Edit, Fix, FixAvailability, Violation};
...
 ///
-/// Fixes are not available for single-quoted strings containing escaped single
-/// quotes (`"''"`).
+/// ## Example
+/// ```f90
+/// foo = 'bar'
+/// ```
+///
+/// Assuming `quotes` is set to `double`, use instead:
+/// ```python
```
LiamPattinson, line 29:
```suggestion
/// ```f90
```

fortitude/src/options.rs
```diff
@@ -353,3 +362,26 @@ impl KeywordWhitespaceOptions {
         }
     }
 }
+
+/// Options for the `keyword-missing-space` and `keyword-has-whitespace` rules
```
LiamPattinson, line 366:
```suggestion
/// Options for the `bad-quote-string` rule.
```

LiamPattinson (commented):
Thanks for this! Just found a couple issues with the docs, otherwise looks good.

LiamPattinson (approved):

PR #446, opened 2025-05-15
Add  rule `avoidable-escaped-quote`

### LINKED ISSUE #424: Avoidable escaped quotes
As an extension to https://github.com/PlasmaFAIR/fortitude/issues/422, and similar to: https://docs.astral.sh/ruff/rules/avoidable-escaped-quote/

Replace `"hello ""world""!"` with `'hello "world"!'`, and `'hello ''world''!'` with `"hello 'world'!"`. I imagine there might be a few tricky edge cases to consider for the auto-fix.

ZedThree:
~Builds on #445 to avoid conflicts~

Closes #424


ZedThree:
`unnecessary-escaped-quote` also isn't relevant, I forgot the doubled-character escaping doesn't apply to the other-quote-character literals

LiamPattinson (commented):
Nice one! There's a bit in the doc comment for `bad-quote-string` saying "Fixes are not currently available..." that can now be removed, but otherwise I think this looks good.

LiamPattinson:
Found an edge case that fails the tests:

```f90
  ! These work fine
  print*, ''
  print*, ""
  print*, ''''
  print*, """"
  ! These result in a single ' or " after the fix, but should have multiple
  print*, ''''''
  print*, ''''''''
  print*, """"""
  print*, """"""""
```

ZedThree:
Thanks!

LiamPattinson (approved):
Thanks, all looks good!

PR #447, opened 2025-05-19
Fix `non-portable-io-unit` triggering on any magic number

ZedThree:
This rule was triggering on _any_ magic number, which a) overlaps with `magic-io-unit`, and b) isn't helpful for most cases.

It is a _bit_ weird that with this fix we only trigger on 5 or 6, but it seems most (still) existing compilers do use these units, except for Cray, which uses 100, 101, 102. Maybe we should also check for those units? Also unit 0 for `stderr`?


LiamPattinson (approved):
Looks good! Since this was released in preview for v7.0, should we make a note to leave it in preview for v8.0?

PR #450, opened 2025-05-21
Catch stderr and Cray file units in `non-portable-io-unit`

ZedThree:
I'm not super sold on the second commit. I think it's _probably_ sensible. Maybe the option name needs improving or just removing?


imciner2:
What is the thinking behind flagging the Cray units specifically? Is this to
a) flag programs that have hard-coded their units to the Cray numbers that this is a problem if they were to use the Cray compiler, or
b) flag Cray-based code that it is non-portable when using those I/O unit numbers?

If it is the latter, then wouldn't the Cray compiler define the `iso_fortran_env` to be the correct unit numbers on it's system, so the guidance to use the standard units would be correct still?

For the former, then perhaps a companion rule would be to flag uses of `open()` that don't have the `newunit` argument to encourage the use of auto-generated unit numbers instead of hard-coding them at all.

ZedThree:
The latter. Basically, it seems that all the extant compilers (that I've
been able to check empirically or via their manual) use 0/5/6, except Cray,
which uses 100/101/102. So we should flag them if they're being used
without an explicit `open` -- which people might be doing, hence the flag
to toggle checking them.

Indeed, if we are checking them, then the guidance remains the same.
Apologies if that's now not clear -- that's unintentional!

We do have `magic-io-unit` which will still catch literals used as units.
Currently in preview.

LiamPattinson:
I would personally lean towards being strict and forbidding the use of any IO units that are known to have a special meaning in common compilers. I'm sure I've seen units 100 and 101 used in the wild for non-Cray projects, and I wouldn't have known this was a portability issue without a tool like this informing me.

Is the intent for the option to allow people to ignore this rule if they have a project that already uses units 100/101/102 everywhere and they're confident they definitely won't be using Cray compilers? If so, that could be made clearer in the docs. If people are deliberately using these units as stdin/stdout/stderr because they're primarily targetting Cray, then I think we should still be flagging this up to them.

ZedThree:
> I would personally lean towards being strict and forbidding the use of any IO units that are known to have a special meaning in common compilers. I'm sure I've seen units 100 and 101 used in the wild for non-Cray projects, and I wouldn't have known this was a portability issue without a tool like this informing me.

The behaviour on all compilers is that these units come "pre-connected", and are overridden by any calls to `open`. So:

```f90
open(100, "input")
read(100, '(a)') text
close(100)

open(6, "output")
write(6, '(a)') text
close(6)
```

is perfectly portable and well defined across all compilers. The following is not:

```f90
read(100, '(a)') text
write(6, '(a)') text
```

(and probably won't work on any single compiler!)

Given the general difficulty of tracking sequences of `open -> read/write -> close` calls, we can only check this rule on `read/write` calls.

> Is the intent for the option to allow people to ignore this rule if they have a project that already uses units 100/101/102 everywhere

Yes

> and they're confident they definitely won't be using Cray compilers?

This isn't the issue

> If so, that could be made clearer in the docs. If people are deliberately using these units as stdin/stdout/stderr because they're primarily targetting Cray, then I think we should still be flagging this up to them.

Yep, this is now the default behaviour of the rule.

It's clear from both of you that the docs need clarifying! :smile:

LiamPattinson:
I think the example you've given also illustrates a possible correctness issue:

```f90
open(6, file="output.txt")
write(output_unit, *) "Writing file..."  ! Goes to the file, not stdout
write(6, *) "Hello world"
close(6)
write(output_unit, *) "File written..."  ! Doesn't get written at all
```

As `input_unit`/`output_unit`/`error_unit` are just integer parameters, then mixing them with number literals risks a clash. I think a rule catching these in `open` statements would also be good, or just enforcing the use of `newunit` would also work.

ZedThree:
This rule doesn't have a fix for exactly that reason, because we can't robustly catch all matching uses of a unit.

`magic-io-unit` will catch `open(6)` as well as `write(6)` -- this rule is an additional warning that `write(6)` looks suspiciously like it should be `write(output_unit)`.

LiamPattinson:
Ah, I'd forgotten about `magic-io-unit`! Good to see we've finally hit the point where we have so many rules I can't keep them all in my head.

ZedThree:
Hopefully I've made the docs clearer now!

PR #451, opened 2025-05-21
Small clarification on technical details of `intent(out)`

### LINKED ISSUE #443: Missing intent (C061)
In the [C061 rule](https://fortitude.readthedocs.io/en/stable/rules/missing-intent/) it is written:
> If an intent is not specified, it will default to intent(inout).

This is incorrect, as there is at least one (subtle) difference between intent(inout) and no intent: the actual argument of an intent(inout) dummy shall be definable (it cannot be an expression, a literal, a parameter, or a variable that is itself an intent(in) dummy), whereas this is not required if the dummy has no intent.

"no intent" has actually some very valid use cases:
- for an argument that can be either intent(in) or intent(out), depending on some other arguments, but not both at the same time.
- for an argument that is just a work space (we don't care about the input and output values)
- (maybe other ones that I can't see at the moment)

Other options such as intent(none) and intent(scratch) would be useful additions to the language. But without them, not specifying an intent is justified.

Edit: Also, it is important to be aware that intent(out) can potentially change the behavior of the program compared to "no intent": an intent(out) allocatable dummy is automatically deallocated on entry of the routine if it was allocated, and a derived type dummy is automatically finalized. In an existing code, this can have unexpected consequences at runtime.

ZedThree:
Closes #443

### DIFF
--- a/docs/rules/missing-intent.md
+++ b/docs/rules/missing-intent.md
@@ -13,12 +13,15 @@ the procedure.
 Arguments with `intent(in)` are read-only input variables, and cannot be
 modified by the routine.
 
-Arguments with `intent(out)` are output variables, and their value on
-entry into the routine can be safely ignored.
+Arguments with `intent(out)` are output variables, and their value
+on entry into the routine can be safely ignored -- technically,
+they become undefined on entry, which includes deallocation and/or
+finalisation.
 
-Finally, `intent(inout)` arguments can be both read and modified by the
-routine. If an `intent` is not specified, it will default to
-`intent(inout)`.
+Finally, `intent(inout)` arguments can be both read and modified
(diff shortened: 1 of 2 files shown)


LiamPattinson:
Should we also clarify the next paragraph, as `intent(inout) != no intent`? I think it's a fair to point out that `intent(inout)` will not allow you to pass a literal or expression, while no intent will (and subsequently seg fault if you try to write to the variable).

ZedThree:
If we can do it briefly -- I'm wary of us becoming a textbook!

LiamPattinson (approved):
Cheers, I really like the wording there. I think saying "it's like inout only more dangerous" gets the point across better than a more technical description.

PR #454, opened 2025-05-26
Reorganise project to `./crates/*`

LiamPattinson:
Refactors the project to make use of a `crates/*` layout, similar to Ruff. Breaks the `fortitude` directory into three sub-components:

- `crates/fortitude`: Top level CLI
- `crates/fortitude_linter`: Most of the important stuff
- `crates/fortitude_workspace`: Settings management

This will help keep things organised as we add `fortitude_server`, `fortitude_format`, `fortitude_preprocessor`, and potentially more. It will also make it easier to adapt some of Ruff's more confusing items, such as the `Resolver` found under `ruff_workspace`.

Some files have been broken up:

- `fortitude/check.rs` is now split between `crates/fortitude/src/check.rs`, which contains the top-level 'check_files' functions, and `crates/fortitude_linter/src/lib.rs`, which contains the individual `check_file` functions.
- `fortitude/cli.rs` has been split into `crates/fortitude/src/cli.rs`, which defines the top-level command, and `crates/fortitude_linter/src/cli.rs`, which defines `CheckArgs` and `ExplainArgs`.
- Moved `DiagnosticMessage` into its own file, `crates/fortitude_linter/src/diagnostic_message.rs`. This was something I did early on in the refactor which turned out to be unnecessary, but I don't see the harm in leaving it like this.

Additional changes:

- Set all dependencies in the top-level `Cargo.toml`, use `workspace = true` elsewhere.
- Testing the whole package now requires running `cargo test --workspace`.
- Many things that were `pub(crate)` are now `pub`, and vice versa. I had to change these quite a lot throughout the refactor, and some of these changes are now likely unnecessary. I don't think we've been particularly strict with visibility before now though, so I think we can tighten this up in a later PR.

TODO (preferably in later PRs):

- Our `get_files` function contains some of the functionality of Ruff's `Resolver` under `ruff_workspace`, but is currently defined within `fortitude_linter/src/fs.rs`. We should move the `get_files` function over so it can be better reused when format and preprocessing modes are added, but we'll probably want our own version of `Resolver` by then anyway.
- In preparation for adding more advanced syntax tree traversal, we should move `crates/fortitude_linter/src/ast.rs` over to its own crate `fortitude_parser`.


crates/fortitude_linter/src/cli.rs
```diff
@@ -329,3 +210,17 @@ pub struct CheckArgs {
     #[arg(long)]
     pub statistics: bool,
 }
+
+/// Get descriptions, rationales, and solutions for each rule.
+#[derive(Debug, clap::Parser, Clone, PartialEq)]
+pub struct ExplainArgs {
```
ZedThree, line 216: Following ruff, I'd prefer to keep all the CLI args in the same file, rather than split them across crates like this

ZedThree (changes requested):
Looks good, thanks for doing this! Just one change please, I think it's best to keep all the CLI arg structs together

crates/fortitude_linter/src/cli.rs
ZedThree: Oh, this is where their other layers come in.

- `CheckCommand` has `.partition() -> (CheckArguments, ConfigArguments)`
  - `CheckArguments` is CLI only stuff and is used directly in `check`
  - `ConfigArguments` has an `ExplicitConfigOverrides` which are the CLI flags that override those in `Configuration`
  - `ConfigArguments` also has `overrides: Configuration` which allows arbitrary overrides via inline TOML on the CLI
- `ConfigArguments` (and `ExplicitConfigOverrides`) implement the `ConfigurationTransformer` trait which allows it to be turned into `Configuration`
- `Configuration` has `into_settings() -> Settings`, and `Settings` is what is actually used in `check`

This lets all the CLI stuff live in the front-end crate, and all the configuration stuff live in the workspace crate, with the cost that we need the following structs with (almost) duplicate info:

1. `CheckCommand` -- needed for CLI
2. `ExplicitConfigOverrides` -- bridges CLI and config files from CLI side
3. `Configuration` -- bridges CLI and config files from file side
4. `Settings` -- concrete settings

1, 2, & 3 have everything as an `Option` (or possibly empty `Vec`), `Settings` has no `Option`s

We have `Configuration::into_settings` taking `CheckArgs` as an argument, when to make this work like ruff, we need `impl ConfigurationTransformer for CheckArgs`

LiamPattinson: Would it be possible to delay this refactor until a future PR? It looks like it could take me an additional couple of days to work all of this out, and I don't know when I'll next have that time to spare.

ZedThree (approved):
Big change but necessary! Thanks @LiamPattinson!

Issue #61, opened 2024-10-03
`allocate`  with `stat` keyword must check `stat` variable


ZedThree:
See: https://github.com/byornski/castep-linter/blob/main/src/castep_linter/tests/allocate_stat_checked.py

LiamPattinson:
I proposed a similar thing for `open` in https://github.com/PlasmaFAIR/fortitude/issues/16, so if we can figure out the logic for one then the other should follow easily.

PR #458, opened 2025-05-27
Add rule `unchecked-stat`

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/61

Currently a work in progress. Aiming to detect any use of the stat variable that isn't another `allocate(..., stat=x)`

### DIFF
--- a/fortitude/src/rules/correctness/error_handling.rs
+++ b/fortitude/src/rules/correctness/error_handling.rs
@@ -51,18 +51,6 @@ impl StatType {
                     Err(anyhow!("Unknown subroutine: {subroutine_text}"))
                 }
             }
-            "call_expression" => {
-                let identifier_node = node.child(0).context("Could not retrieve routine name")?;
-                let identifier_text = identifier_node
-                    .to_text(src)
-                    .context("Failed to parse identifier text")?
-                    .to_lowercase();
-                if matches!(identifier_text.as_str(), "wait" | "flush") {
-                    Ok(StatType::IoStat)
-                } else {
-                    Err(anyhow!("Unknown routine: {identifier_text}"))
-                }
-            }
             _ => Err(anyhow!("Node does not have a stat type")),
         }
     }
@@ -128,7 +116,7 @@ impl AstRule for UncheckedStat {
         let stat_name: &'static str = stat_type.into();
 
         // Find a 'stat' argument in the allocate statement
-        let arg_list = if matches!(node.kind(), "subroutine_call" | "call_expression") {
+        let arg_list = if node.kind() == "subroutine_call" {
             node.child_with_name("argument_list")?
         } else {
             *node
@@ -220,7 +208,6 @@ impl AstRule for UncheckedStat {
             "inquire_statement",
             "file_position_statement",
             "subroutine_call", // for execute_command_line
-            "call_expression", // for wait and flush
         ]
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2065,7 +2065,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter-fortran"
 version = "0.5.1"
-source = "git+https://github.com/stadelmanma/tree-sitter-fortran.git?branch=more-extensions#92952edf3f825045b2196394891c67fea9586f9a"
+source = "git+https://github.com/stadelmanma/tree-sitter-fortran.git?rev=1b08f4c#1b08f4c65fb47c76d3532cd3f7432d03abb9ef69"
 dependencies = [
  "cc",
  "tree-sitter-language",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ strum_macros = { version = "0.27.1" }
 textwrap = { version = "0.16.0" }
 tree-sitter = "~0.25.0"
 tree-sitter-cli = "~0.25.0"
-tree-sitter-fortran = { git = "https://github.com/stadelmanma/tree-sitter-fortran.git", branch = "more-extensions" }
+tree-sitter-fortran = { git = "https://github.com/stadelmanma/tree-sitter-fortran.git", rev = "1b08f4c" }
 
 [profile.dev.package]
 insta.opt-level = 3


LiamPattinson:
Added some more complex checking to see if the allocate stat is checked after use. It needs some refinement:

- Currently, if it finds another allocate before the stat is checked, an error is raised. However, this might be okay:
```f90
allocate (x(...), stat=status_x)
allocate (y(...), stat=status_y)
if (status_x /= 0 .or. status_y /= 0) then
  ...
```
  The correct response would be to check that the same variable isn't reused in another stat.
- The way I'm iterating over the rest of the current function/program etc is awful. I might add a new iterator type to help with this.
- We should also check that the allocated variable isn't used before the stat is checked. That might be best added as another rule.

I'll squash these commits down to clean up the history when I'm done.

fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,94 @@
...
+///
+/// ```f90
+/// allocate (x(BIG_INT), stat=status)
+/// if (status /= 0) then
+///   call handle_error(status)
+/// end if
+/// ```
+///
+/// However, if the `stat` variable is not checked, the program will continue to
+/// run despite the allocation failure, which can lead to undefined behaviour.
```
ZedThree, line 30: Without wanting this to become a textbook, maybe we should mention `errmsg` here too?

fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,94 @@
...
+                    // Found the variable, so stop checking.
+                    // Assume the user is doing something sensible with it!
+                    return None;
+                }
+            }
+        }
+        some_vec!(Diagnostic::from_node(Self { name }, &stat_node))
+
+        // TODO Also check that anything allocated in the failed statement is not used
+        // until after the variable is checked.
```
ZedThree, line 88: This is almost a separate rule? `use-before-stat-check` or something. There's an argument for doing it in the same rule. Certainly it would be best to check them both at the same time -- another rule where the ruff style visitor would be useful!

LiamPattinson: Yes, I think it should be an extra rule.

fortitude/src/rules/correctness/error_handling.rs
LiamPattinson: What do you think we should say about it? Should we just highlight that it's probably a good idea to set both?

LiamPattinson: Do you think a rule to enforce always setting both would be a good idea?

ZedThree: Yes, makes sense as a separate rule

ZedThree:
I think the second example is perverse and handled by saying the docs that the rule checks for `stat` being handled in the same scope as the `allocate`. If you want to do something like the second example, don't use the rule or use an allow comment.

The first example is a little suspect too, it's a bit weird to unconditionally check the status, but maybe something like

```f90
if (extra_size) then
  allocate(x(N + 2), stat=status)
else
  allocate(x(N), stat=status)
end if

if (status /= 0) then
...
```

might be a reasonable situation.

It's worth checking the corpus to see what situations this currently flags up

ZedThree:
Hmm, looks fine to me:

```
7:2  - 7:41       allocate_statement 
7:2  - 7:10         "allocate"
7:10 - 7:11         "("
7:11 - 7:15         allocation: sized_allocation 
7:11 - 7:12           identifier `x`
7:12 - 7:15           size 
7:12 - 7:13             "("
7:13 - 7:14             identifier `N`
7:14 - 7:15             ")"
7:15 - 7:16         ","
7:17 - 7:28         keyword_argument 
7:17 - 7:21           name: identifier `stat`
7:21 - 7:22           "="
7:22 - 7:28           value: identifier `status`
7:28 - 7:29         ","
7:30 - 7:40         keyword_argument 
7:30 - 7:36           name: identifier `errmsg`
7:36 - 7:37           "="
7:37 - 7:40           value: identifier `msg`
7:40 - 7:41         ")"
```

LiamPattinson:
I think we're almost there, now generalised it to handle IO and command line functions. Still have these tasks to resolve:

- There's a TODO comment in the code. When checking if a `stat` is overwritten, it's just looking for instances of `stat=status`. This will false-positive on user-defined error handling functions, so it should specifically look for a built-in function that uses `stat`, `iostat`, etc.
- In the tests, for some reason the last `deallocate` in the `io` subroutine isn't triggering, and I'm unsure why.

LiamPattinson:
Updated to latest tree-sitter changes. I'm still having to treat `wait` and `flush` as `call_expression`, but everything else seems to work fine now.

fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,337 @@
...
+        let arg_list = if matches!(node.kind(), "subroutine_call" | "call_expression") {
+            node.child_with_name("argument_list")?
+        } else {
+            *node
+        };
+        let stat_node = arg_list.named_children(&mut node.walk()).find(|child| {
+            child.kind() == "keyword_argument"
+                && child.child_by_field_name("name").is_some_and(|n| {
+                    n.to_text(src).map(|s| s.to_lowercase()) == Some(stat_name.to_string())
+                })
```
ZedThree: Do we already have a helper function for keyword args? Might be nice to add one if not!

fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+fn new_branch(node: &Node) -> bool {
+    matches!(
+        node.kind(),
+        "elseif_clause" | "else_clause" | "case_statement"
+    )
+}
+
+fn is_error_checking_routine(node: &Node, src: &str) -> Result<bool> {
```
ZedThree: This looks very similar to `StatType::from_node` -- anyway to reuse that here?

ZedThree (approved):
Looks great, thanks @LiamPattinson!

fortitude/src/rules/correctness/error_handling.rs
LiamPattinson: We have `is_keyword_argument` in `ast.rs`, but I think we can generalise it a little and make it part of the `FortitudeNode` trait. I'll see what I can do.

ZedThree: Ok, we can do that in another PR though. I'm imagining we could have something like `node.get_keyword_arg("stat").value`

LiamPattinson: Whoops, just pushed a fix to https://github.com/PlasmaFAIR/fortitude/pull/458/commits/476d5c23fa90b6c4d77fc8838e22f00005b100b4! I went with a trio of functions:
- `node.kwarg()` -> the `keyword_argument` node
- `node.kwarg_value()` -> the `value` node of the kwarg
- `node.kwarg_exists()` -> utility that just calls `.is_some()` on the result of `kwarg`
I cleaned up a few other bits of the code with them too.

ZedThree: Nice, looks great!

fortitude/src/rules/correctness/error_handling.rs
ZedThree: @LiamPattinson Could you have a quick look at this? If it's not straightforward, feel free to go ahead and merge

LiamPattinson: I've fixed it up in https://github.com/PlasmaFAIR/fortitude/pull/458/commits/1d4e4e5224e0aca4cfbe0113a8338d2d6ff4545f. Also found and fixed a lot of case-sensitivity issues while I was at it.

LiamPattinson:
Is it worth merging this before the `wait`/`flush` updates to tree-sitter-fortran?

LiamPattinson:
I've pinned the `tree-sitter-fortran` version to the specific commit we need, and removed the logic to handle `call_expression` nodes -- everything seems to be working fine!

ZedThree (approved):

PR #459, opened 2025-05-27
Add dev command to print the AST

ZedThree:
This is just a convenience wrapper around `tree-sitter parse` to make it easier for new developers


LiamPattinson (approved):
Amazing, this will save me a lot of time spent jumping back and forth between `fortitude` and `tree-sitter-fortran`!

PR #460, opened 2025-05-27
Add incorrect-space-between-brackets rule

connoraird:


LiamPattinson:
Cheers @connoraird! Could this be expanded to do both open and close brackets in one rule?

PR #464, opened 2025-05-30
Add rule for space between statements (if, allocate, then, etc)

connoraird:
I have also extracted shared logic from S104 into function to get the whitespace range between open and close nodes.

### DIFF
--- a/fortitude/src/rules/style/whitespace.rs
+++ b/fortitude/src/rules/style/whitespace.rs
@@ -191,51 +191,14 @@ impl AstRule for IncorrectSpaceBetweenBrackets {
     fn check(_settings: &Settings, node: &Node, src: &SourceFile) -> Option<Vec<Diagnostic>> {
         let node_as_str = node.to_text(src.source_text())?;
 
-        let source = src.to_source_code();
-        let bracket_start = node.start_textsize();
-        let bracket_end = node.end_textsize();
-        let line_index = source.line_index(bracket_end);
-
         let is_open_bracket = matches!(node_as_str, "(" | "[");
-        let (whitespace_start, whitespace_end) = if is_open_bracket {
-            // Get line after bracket
-            let line_end = source.line_end(line_index);
-            let range_after = TextRange::new(bracket_end, line_end);
-            let line_after = source.slice(range_after);
-
-            // Ignore if preceding a line wrap, i.e. &
-            if line_after.trim_start().starts_with('&') {
-                return None;
-            }
-
-            // Count whitespace characters after the bracket
-            let whitespace_iter = line_after.chars().take_while(|c| c.is_whitespace());
-            let whitespace_count = whitespace_iter.count();
-            let whitespace_end = bracket_end + TextSize::from(whitespace_count as u32);
-
-            (bracket_end, whitespace_end)
-        } else {
-            // Get line before bracket
-            let line_start = source.line_start(line_index);
-            let range_before = TextRange::new(line_start, bracket_start);
-            let line_before = source.slice(range_before);
-
-            // Ignore if following a line wrap, i.e. &
-            if line_before.trim_end().ends_with('&') || line_before.trim().is_empty() {
-                return None;
-            }
-
-            // Count whitespace characters before the bracket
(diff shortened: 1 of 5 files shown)


ZedThree:
I think this rule might need some options -- a space after `if` is very common, but most people write `allocate`, `write`, etc without a space:

| Count | version |
|-|-|
| 141272 | `allocate(` |
|  39960 | `ALLOCATE(` |
|  22744 | `allocate (` |
|  15905 | `ALLOCATE (` |
|   1275 | `Allocate(` |
|     68 | `allocate  (` |
|     51 | `ALLOCATE  (` |
|     32 | `Allocate (` |
|     25 | `allocate   (` |
|      8 | `allocate    (` |
|      2 | `allocatE(` |
|      1 | `ALLOCATE   (` |
|      1 | `allocate         (` |


| Count | version |
|-|-|
| 248606| `write(` |
| 177731| `write (` |
| 111666| `WRITE(` |
|  30893| `WRITE (` |
|   1359| `Write(` |
|    340| `Write (` |
|     11| `WRITE  (` |
|      8| `write  (` |

| Count | version |
|-|-|
|1015817| `if (` |
| 725554| `IF (` |
| 279594| `if(` |
| 244944| `IF(` |
|   7650| `If (` |
|   2713| `if     (` |
|   1898| `If(` |
|    899| `IF  (` |
|    398| `if  (` |
|    383| `IF      (` |
|    264| `IF     (` |
|    143| `if      (` |
|     78| `IF    (` |
|     45| `if   (` |
|     39| `if    (` |
|     38| `IF   (` |
|      6| `iF (` |
|      4| `iF(` |
|      4| `if       (` |
|      3| `if        (` |
|      2| `IF       (` |
|      2| `IF        (` |
|      2| `IF           (` |
|      2| `If  (` |
|      1| `IF          (` |

(`read` and `deallocate` follow `write` and `allocate`)

ZedThree:
Yeah, I think this might trigger on variables called `read` as well, for instance

fortitude/src/rules/style/whitespace.rs
```diff
@@ -245,3 +208,186 @@ impl AstRule for IncorrectSpaceBetweenBrackets {
...
+                        .to_text(src.source_text())?
+                        .to_lowercase()
+                        .as_str(),
+                    "if"
+                ) {
+                    Some(start_node)
+                } else {
+                    None
+                }
+            } else if is_allocate_type_statement {
```
LiamPattinson, line 277: In situations like this with a lot of `if`/`else if` statements, I'd recommend instead using a [match](https://doc.rust-lang.org/rust-by-example/flow_control/match.html). As this rule may have to target many more node types going forward, I'm thinking a structure like this might be helpful:

```rust
let diagnostics = match node_kind {
    "if_statement" => check_if(settings, node, src),
    "elseif_clause" => check_elseif(settings, node, src),
    "allocate_statement" | "deallocate_statement" | "read_statement" | "write_statement"  => check_standard_statement(settings, node, src)
}?;
```

For node types that need special treatment, like `if_statement`, we can write a separate `check` function, but could probably group together similar statement types that can all be checked the same way (`deallocate_statement` will need the tree-sitter updates from https://github.com/PlasmaFAIR/fortitude/pull/465 to compatible in this way).

LiamPattinson (commented):
Sorry for the delay in reviewing this one, I've been very busy with other things all week. I'm a little worried about the scalability of the solution here, so I've left a comment below with some recommendations, but it may be that we should rethink the scope of the rule entirely.

As @ZedThree has shown, there doesn't seem to be much consistency regarding whether or not a space should be included in the various statement types (it looks like I'm in a small minority that thinks a space should always be included!). This suggests that we should maybe instead narrow the scope of the rule to just something like `incorrect-space-in-if-statement` which would ensure:
- space between `if` and `(`
- space between `)` and `then`
- space between `else if` and `(`
- _should not_ check for space between `else` and `if`, as this is handled by a different rule

That would reduce the complexity of this rule significantly, and we could then have separate rules like `incorrect-space-in-io-statement` to handle the other cases.

Issue #461, opened 2025-05-27
Rule to enforce single allocation per statement when using stat


imciner2:
At least in the Cray compiler, if an `allocate` statement fails and a `stat` is being returned, it will stop all allocations in the statement and return (https://cpe.ext.hpe.com/docs/24.07/guides/CCE/HPE_Cray_Fortran_Reference_Manual_18.0.0_S-3901.html#allocate-error-status). This means if the `allocate` statement is allocating multiple variables, the ones after the failure will also be unallocated, but it doesn't seem possible to know what has been unallocated unless you actually check each variable specifically. It might make sense to have a linting rule that enforces only allocating a single variable if the `allocate` statement is using the `stat` argument, because that will enforce knowing what variable failed allocation.

ZedThree:
Looks like this is the case for gfortran and intel too. `errmsg` doesn't provide any more information

LiamPattinson:
Thanks @imciner2, as you've seen I'm working on a similar rule right now, so we should be able to add this fairly easy.

PR #466, opened 2025-05-30
Add rule `multiple-allocations-with-stat`

LiamPattinson:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/461

Builds on https://github.com/PlasmaFAIR/fortitude/pull/458

Not yet ready to merge, as the rule has been written in anticipation of updates to the tree-sitter grammar, and currently is not catching deallocate statements.

### DIFF
--- a/fortitude/src/rules/correctness/error_handling.rs
+++ b/fortitude/src/rules/correctness/error_handling.rs
@@ -348,10 +348,10 @@ impl AstRule for MultipleAllocationsWithStat {
         let stat_node = node.kwarg("stat", src)?;
 
         // Count allocations
-        let count = match node.kind() {
-            "allocate_statement" => count_allocations(node),
-            "deallocate_statement" => count_deallocations(node),
-            _ => return None,
+        let count = if node.kind() == "allocate_statement" {
+            count_allocations(node)
+        } else {
+            count_deallocations(node)
         };
         if count <= 1 {
             return None;
--- a/fortitude/src/rules/correctness/error_handling.rs
+++ b/fortitude/src/rules/correctness/error_handling.rs
@@ -63,13 +63,13 @@ enum CheckStatus {
 }
 
 /// ## What does it do?
-/// This rule detects whether a `stat`, `iostat`, and `cmdstat` variable is checked
+/// This rule detects whether a `stat`, `iostat`, and `cmdstat` argument is checked
 /// within the same scope it is set.
 ///
 /// ## Why is this bad?
-/// By default, `allocate` statements will crash the program if the allocation
+/// By default, `allocate` statements will abort the program if the allocation
 /// fails. This is often the desired behaviour, but to provide for cases in
-/// which the user wants to handle allocation errors gracefully, they may
+/// which the developer wants to handle allocation errors gracefully, they may
 /// optionally check the status of an `allocate` statement by passing a variable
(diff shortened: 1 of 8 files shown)


fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+/// ## What does it do?
+/// This rule detects whether a `stat`, `iostat`, and `cmdstat` variable is checked
+/// within the same scope it is set.
+///
+/// ## Why is this bad?
+/// By default, `allocate` statements will crash the program if the allocation
+/// fails. This is often the desired behaviour, but to provide for cases in
+/// which the user wants to handle allocation errors gracefully, they may
```
ZedThree: Oh, this is very minor, but I meant to say in the previous PR, just some tweaks to the precision of the language:

```suggestion
/// This rule detects whether a `stat`, `iostat`, and `cmdstat` argument is checked
/// within the same scope it is set.
///
/// ## Why is this bad?
/// By default, `allocate` statements will abort the program if the allocation
/// fails. This is often the desired behaviour, but to provide for cases in
/// which the developer wants to handle allocation errors gracefully, they may
```

fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,379 @@
...
+/// deallocations.
+///
+/// ## Why is this bad?
+/// When allocating or deallocating multiple arrays at once, the use of a `stat`
+/// parameter will permit the program to continue running even if one of the
+/// allocations or deallocations fails. However, it may not be clear which
+/// allocation or deallocation caused the error.
+///
+/// To avoid confusion, it is recommended to use separate allocate or deallocate
+/// statements for each array and check the `stat` parameters individually.
```
ZedThree: Again, minor change for precision:

```suggestion
/// When allocating or deallocating multiple variables at once, the use of a `stat`
/// parameter will permit the program to continue running even if one of the
/// allocations or deallocations fails. However, it may not be clear which
/// allocation or deallocation caused the error.
///
/// To avoid confusion, it is recommended to use separate allocate or deallocate
/// statements for each variable and check the `stat` parameters individually.
```

`allocate` is also used for allocatable-length `character` variables, and for polymorphic variables, e.g.

```f90
class(*) :: var
allocate(integer::var)
```

fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -0,0 +1,379 @@
...
+
+        // Check this has a stat parameter
+        let stat_node = node.kwarg("stat", src)?;
+
+        // Count allocations
+        let count = match node.kind() {
+            "allocate_statement" => count_allocations(node),
+            "deallocate_statement" => count_deallocations(node),
+            _ => return None,
+        };
```
ZedThree: As we only trigger on the two nodes, we could just do:

```suggestion
        let count = if node.kind() == "allocate_statement" {
            count_allocations(node)
        } else {
            count_deallocations(node)
        };
```

LiamPattinson: Oh yeah, that's super obvious now I'm looking at it. I often find myself defaulting to match statements when I'm thinking in Rust.

ZedThree (approved):

PR #467, opened 2025-06-02
Add an option to exempt specific modules from rule C121

hansec:
Often I find it useful to have one or few more modules that are "safe" to `USE` without `ONLY` specification. For example a base runtime or environment set. This pull request enables exempting specific modules from [C121](https://fortitude.readthedocs.io/en/stable/rules/use-all/) by name.

### DIFF
--- a/docs/rules/use-all.md
+++ b/docs/rules/use-all.md
@@ -3,6 +3,8 @@ This rule is turned on by default.
 
 ## What it does
 Checks whether `use` statements are used correctly.
+Specific modules may be exempted from this rule by setting the
+[`allow-mods`](../settings.md#allow-mods) option.
 
 ## Why is this bad?
 When using a module, it is recommended to add an 'only' clause to specify which
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -730,3 +730,32 @@ Quote style to prefer for string literals (either "single" or "double").
 
 ---
 
+### `check.use-all`
+
+Options for the use all rule
+
+#### [`allow-mods`](#check_use-all_allow-mods) {: #check_use-all_allow-mods }
+<span id="allow-mods"></span>
+
+Allow use all for specific modules.
+
+**Default value**: `[]`
+
+**Type**: `List[str]`
+
+**Example usage**:
+
+=== "fpm.toml"
+
+    ```toml
+    [extra.fortitude.check.use-all]
+    allow-mods = ["my_env_mod"]
+    ```
+=== "fortitude.toml"
+
+    ```toml
+    [check.use-all]
+    allow-mods = ["my_env_mod"]
+    ```
+
+---
--- a/fortitude/src/options.rs
+++ b/fortitude/src/options.rs
@@ -11,7 +11,10 @@ use serde::{Deserialize, Serialize};
 use crate::{
     rule_selector::RuleSelector,
     rules::{
-        correctness::exit_labels,
+        correctness::{
+            exit_labels,
+            use_statements,
+        },
         portability::{self},
         style::{
             keywords,
@@ -296,6 +299,10 @@ pub struct CheckOptions {
     #[option_group]
     pub keyword_whitespace: Option<KeywordWhitespaceOptions>,
 
+    /// Options for the `use-all-allow-mods` rule
+    #[option_group]
+    pub use_all: Option<UseAllOptions>,
+
     /// Options for the `bad-string-quote` rule
     #[option_group]
     pub strings: Option<StringOptions>,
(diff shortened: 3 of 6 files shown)


docs/rules/use-all.md
```diff
@@ -3,6 +3,8 @@ This rule is turned on by default.
 
 ## What it does
 Checks whether `use` statements are used correctly.
+Specific modules may be exempted from this rule by setting the
+[`allow-mods`](../settings.md#allow-mods) option.
```
ZedThree, line 7: This file is actually automatically built (with `cargo dev generate-all`) from the docstring on the `UseAll` struct in `use_statements.rs`. Please could you put this text in there? Otherwise it will be clobbered next time the docs are generated :)

I don't think we've documented this, but if you put a section:

```markdown
## Options
- `check.use-all.allow-mods`
```

then `cargo dev generate-all` will take care of linking to the correct anchor on the settings page.

fortitude/src/options.rs
```diff
@@ -296,6 +299,10 @@ pub struct CheckOptions {
     #[option_group]
     pub keyword_whitespace: Option<KeywordWhitespaceOptions>,
 
+    /// Options for the `use-all-allow-mods` rule
```
ZedThree, line 302:
```suggestion
    /// Options for the `use-all` rule
```

fortitude/src/options.rs
```diff
@@ -368,6 +375,32 @@ impl KeywordWhitespaceOptions {
     }
 }
 
+/// Options for the `use-all-allow-mods` rules
```
ZedThree, line 378:
```suggestion
/// Options for the `use-all` rules
```

fortitude/src/options.rs
```diff
@@ -368,6 +375,32 @@ impl KeywordWhitespaceOptions {
...
+#[derive(
+    Clone, Debug, PartialEq, Eq, Default, OptionsMetadata, CombineOptions, Serialize, Deserialize,
+)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct UseAllOptions {
+    /// Whether to enforce the use of `in out` instead of `inout`.
```
ZedThree, line 384: This looks like a copy-paste leftover :)

fortitude/src/rules/correctness/use_statements.rs
```diff
@@ -39,8 +39,16 @@ impl Violation for UseAll {
...
 impl AstRule for UseAll {
     fn check(_settings: &Settings, node: &Node, _src: &SourceFile) -> Option<Vec<Diagnostic>> {
-        if node.child_with_name("included_items").is_none() {
-            return some_vec![Diagnostic::from_node(UseAll {}, node)];
+        let module_name = node
+            .child_with_name("module_name")?
+            .to_text(_src.source_text())?
+            .to_lowercase();
+
+        if _settings.check.use_all.allow_mods.iter().all(|m| *m != module_name)
```
ZedThree, line 47: I _think_ we should be able to do:

```suggestion
        if !_settings.check.use_all.allow_mods.contains(module_name)
```

Also, now that we are using the `_settings` variable, please could you rename it to `settings`?

Lastly, if you run `cargo clippy`, it gives a suggestion for collapsing the nested `if` statements into one

fortitude/src/options.rs
```diff
@@ -368,6 +375,32 @@ impl KeywordWhitespaceOptions {
...
+    #[option(
+        default = "[]",
+        value_type = "list[str]",
+        example = r#"
+            # In addition to the standard set of exclusions, omit all tests, plus a specific file.
+            extend-exclude = ["tests", "src/bad.f90"]
```
ZedThree, line 390: Another copy-paste leftover :)

fortitude/src/rules/correctness/use_statements.rs
LiamPattinson, line 47: A very minor request: please could the underscored variable names have their underscores removed? We're using the naming convention that `_variable` is unused.

```suggestion
            .to_text(src.source_text())?
            .to_lowercase();

        if settings.check.use_all.allow_mods.iter().all(|m| *m != module_name)
```

fortitude/src/options.rs
```diff
@@ -368,6 +375,32 @@ impl KeywordWhitespaceOptions {
...
+        "#
+    )]
+    pub allow_mods: Option<Vec<String>>
+}
+
+impl UseAllOptions {
+    pub fn into_settings(self) -> use_statements::settings::Settings {
+        use_statements::settings::Settings {
+            allow_mods: self.allow_mods.unwrap_or_default()
```
ZedThree, line 399: We probably want to ensure `allow_mods` is always all lowercase:

```suggestion
            allow_mods: self.allow_mods.unwrap_or_default()
                .iter()
                .map(|m| m.to_lowercase())
                .collect()
```

fortitude/src/rules/correctness/use_statements.rs
```diff
@@ -126,3 +134,28 @@ impl AstRule for MissingIntrinsic {
...
+        fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+            if self.allow_mods.is_empty() {
+                write!(f, "{{}}")?;
+            } else {
+                writeln!(f, "{{")?;
+                for allow_mods in &self.allow_mods {
+                    writeln!(f, "\t{allow_mods}")?;
+                }
+                write!(f, "}}")?;
+            }
```
ZedThree, line 157: There's a macro we've stolen from ruff for this: `display_settings!` (that we also haven't documented, sorry!)

```suggestion
           display_settings! {
                formatter = f,
                namespace = "check.use-all",
                fields = [self.allow_mods | array]
            }
```

docs/settings.md
```diff
@@ -730,3 +730,32 @@ Quote style to prefer for string literals (either "single" or "double").
 
 ---
 
+### `check.use-all`
```
ZedThree, line 733: This is another auto-generated file btw, so any manual changes will be lost

ZedThree (changes requested):
Thanks @hansec! I have some a few comments and things that could be changed.

I think we should also emphasise that this option should be used sparingly, and `! allow(use-all)` can be used for individual instances.

PR #468, opened 2025-06-10
Add rule `stat-without-message`

LiamPattinson:
Builds on https://github.com/PlasmaFAIR/fortitude/pull/466

Ensures that an error message is always captured alongside `stat`.

### DIFF
--- a/fortitude/src/rules/correctness/error_handling.rs
+++ b/fortitude/src/rules/correctness/error_handling.rs
@@ -348,10 +348,10 @@ impl AstRule for MultipleAllocationsWithStat {
         let stat_node = node.kwarg("stat", src)?;
 
         // Count allocations
-        let count = match node.kind() {
-            "allocate_statement" => count_allocations(node),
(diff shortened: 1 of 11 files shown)


fortitude/src/rules/correctness/error_handling.rs
```diff
@@ -297,3 +305,152 @@ fn new_branch(node: &Node) -> bool {
...
+///
+/// ```f90
+/// real, allocatable :: x(:)
+/// integer :: status
+/// character(256) :: message
+///
+/// allocate (x(100), stat=status, errmsg=message)
+/// open (unit=10, file="data.txt", iostat=status, iomsg=message)
+/// call execute_command_line("ls", cmdstat=status, cmdmsg=message)
+/// ```
```
ZedThree, line 409: Might be worth mentioning that in F2023, these functions can automatically allocate `errmsg` to be the correct size (although I don't think anybody actually implements this yet!)

ZedThree (approved):

Issue #271, opened 2025-01-10
Support multiple locations/ranges in diagnostics


ZedThree:
There are several cases where it would be either really helpful or even necessary to support being able to annotate multiple ranges or locations in diagnostics. For example, the original location of shadowed variables, the first module defined in a file, the originating macro for a warning in a preprocessed file.

I believe the `annotate-snippets` package supports this, although we'd need to overhaul the diagnostic infrastructure to support extra ranges. Maybe something like:

```rust
pub struct Note {
    pub range: TextRange,
    pub message: String,
}

pub struct Diagnostic {
    pub kind: DiagnosticKind,
    pub range: TextRange,
    pub fix: Option<Fix>,
    pub parent: Option<TextSize>,
    pub notes: Vec<Note>,
}

Diagnostic::from_node(MyViolation{}, node)
  .with_note(Note { range, message }
  .with_note(Note { second_range, second_message };
```

imciner2:
An extension of this that I just thought of today is being able to have the multiple locations possibly in different files. That would be useful in cases where something is defined in a different module and you want to actually refer to its declaration.

An example of this I just thought of today is this: Eventually when we have a better AST representation we can keep track of constants and their values, and then inside some rules we can reference those values directly and show them in the diagnostics. The first rule I was thinking is actually for detecting when a do loop won't perform any iterations due to the bounds being incorrect - and that is most useful when the bounds come from variables, e.g.
```
do j = arr_start, arr_stop, stride

end do
```
and it would be nice if the diagnostic could reference the exact location where these constants come from.

ZedThree:
Ruff has done a big refactoring of its `Diagnostics` and how they're rendered (see https://github.com/astral-sh/ruff/pull/19415 for instance), and it looks like it now supports multiple annotations

ZedThree:
Done in #668

PR #471, opened 2025-06-13
Add rule `maths-missing-parentheses`

### LINKED ISSUE #469: Check for missing parentheses in mathematical expressions that involve operators of different priorities
The clang-tidy rule [math-missing-parentheses](https://clang.llvm.org/extra/clang-tidy/checks/readability/math-missing-parentheses.html) is interesting and we can probably implement it fairly easily, I think it's just check if two `math_expression`s are siblings and have different precedences.

Example from the clang-tidy check:

Before:

```cpp
int x = 1 + 2 * 3 - 4 / 5;
```
After:

```cpp
int x = 1 + (2 * 3) - (4 / 5);
```

ZedThree:
Closes #469 

Needs a bump to `tree-sitter-fortran` because `unary_expression` had the wrong precedence.

This doesn't catch `1. * 2. / 3. * 4.` because `*` and `/` have the same precedence (and so can be simply read left-to-right), but I do think this sort of construction is a little confusing. Any suggestions on what would be sensible to flag here (presumably with a different message too)? 

Maybe just `*` with a parent of `/`? So:

```f90
1 / 2 * 3  ! warn
1 * 2 / 3  ! ok
```

### DIFF
--- a/crates/fortitude_linter/src/rules/style/maths_missing_parentheses.rs
+++ b/crates/fortitude_linter/src/rules/style/maths_missing_parentheses.rs
@@ -1,0 +1,136 @@
+use crate::ast::FortitudeNode;
+use crate::settings::Settings;
+use crate::{AstRule, FromAstNode};
+use ruff_diagnostics::{AlwaysFixableViolation, Diagnostic, Fix};
+use ruff_macros::{derive_message_formats, ViolationMetadata};
+use ruff_source_file::SourceFile;
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks if mathematical expressions are missing parentheses when operators have
+/// different precedences.
+///
+/// ## Why is this bad?
+/// Long or complex expressions can be difficult or confusing to read, especially when
+/// mixing operators with different precedences. Adding parentheses can clarify the code
+/// and make the author's intent clearer, reducing the likelihood of misunderstandings
+/// or bugs.
+///
+/// ## Example
+/// ```f90
+/// x = 1. + 2. * 3. - 4. / 5.
+/// ```
+///
+/// Use instead:
+/// ```f90
+/// x = 1. + (2. * 3.) - (4. / 5.)
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct MathsMissingParentheses {
+    op1: String,
+    op2: String,
+}
+
+impl AlwaysFixableViolation for MathsMissingParentheses {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let Self { op1, op2 } = &self;
+        format!("'{op1}' has higher precedence than '{op2}'; add parentheses for clarity")
+    }
+
+    fn fix_title(&self) -> String {
+        "Add parentheses".to_string()
+    }
+}
+
+#[derive(PartialEq)]
+enum Operator {
+    Exponential,
+    Multiplication,
+    Division,
+    Addition,
+    Subtraction,
+}
+
+impl Operator {
+    fn precedence(&self) -> u8 {
+        use Operator::*;
+        match self {
+            Exponential => 4,
+            Multiplication | Division => 3,
+            Addition | Subtraction => 2,
+        }
+    }
+}
+
+impl TryFrom<&str> for Operator {
+    type Error = &'static str;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        use Operator::*;
+        match value {
+            "**" => Ok(Exponential),
+            "*" => Ok(Multiplication),
+            "/" => Ok(Division),
+            "-" => Ok(Addition),
+            "+" => Ok(Subtraction),
+            _ => Err("not a maths operator"),
+        }
+    }
+}
+
+impl From<Operator> for String {
+    fn from(value: Operator) -> String {
+        use Operator::*;
+        match value {
+            Exponential => "**",
+            Multiplication => "*",
+            Division => "/",
+            Addition => "-",
+            Subtraction => "+",
+        }
+        .to_string()
+    }
+}
+
+impl AstRule for MathsMissingParentheses {
+    fn check<'a>(
+        _settings: &Settings,
+        node: &'a Node,
+        src: &'a SourceFile,
+    ) -> Option<Vec<Diagnostic>> {
+        let text = src.source_text();
+        let op: Operator = node.child(1)?.to_text(text)?.try_into().ok()?;
+
+        let parent = node.parent()?;
+        if !matches!(parent.kind(), "math_expression" | "unary_expression") {
+            return None;
+        }
+
+        let edit = node.edit_replacement(src, format!("({})", node.to_text(text)?));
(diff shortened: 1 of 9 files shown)


LiamPattinson (approved):
Thanks @ZedThree, I've tested it myself and it seems to be behaving sensibly to me.

I'm not sure how to include `a * b / c * d` in this one. Perhaps that could be a different rule, `mixed-multiply-divide` or something?

ZedThree:
It was pretty easy to add that as a check, and it looks sensible:

```
./resources/test/fixtures/style/S251.f90:5:11: S251 [*] '*' has higher precedence than '/'; add parentheses for clarity
  |
3 |   print*, 1. + 2. * 3. - 4. / 5.;
4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
5 |   print*, 1. * 2. / 3. * 4.;
  |           ^^^^^^^^^^^^^^^^^ S251
6 |   print*, -1.**2
  |
  = help: Add parentheses

ℹ Safe fix
2 2 |   implicit none (type, external)
3 3 |   print*, 1. + 2. * 3. - 4. / 5.;
4 4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
5   |-  print*, 1. * 2. / 3. * 4.;
  5 |+  print*, (1. * 2. / 3.) * 4.;
6 6 |   print*, -1.**2
7 7 | 
8 8 |   ! These are all fine
```

although perhaps the highlighted range could be adjusted.

I do think it's a sensible check, I'm in two minds whether it should be in this rule (`maths-missing-parentheses` isn't a terrible name for this), or in its own one (`potentially-confusing-multiplication`? `division-missing-parentheses`?)

---

Related, is there a nicer name for this than `maths-missing-parentheses`?

crates/fortitude_linter/src/rules/style/maths_missing_parentheses.rs
```diff
@@ -0,0 +1,136 @@
...
+/// ```f90
+/// x = 1. + (2. * 3.) - (4. / 5.)
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct MathsMissingParentheses {
+    op1: String,
+    op2: String,
+}
+
+impl AlwaysFixableViolation for MathsMissingParentheses {
```
imciner2, line 34: Is this `AlwaysFixable`? I would actually consider something like this more of an unsafe edit because it is something where you would want to verify the code and simply adding the parenthesis to make the Fortran order of operations clear could even be incorrect.

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__maths-missing-parentheses_S251.f90.snap
```diff
@@ -0,0 +1,128 @@
...
+4 4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
+5 5 |   print*, 1. * 2. / 3. * 4.;
+6 6 |   print*, -1.**2
+
+./resources/test/fixtures/style/S251.f90:4:30: S251 [*] '*' has higher precedence than '+'; add parentheses for clarity
+  |
+2 |   implicit none (type, external)
+3 |   print*, 1. + 2. * 3. - 4. / 5.;
+4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
+  |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^ S251
```
imciner2, line 51: I think it would be better if the rule would annotate the error including both the problematic operators underscored. e.g., in this case I think
```
4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
  |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ S251
```
is easier to parse mentally and locate the problem - because I probably spent a good few minutes trying to parse what this line was saying and where that `+` operator actually was at first I thought it was inside a set of parenthesis).

I guess even better would be if we had https://github.com/PlasmaFAIR/fortitude/issues/271 actually, because then we can highlight each operation separately (which is potentially better if the expressions span multiple lines in a complicated equation).

crates/fortitude_linter/src/rules/style/maths_missing_parentheses.rs
LiamPattinson: I guess it depends on whether a 'safe fix' should aim to preserve the the 'code-as-written' or the 'code-as-intended'. We've taken the opposite interpretations in `misleading-inline-if-statement`, which keeps the 'as-written' behaviour, and `split-escaped-quote`, which switches the user's code to the 'as-intended' behaviour. Personally, I think the latter should be an unsafe fix, as safe fixes shouldn't change the behaviour of any code, even if we're confident it was written in error.

In this case, I think it should be fine to have an auto-fix as long as it preserves the 'as-written' behaviour.

Do you have an example of how the fix would be incorrect? I can't immediately see a case where the addition of parentheses by this rule would change the result of an expression.

imciner2: It isn't that it will change the behavior of the code before/after the fix, but I guess I generally think that a safe fix that can always be applied should be something that the programmer doesn't need to look at the occurrence and verify correctness for, and possible correctness problems (which could be incorrect order of operations) should be examined individually to ensure they are actually safely fixed by the change (similar to how C122 is an unsafe fix).

ZedThree: That's a reasonable interpretation, especially if we're highlighting that something is bugprone. `clang-tidy` has this rule under `readability`, which is why I have it under `style` here. I did originally go for `correctness`, and I think there's a good argument for that too.

I'm leaning towards making this an unsafe fix -- we _can_ fix it automatically, but users should probably double-check their intention beforehand.

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__maths-missing-parentheses_S251.f90.snap
ZedThree: I agree some of these are quite difficult to read! Especially as they tend to already be in complex expressions already.

Maybe if we just highlighted from one operator to the other?

```
4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
  |                            ^^^^^^^^^^^^^^^^^^^^^ S251
```

Do we think that's any clearer?

Yeah, #271 would be very helpful, we could do something like:
```
4 |   print*, 1.*(cos(2.) - 3. + (4.*5.-6.*sin(7.))*sin(8.))
  |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ S251
                               ~ '+'               ~ '*'
```

Issue #168, opened 2024-11-29
Obsolescent features


ZedThree:
The standard has formally obsolescent and deleted features:

[ ] Alternate return — see B.3.2.
[ ] Computed GO TO — see B.3.3.
[x] Statement functions — see B.3.4.
[ ]DATA statements amongst executable statements — see B.3.5.
[ ] Assumed length character functions — see B.3.6.
[ ] Fixed form source — see B.3.7.
[x] CHARACTER* form of CHARACTER declaration — see B.3.8.
[ ] ENTRY statements — see B.3.9.
[ ] Label form of DO statement – see B.3.10.
[ ] COMMON and...
[ ] ...EQUIVALENCE statements, and...
[ ] ...the block data program unit – see B.3.11.
[ ] Specific names for intrinsic functions – see B.3.12.
[ ] FORALL construct and statement – see B.3.13
           
This implies we should move the `character*` to `Obsolescent`

imciner2:
Specific names for intrinsic functions check proposed in https://github.com/PlasmaFAIR/fortitude/pull/254.

ZedThree:
Computed goto done in #264

imciner2:
>  DATA statements amongst executable statements — see B.3.5.

I think we actually should go further than this and mark any `DATA` statement as obsolete. Even in non-executable regions, it can be not easy to read the statement, and could lead to errors if people add/remove things from the statement.

PR #478, opened 2025-07-01
Rules to detect deprecated and discouraged OpenMP and MPI uses

imciner2:
This adds three new rules:
* Obsolence: deprecated `omp_lib.h` include
* Obsolence: deprecated `mpif.h` include
* Modernization: discouraged `mpi` module

The `omp_lib.h` include was technically only deprecated in the most recent version (6.0) released last June, however the standard does make reference to the module quite earlier than that (it is on the OpenMP 3.0 summary card from 2009). So I think it is fine to have that flagged.

The `mpif.h` include was deprecated as of MPI 4.1 (Nov 2023), but it's use has been strongly discouraged starting in MPI 3.0 (September 2012).

The `mpi` module is still technically supported, but its use is discouraged because it isn't as type-safe, so I put this under the modernization category and is off by default.

I have also updated the dev readme to reflect recent changes in the docs building and crate organization.


imciner2:
Hmm, there are still a few deprecated Fortran features to be implemented (https://github.com/PlasmaFAIR/fortitude/issues/168). Is it worth keeping all the Fortran deprecations close in number and shifting these up by a few hundred?

LiamPattinson:
> I have also updated the dev readme to reflect recent changes in the docs building and crate organization.

Thanks for that! I hadn't noticed that I'd missed that during the upgrade.

> Is it worth keeping all the Fortran deprecations close in number and shifting these up by a few hundred?

It might be worth having them at `201` and `211`, as then users could activate all parallelism rules together using the rule code `OB2`, but I don't think it's particularly important.

README.dev.md
```diff
@@ -229,7 +229,8 @@ Once you've completed the code for the rule itself, you can define tests with th
 The documentation can be built locally using:
 
 ```bash
-pip install mkdocs-material
+pip install mkdocs-material mkdocs-include-markdown-plugin
```
LiamPattinson: I recently edited this line in https://github.com/PlasmaFAIR/fortitude/pull/476, hence the merge conflict. I like the addition of `cargo dev generate-all`, but I'd prefer to keep `pip install -r requirements.doc.txt`.

crates/fortitude_linter/src/rules/obsolescent/mpi.rs
```diff
@@ -0,0 +1,123 @@
...
+        "mpif.h include is deprecated, use mpi or mpi_f08 module instead".to_string()
+    }
+}
+
+impl AstRule for DeprecatedMPIInclude {
+    fn check(_settings: &Settings, node: &Node, _src: &SourceFile) -> Option<Vec<Diagnostic>> {
+        let include_file = node
+            .child_with_name("filename")?
+            .to_text(_src.source_text())?
+            .to_lowercase();
```
LiamPattinson, line 113: I think `to_lowercase()` should be removed here.

crates/fortitude_linter/src/rules/obsolescent/mpi.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        if include_file.contains("mpif.h") {
```
LiamPattinson: It's pretty unlikely, but this could result in a false positive. The following will check that the filename matches exactly:

```suggestion
        // Strip quotes from the include file name
        let include_file = include_file.trim_matches('"').trim_matches('\'');

        if Path::new(&include_file).file_name() == Some("mpif.h".as_ref()) {
```

You'll have to use `std::path::Path` at the top of the file. The weirdness with `.as_ref()` on the right-hand-side is because the left returns an `OsStr` type.

crates/fortitude_linter/src/rules/obsolescent/openmp.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl AstRule for DeprecatedOmpInclude {
+    fn check(_settings: &Settings, node: &Node, _src: &SourceFile) -> Option<Vec<Diagnostic>> {
+        let include_file = node
+            .child_with_name("filename")?
+            .to_text(_src.source_text())?
+            .to_lowercase();
+
+        if include_file.contains("omp_lib.h") {
```
LiamPattinson: See the comment I raised above for the other include rule

crates/fortitude_linter/src/rules/modernisation/mpi.rs
```diff
@@ -0,0 +1,107 @@
...
+/// The original `mpi` module is inconsistent with the Fortran standard, and
+/// is not recommended by the MPI standard. Instead, the `mpi_f08` module is
+/// recommended because it is consistent with the Fortran standard (starting
+/// with Fortran 2008).
+///
+/// The older `mpi` module and `mpif.h` include used integers for named
+/// constants and the various MPI objects, which could lead to passing the wrong
+/// constants into arguments (e.g., passing a communicator into an operation)
+/// becausethis was not caught by the compiler. The new `mpi_f08` module uses
+/// derived types, allowing compile-time catching of these errors.
```
LiamPattinson: There's a typo `'becausethis'` here, but I think we could also shorten the paragraph a little:

```suggestion
/// The older `mpi` module and `mpif.h` include used integers for named
/// constants and the various MPI objects, which could lead to passing the wrong
/// constants into arguments (e.g., passing a communicator into an operation).
/// The new `mpi_f08` module uses derived types, allowing these errors to be
/// caught at compile-time.
```

LiamPattinson (commented):
Cheers for this, great to see the start of OpenMP and MPI linting! I'd love to see a few more OpenMP ones in particular (I'd personally like something to catch a missing `default(none)`).

Pretty much everything looks good here, I only have a few minor suggestions.

crates/fortitude_linter/src/rules/obsolescent/mpi.rs
imciner2: I included it here because some file systems are case-insensitive, and so then that could propagate up to the compiler level with the include statement being case-insensitive as well. So while technically it should be all lowercase, I felt this was just covering the bases.

imciner2:
Major comments addressed, but the Windows CI build seems to have failed downloading a package and I can't retrigger it. @LiamPattinson can you retrigger that workflow?

> It might be worth having them at 201 and 211, as then users could activate all parallelism rules together using the rule code OB2, but I don't think it's particularly important.

I've renumbered all three rules here to be in the 200s, so that can be the "parallel" region for these rule sets.


I am also seeing some Clippy errors now, in code that isn't touched in this PR: https://github.com/imciner2/fortitude/actions/runs/16219682004/job/45796992332.

The errors are all the same type, here is one of them:
```
error: variables can be used directly in the `format!` string
   --> crates/fortitude_linter/src/allow_comments.rs:212:37
    |
212 |             .edit_replacement(file, format!("! allow({})", remaining_codes))
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
    = note: `-D clippy::uninlined-format-args` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::uninlined_format_args)]`
help: change this to
    |
212 -             .edit_replacement(file, format!("! allow({})", remaining_codes))
212 +             .edit_replacement(file, format!("! allow({remaining_codes})"))
    |
```

crates/fortitude_linter/src/rules/obsolescent/mpi.rs
LiamPattinson: Ah, that's fair, I hadn't thought about that. I think it's pretty unlikely that anybody will have two separate similarly-named files -- the system-level file `mpi.f`, and a local user-defined file `MPI.f` -- so I'm happy to leave this for now.

LiamPattinson:
Thanks @imciner2, I'm happy with this PR. I've also raised a quick patch to the problems you raised (https://github.com/PlasmaFAIR/fortitude/pull/480), so once that's merged I think we'll be good to go unless @ZedThree has any further comments.

LiamPattinson (approved):
:tada: All looks good to me!

imciner2:
Gentle ping if this is good to be merged?

PR #484, opened 2025-08-06
Move CLI arguments to top-level crate

### LINKED ISSUE #470: Refactor CLI args and workspace configuration
Here's how ruff splits up their CLI args and configuration handling:

- `CheckCommand` has `.partition() -> (CheckArguments, ConfigArguments)`
  - `CheckArguments` is CLI only stuff and is used directly in `check`
  - `ConfigArguments` has an `ExplicitConfigOverrides` which are the CLI flags that override those in `Configuration`
  - `ConfigArguments` also has `overrides: Configuration` which allows arbitrary overrides via inline TOML on the CLI
- `ConfigArguments` (and `ExplicitConfigOverrides`) implement the `ConfigurationTransformer` trait which allows it to be turned into `Configuration`
- `Configuration` has `into_settings() -> Settings`, and `Settings` is what is actually used in `check`

This lets all the CLI stuff live in the front-end crate, and all the configuration stuff live in the workspace crate, with the cost that we need the following structs with (almost) duplicate info:

1. `CheckCommand` -- needed for CLI
2. `ExplicitConfigOverrides` -- bridges CLI and config files from CLI side
3. `Configuration` -- bridges CLI and config files from file side
4. `Settings` -- concrete settings

1, 2, & 3 have everything as an `Option` (or possibly empty `Vec`), `Settings` has no `Option`s

We have `Configuration::into_settings` taking `CheckArgs` as an argument, when to make this work like ruff, we need `impl ConfigurationTransformer for CheckArgs`

_Originally posted by @ZedThree in https://github.com/PlasmaFAIR/fortitude/pull/454#discussion_r2109400288_

ZedThree:
Closes #470 

This definitely feels like a little overkill at the moment for what it does, but it will be useful for the LSP integration, being able to combine config from the editor with the config file, and if we ever want hierarchical/inherited config files, this is essential machinery. And although we end up with some semi-duplicated structs, actually the separation of responsibility is much clearer.

I should probably put this in the developer docs, but here's how all the classes are related:

```mermaid
flowchart TD
  B{CLI} --> CheckCommand
  CheckCommand --> |partition| CheckArguments
  CheckCommand --> |partition| ExplicitConfigOverrides
  ExplicitConfigOverrides --> |transform| Configuration
  A{Config file} --> Options
  Options --> |from_options| Configuration
  Configuration --> |into_settings| Settings
```

- `CheckCommand::partition` takes care of merging things like `--fix` and `--no-fix`, and splitting up into two structs, `CheckArguments` which is CLI-specific stuff, and `ExplicitConfigOverrides` which are explicit CLI arguments that override config file options
  - This is also where we'd handle other overrides provided through `--config`, that we don't support yet but ruff does
- `Configuration::from_options` makes a new instance from the config file `Options` (this is unchanged in this PR)
- `transform` makes sure `ExplicitConfigOverrides` takes priority over the base `Configuration`, whether that's the default or from file
- Lastly, `Configuration::into_settings` gives us a `Settings` where everything has been concretised (that is, nothing is an `Option` so we don't have to handle anything being `None` after this)
  - This PR actually makes this simpler, because previously this also handled the CLI overrides


crates/fortitude_workspace/src/configuration.rs
```diff
@@ -221,22 +191,16 @@ impl Configuration {
...
-            fix_only: check.fix_only.unwrap_or_default(),
-            show_fixes: check.show_fixes.unwrap_or_default(),
-            unsafe_fixes: check
-                .unsafe_fixes
-                .map(UnsafeFixes::from)
-                .unwrap_or_default(),
-            output_format: check.output_format.unwrap_or_default(),
-            progress_bar: check.progress_bar.unwrap_or_default(),
-            preview: check.preview.map(PreviewMode::from).unwrap_or_default(),
+            extend_per_file_ignores: vec![],
```
LiamPattinson, line 194: How come this is just the empty vec? Is this a feature we're not offering in the config file yet?

LiamPattinson (approved):
Thanks for this, it was the one thing I couldn't quite figure out for https://github.com/PlasmaFAIR/fortitude/pull/454.

crates/fortitude_workspace/src/configuration.rs
ZedThree: Yes, exactly, currently it's a command line only option, and we don't have hierarchical config files (yet?) -- the struct member is a `Vec` not an `Option<Vec>`, which is why it's not just `None`

Issue #3, opened 2024-01-12
 Formatting mode


LiamPattinson:
Alongside a linting mode,  we could also provide a formatting mode. This could work by walking through the `tree_sitter::Tree` and rewriting the code with fixed whitespace rules. This would hopefully be more robust than a purely regex-based formatter, as it would be easier to distinguish between formatting rules in different contexts:

```f90
! Assignment operator should have space either side
myvar = x + y
! ... but not for keyword arguments
character(len=20, kind=c_char) :: string

! Colon should have a space after it
use mymodule, only: myfunc
! ... except when slicing
array(1:6) = 1
```

The following snippet can walk through the code via `tree_sitter` and print each token with spaces between. Some nodes appear to print twice, particularly those composed of multiple words (`double precision` goes to `doubleprecision doubleprecision`, `end module` goes to `endmodule endmodule`), but otherwise it reproduces simple codes well enough:

```rust
fn walk_and_print(node: &tree_sitter::Node, src: &str) {
    let mut cursor = node.walk();
    let mut children = node.children(&mut cursor);
    if children.len() > 0 {
        for child in children {
            walk_and_print(&child, &src);
        }
    } else {
        if node.is_named() {
            print!("{} ", node.utf8_text(src.as_bytes()).unwrap());
        } else {
            print!("{} ", node.kind());
        }
    }
}
```

To turn this into a formatter, it should be sufficient to implement a collection of functions for each node kind, and to call these via `match` instead of the simple recursion above. These could track indentation level as they traverse the tree, and different contexts could implement different rules for handling 'anonymous nodes' (mostly punctuation tokens).

Possible complications that are immediately apparent (I'm sure there will be more that I can't even imagine now):

- Maximum line length wrapping could be tricky, and there are multiple ways it could be achieved.
- Fortran has a lot of strange legacy options. It'll be no easy task to track down and understand all of them.
- [Edit] Will need to preserve newlines that the code author inserted to improve readability, but limit this to one newline.

ZedThree:
I've got a proof-of-concept/WIP topiary config: https://github.com/PlasmaFAIR/topiary/tree/fortran

Some issues that would need addressing:

- tree-sitter doesn't see `end if` etc as separate tokens/nodes, so topiary is unable to whitespace them
- topiary won't handle lower/uppercasing keywords
- I don't think topiary handles long-lines (although it can do multi-line stuff), so we'd need to handle that ourselves

LiamPattinson:
[...]

PR #486, opened 2025-08-09
Introduce line breaking in format command

sorenhartmann:
WIP


crates/fortitude_formatter/src/lib.rs
```diff
@@ -23,31 +31,342 @@ pub fn create_formatter() -> Language {
...
     println!("formatting {file:?}");
-    let input = File::open(file)?;
 
-    let mut buf_input = BufReader::new(input);
+    let mut lines: Vec<String> = get_uncontinued_lines(&file)?;
+
+    let format_input = lines.join("\n");
+    let mut format_output = vec![];
+
+    for _ in 0..10 {
```
ZedThree: Is this something like `MAX_TRIES`?

crates/fortitude_formatter/src/lib.rs
```diff
@@ -23,31 +31,342 @@ pub fn create_formatter() -> Language {
...
+            strings.push(stripped);
+        } else {
+            strings.push(
+                stripped
+                    .trim_start_matches(|c: char| c.is_whitespace() || c == '&')
+                    .to_string(),
+            )
+        }
+
+        if strings.last().unwrap().ends_with('&') {
```
ZedThree, line 117: This doesn't handle comments after a `&`, but maybe that's correct. Presumably if we see something like:

```f90
foo = bar & ! comment
  + zag
```
we only want to indent `+ zag` correctly, but leave the line break and comment alone, otherwise we have to work out where to put the comment.

crates/fortitude_formatter/src/lib.rs
```diff
@@ -23,31 +31,342 @@ pub fn create_formatter() -> Language {
...
-        &mut buf_input,
-        output,
-        language,
-        // TODO: user args?
-        Operation::Format {
-            skip_idempotence: true,
-            tolerate_parsing_errors: true,
-        },
-    )?;
+    println!("{}", lines.join("\n"));
```
ZedThree: This wants to be:

```suggestion
    writeln!(output, "{}", lines.join("\n"))?;
```

Although currently we do just write to `stdout`, for the tests we collect the output into a `String` so this accidentally causes the tests to fail!

(We'll probably actually want to just return this, but we'll need to wrap up the error types in our own enum -- we'll handle that separately)

fortitude/resources/test/fixtures/format/long-lines.f90
ZedThree, line 1: It looks like this file was left over from the rebase, so it can go?

crates/fortitude_formatter/resources/test/fixtures/format/long-lines.f90
ZedThree, line 1: You'll want to add this as a new test case in in `fortitude_formatter/src/lib.rs` around line 410, just add:
```rust
#[test_case(Path::new("long-lines.f90"))]
```

crates/fortitude_formatter/src/lib.rs
```diff
@@ -23,31 +31,342 @@ pub fn create_formatter() -> Language {
...
 
     Ok(())
 }
 
+fn get_uncontinued_lines(file: &PathBuf) -> Result<Vec<String>, FormatterError> {
```
ZedThree: My first thought was that this was just getting lines without continuations, but it's more like joining continued lines; is that right?

ZedThree (changes requested):
Thanks @sorenhartmann! This looks great, it's really appreciated!

I have some high level comments:

- it would be good to get some unit tests on some of the individual functions -- I'm happy to lend a hand with this
- after fixing writing the output, I notice there are some test failures: it looks like a line break is being forced in cases where one isn't necessary? For example, this snippet:
```diff
-  function real_not(a)
+  function real_not( &
+    a)
```
Ahh, this is because the line length is currently hardcoded to `20`, with an indent of `2`; changing this to `80` and all the tests pass -- ok, so this is just a matter of needing to handle user options. Again, something I'm happy to help with
- we'll have to handle the different newline styles -- `\n`/`\r\n`/`\r`, but we can put this off for now
- 
- this "black-esque" style, with the vertical wrapping, is not going to be to everyone's tastes -- we've had a fair bit of discussion about this. I do like it, because it basically requires no options, and so there can be no arguments about tuning it really. However, we'll likely end up having this as one of two options, with the other being a flatter wrapping style -- just something to bear in mind

Also, a couple of points to help with developing rust:
- it's good to run `cargo clippy` fairly frequently (or even better, set up LSP or something in your editor to run it automatically). This is the rust linter, and it's _fantastic_, it's really helpful for learning rust.
- `README.dev.md` describes how to run the tests

crates/fortitude_formatter/src/lib.rs
sorenhartmann: Yes, I have updated the methods to reflect this more clearly

fortitude/resources/test/fixtures/format/long-lines.f90
sorenhartmann: Yes, that is correct

crates/fortitude_formatter/src/lib.rs
sorenhartmann: I have renamed the method

crates/fortitude_formatter/src/lib.rs
sorenhartmann: I'll admit, I did not consider this edge case, but it seems to me it is handled somewhat reasonably

As far as I can tell, indentation should in general be handled by the `topiary` query, since indentation is stripped prior to formatting, with no way to preserve it.

Currently the line breaking logic does not allow the breaking of any nodes spanning more than on line, so introducing a comment, would effectively disable line breaking across the entire math expression, i.e.
```fortran
foo = abc + bar & ! comment
  + zag + def
```
would break to
```fortran
foo = abc + bar & ! comment
  + zag + def
```
and not
```fortran
foo = abc &
  + bar & ! comment
  + zag &
  + def
```

sorenhartmann:
@ZedThree @LiamPattinson Thank you for the feedback, it is really appreciated!

First of all, I apologize for the late reply.
 
- I am not really used to the GitHub workflow, but I have tried to respond to your comments above. Also, for rebasing, should I just force-push this branch?
- I have updated the topiary queries to handle some more constructs, and also handle some indentation across line-breaks.
- As far as I understand, according to https://github.com/PlasmaFAIR/fortitude/issues/3, `ruff` may also be able to handle formatting / line breaking to some extend? This seems very interesting! However, maybe we/I should wait with further development on this PR until the scope of a `ruff` solution is more clear?

ZedThree:
Thanks @sorenhartmann! Seems like the clippy issue is resolved, so no need to rebase.

We've got a session at a workshop next week, so I might see if we can get some people to try out this branch and get some real world feedback.

> As far as I understand, according to https://github.com/PlasmaFAIR/fortitude/issues/3, ruff may also be able to handle formatting / line breaking to some extend? This seems very interesting! However, maybe we/I should wait with further development on this PR until the scope of a ruff solution is more clear?

That's probably wise, I would really very much hate to waste your time! One thing about using ruff's solution would be that we can take advantage of any improvements they make very easily, but it's possible that it's a lot of work to use it in the first place.

I'm also not sure how flexible their solution is. If it's a bunch of work to achieve the same results you've got here, then it is probably not worth it.

PR #487, opened 2025-08-12
Prepare for v0.7.4 release

LiamPattinson:
Spotted one issue: we're currently using a specific commit of `tree-sitter-fortran` instead of a release version. Should we wait for a new `tree-sitter-fortran` release, or is that okay for now?

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -9,5 +9,5 @@ authors:
   orcid: "https://orcid.org/0000-0003-3092-1858"
 title: "fortitude"
 version: 0.7.4
-date-released: 2025-08-12
+date-released: 2025-08-14
 url: "https://github.com/PlasmaFAIR/fortitude"


LiamPattinson:
That's frustrating, not sure what's going wrong in the CI.

I don't think it's related to the failing build, but we're now getting a lot of quite annoying new linter warnings that don't come with a fix. Ruff ran into the same problem: https://github.com/astral-sh/ruff/pull/19807

Some of them look more serious, like `PartialEq` being basically meaningless in `OptionSet`. Weirdly, `SuperfluousImplicitNone` is listed as dead code, but no other rule structs.

ZedThree:
I don't think it's necessary to wait for a `tree-sitter-fortran` release, although we might want to bump the commit, as we're not on the latest one.

From what I can tell, we need to define `_BSD_SOURCE` to fix those undefined symbols. They come from `endian.h`, but are only defined if that feature macro is defined (see https://linux.die.net/man/3/be16toh). `tree-sitter` uses them to properly handle UTF-16, but they're non-standard so they have some portability header. The above is the correct fix for *nix systems.

Actually, following some links from `tree-sitter` issues about this, it might be sufficient to switch from `manylinux: auto` to `manylinux: 2_28`. Looks like `auto` means `2014` which is pretty old!

crates/fortitude_linter/src/rules/correctness/implicit_typing.rs
```diff
@@ -89,66 +89,6 @@ impl AstRule for InterfaceImplicitTyping {
...
 }
 
-/// ## What it does
-/// Checks for unnecessary `implicit none` in module procedures
-///
-/// ## Why is this bad?
-/// If a module has 'implicit none' set, it is not necessary to set it in contained
-/// functions and subroutines (except when using interfaces).
-#[derive(ViolationMetadata)]
-pub(crate) struct SuperfluousImplicitNone {
```
ZedThree, line 99: Had we moved this or something?

LiamPattinson: It was in the same file as the other `implicit none` rules, so I think it's just been lurking there since before we overhauled the rules categories. No idea how the linters missed it until now.

imciner2: Yea, this was moved during the rule changes, but that was pre-0.7.0, so interesting it was never flagged by the tooling until now. (I noted it here, but never got around to opening a PR to actually remove it https://github.com/PlasmaFAIR/fortitude/pull/329#discussion_r1995523277).

PR #489, opened 2025-08-13
CI: Attempt to simplify build-binaries by merging arch workflows

ZedThree:
Experiment to see if we can both simplify the CI a little and also switch to `manylinux_2_28`

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -249,7 +249,6 @@ jobs:
           install: |
             apt-get update
             apt-get install -y --no-install-recommends python3 python3-pip
-            pip3 install --break-system-packages -U pip
           run: |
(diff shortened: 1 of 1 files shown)


ZedThree:
I think the problem is that maturin falls back to the x86 manylinux container if it doesn't have one for the selected architecture, which isn't helpful.

ZedThree:
`ppc64` is no longer supported by `manylinux`, but as far as I can tell `ppc64le` should work fine on those systems, so nothing should be lost.

I forget exactly why I dropped `-musleabihf` now, I think it was another issue with the `manylinux` containers. I believe this is for 32-bit arm systems, so I don't think anyone will notice

LiamPattinson (approved):
It would be nice to have some usage stats across platforms to know if dropping the less common systems will actually affect anyone. If someone comes across this comment wondering why we dropped support for your platform, please raise an issue and let us know!

PR #491, opened 2025-08-18
Bump `tree-sitter-fortran` (fix `fmt` argument to IO statements)

### LINKED ISSUE #490: Fortitude 0.7.4 `E001 Syntax error` for valid program
Running fortitude 0.7.4 on the following program

```f90
program main
  use, intrinsic :: iso_fortran_env, only: error_unit
  implicit none

  character(len=*), parameter :: fmt = '(a)'
  write(error_unit, fmt) "Test"

end program main
```

gives

```
...
test_syntax.f90:6:19: E001 Syntax error
  |
5 |   character(len=*), parameter :: fmt = '(a)'
6 |   write(error_unit, fmt) "Test"
  |                   ^^^^^ E001
7 |
8 | end program main
  |
```

This wasn't reported in 0.7.3. This program compiles and runs as expected with `gfortran`.

ZedThree:
Closes #490

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ textwrap = { version = "0.16.0" }
 thiserror = { version = "1.0.58" }
 tree-sitter = "~0.25.0"
 tree-sitter-cli = "~0.25.0"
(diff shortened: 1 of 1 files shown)


LiamPattinson (approved):
Seems to be working just fine, thanks. Should I make 0.7.5 as a quick patch release?

Issue #493, opened 2025-08-20
Add check_name to gitlab code quality report format


fbjones-lanl:
The gitlab code quality report format output by `fortitude` lacks the `check_name` key.

See [this link] for the list of required keys.

ZedThree:
Fixed in #494

PR #494, opened 2025-08-20
Add check_name key to gitlab code quality report

fbjones-lanl:
This change adds the `check_name` key to the GitLab code quality report.

It is required per [this documentation](https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format) and it enables GitLab to provide nicely formatted feedback to users.

I tested these changes on the following Fortran file:

```fortran
program hello
    implicit none

end program hello
```

Command line:

```bat
..\target\release\fortitude.exe check --output-format=gitlab hello.f90
```

Output:

```json
[
  {
    "check_name": "C003",
    "description": "(C003) 'implicit none' missing 'external'",
    "fingerprint": "f1b8904469968fd6",
    "location": {
      "lines": {
        "begin": 2,
        "end": 2
      },
      "path": "hello.f90"
    },
    "severity": "major"
  }
]
```
... which is as intended.

`fortitude` currently returns the code a an `Option`, so to handle the `None` case I simply return the name `fortitude`. I considered `no-rule` or `no-code` but thought that might be confusing. The main point is to have that JSON field present so that the GitLab CI works.

Closes issue #493

### DIFF
--- a/crates/fortitude_linter/src/message/snapshots/fortitude_linter__message__gitlab__tests__output.snap
+++ b/crates/fortitude_linter/src/message/snapshots/fortitude_linter__message__gitlab__tests__output.snap
@@ -5,6 +5,7 @@ expression: redact_fingerprint(&content)
 [
   {
     "description": "(S201) 'implicit none' set on the enclosing module",
+    "check_name": "S201: superfluous-implicit-none",
     "severity": "major",
     "fingerprint": "<redacted>",
     "location": {
@@ -17,6 +18,7 @@ expression: redact_fingerprint(&content)
   },
   {
     "description": "(S061) end statement should read 'end subroutine foo'",
+    "check_name": "S061: unnamed-end-statement",
     "severity": "major",
     "fingerprint": "<redacted>",
     "location": {
@@ -29,6 +31,7 @@ expression: redact_fingerprint(&content)
   },
   {
     "description": "(PORT021) integer*4 is non-standard, use integer(4)",
+    "check_name": "PORT021: star-kind",
     "severity": "major",
     "fingerprint": "<redacted>",
     "location": {


crates/fortitude_linter/src/message/gitlab.rs
```diff
@@ -89,9 +89,15 @@ impl Serialize for SerializedMessages<'_> {
             } else {
                 message.body().to_string()
             };
+            let name = if let Some(rule) = message.rule() {
+                format!("{}", rule.noqa_code())
```
ZedThree: We could tweak this a little and put the text name here too:

```suggestion
                format!("{}: {}", rule.noqa_code(), rule.as_ref())
```

crates/fortitude_linter/src/message/gitlab.rs
```diff
@@ -89,9 +89,15 @@ impl Serialize for SerializedMessages<'_> {
             } else {
```
ZedThree, line 91: We could squash the two `if` statements together like this:

```rust
let (description, name) = if let Some(rule) = message.rule() {
    let description = format!("({}) {}", rule.noqa_code(), message.body());
    let name = format!("{}: {}", rule.noqa_code(), rule.as_ref())
    (description, name)
} else {
    (message.body().to_string(), String::from("fortitude"))
};
```

ZedThree:
Oh, you'll also need to update the test snapshots. The [dev README](https://github.com/PlasmaFAIR/fortitude/blob/main/README.dev.md) goes through this, but briefly:

1. Install `insta`: `cargo install cargo-insta`
2. Run the tests: `cargo test --workspace`
3. Review and approve the changed snapshots: `cargo insta review`

crates/fortitude_linter/src/message/gitlab.rs
fbjones-lanl: Sounds good - I'll make the change.

crates/fortitude_linter/src/message/gitlab.rs
fbjones-lanl: Agreed.

fbjones-lanl:
Thanks @ZedThree for the concise overview of testing with `insta`. I've never used it before. It's quite impressive!

ZedThree (approved):
Thanks again @fbjones-lanl, very much appreciated!

fbjones-lanl:
My pleasure @ZedThree - thanks for being so responsive. `fortitude` is a great product, and I'm encouraging my team to adopt it.

imciner2:
Hi @fbjones-lanl, it could be good if you can contribute to the documentation for how to configure the GitLab integration with Fortitude to display the results. This is actually something I will be needing to do eventually, and I think others might also find it useful as well.

PR #505, opened 2025-09-15
Small refactor for updated `do` loop grammar

ZedThree:
The current grammar struggled to deal with the two forms of `do` loops:

```f90
do
  ! ...
end do

do i = 1, 10
  ! ...
end do
```

This is particularly annoying for the format command where we want a newline at
the end of the statement, which isn't possible to capture with the query
language.

This update to the grammar adds a new node, `do_statement`, for the opening `do`
statement, and renames the block node to `do_loop`. This requires some small
refactoring to a couple of rules.

One side effect is being able to remove some allocations, which is nice.

(This is going into the `format-command` branch too, but I want to keep the two
branches from diverging too far)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2081,7 +2081,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter-fortran"
 version = "0.5.1"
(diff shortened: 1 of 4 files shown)


LiamPattinson (approved):
Looks good! Nice that we also got to upgrade a rule as a result.

PR #507, opened 2025-09-15
Bump Rust Edition and MSRV

ZedThree:
First commit bumps the edition and applies `cargo fix --edition`. Nothing major, just a couple of lifetime things, mostly around `if let Some(...)`.

Second commit applies new formatting. This is a much bigger change, but mostly looks pretty sensible.

Third commit just ignores the previous commit for `git blame` (as it's changed a lot of lines)

### DIFF
--- a/crates/fortitude/src/main.rs
+++ b/crates/fortitude/src/main.rs
@@ -1,4 +1,4 @@
-use std::io::{stdout, Write};
+use std::io::{Write, stdout};
 use std::process::ExitCode;
 
 use anyhow::Result;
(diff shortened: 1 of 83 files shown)


crates/fortitude/src/main.rs
```diff
@@ -34,10 +34,10 @@ fn main() -> Result<ExitCode> {
                 //
                 // See: https://github.com/BurntSushi/ripgrep/blob/bf63fe8f258afc09bae6caa48f0ae35eaf115005/crates/core/main.rs#L47C1-L61C14
                 for cause in err.chain() {
-                    if let Some(ioerr) = cause.downcast_ref::<std::io::Error>() {
-                        if ioerr.kind() == std::io::ErrorKind::BrokenPipe {
-                            return Ok(ExitCode::from(0));
-                        }
+                    if let Some(ioerr) = cause.downcast_ref::<std::io::Error>()
+                        && ioerr.kind() == std::io::ErrorKind::BrokenPipe
```
LiamPattinson, line 38: Oh neat, if I'd known upgrading the Rust edition would allow this I'd have done it ages ago.

LiamPattinson (approved):
Cheers, looks good :+1:

Issue #405, opened 2025-04-04
Better support for preprocessing files


melven:
Thanks for the great work, this new linter already gives helpful insights for existing Fortran code and we are currently integrating it in our CI pipelines.

Unfortunately, we use some C-preprocessor macros for error handling in our code - so in some places half of the lines use macros.
(In the style: `MY_ASSERT(allocated(x), ierr)` or `MY_CHECK_IERR(call someFunction(..., ierr), ierr)`)

It is already possible to provide preprocessed input via stdin with, e.g., :
`gfortran -E -I... example.F90 | fortitude check --line-length=132 --stdin-filename=example.F90`

This has 2 shortcomings to me so far:
* the line numbers are wrong as stated in #271 
* this is only for one file - I would need to script this and combine the results for all files

Related issue #356 

To me, providing some preprocessing command could be sufficient, like, e.g.:
```
fortitude check --preprocessor-command="gfortran -E -IsomePath"
```

In the long run, supporting `compile_commands.json` could give a real benefit (can be exported from CMAKE, see https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).
I can imagine that this would allow checking for more complex erroneous patterns that span multiple files.

ZedThree:
Hi @melven, thanks for your support, I'm really glad fortitude is proving useful to you!

This is absolutely on our to do list! Unfortunately, to get this right is really quite complicated. See [this comment](https://github.com/PlasmaFAIR/fortitude/issues/356#issuecomment-2748782683) on fypp support for some details.

Essentially, in order to be most useful, we have to track how the processed text maps back onto the original text. For simple `#if` blocks, this is fairly easy as basically only the line numbers change, but for macros as in your examples, then both the line and column numbers change. We know what to do, just need to find the time to do it!

We will absolutely support `compile_commands.json`.

LiamPattinson:
Hi @melven, thanks for the kind words!

We've been discussing preprocessor support for some time now, and I'm surprised to see that we don't already have an issue raised for it. It's definitely something we would like to see, but there are numerous technical challenges to implementing it correctly.

The primary challenge is that, in general, it isn't possible to know if preprocessor code will result in valid Fortran without first preprocessing. The tree-sitter parser we're using can successfully handle simple preprocessor commands that don't break Fortran syntax, such as:

```f90
call some_subroutine(x, y, z)
#ifdef DEBUG
print *, x
#endif
```

[...]

PR #510, opened 2025-09-17
Add C preprocessor

LiamPattinson:
Adds the beginnings of a provenance-tracking C-preprocessor.

Takes some steps steps towards https://github.com/PlasmaFAIR/fortitude/issues/405.

I'm raising this as a draft PR as I'll be away for a while and won't have much of a chance to work on it further. I might have some time in early October, but after that I won't be able to give it much thought until late November at the earliest. I'd be happy for others to open PRs into this branch in the meantime.

So far handles the following:

- Fortran-appropriate tokenisation
- Backslash line continuations
- `#define` (object macros only)
- `#undef`
- Some conditional control:
  - `#ifdef`
  - `#ifndef`
  - `#else`
  - `#endif`
- `__LINE__`
- `__FILE__`
- `__DATE__`
- `__TIME__`
- Tracks the provenance of each snippet, e.g. code from within a macro expansion is mapped back to the relevant `#define`.

There's still a lot to add though:

- `#if`, `#elif`
  - `defined()`
  - All associated boolean logic...
- `#include`
  - Cross-platform include paths
- Function macros
  - Stringification
  - Concatenation (need to handle old style `/**/` and new style `##`)
- Provenances should be nestable
- Error reporting currently isn't very helpful

It will also need to be integrated into the rest of the code:

- `fortitude preprocess` mode and relevant options
  - Global defines
  - Per-file defines
  - Include path
  - Per-file include path (?)
- Integration into the linter
- Integration into the LSP/formatter (I'd rather not think about this for now!)

Some stretch goals:

- Ability to set user profiles for things like debug mode or to set environment-specific defines. They could then be activated by calling something like `fortitude preprocess --profile=debug`.
- Ability to read from `compile_commands.json` to set per-file defines and include paths.

---

There's a known bug in `#define`, as it currently saves the text string at the point the macro is defined, but it should actually save a token stream, as macro expansion occurs where the macro is used, not where it's defined:

```c
#define X Y
#define Y 5

X  // expands to '5', not 'Y'
```


crates/fortitude_preprocessor/src/cpp_tokens.rs
```diff
@@ -0,0 +1,618 @@
...
+        if self.iter.peek() == Some(&b'#') {
+            self.step();
+            return Some(self.emit(start, CppTokenKind::Concatenation));
+        }
+        // Consume any whitespace after the '#'.
+        let _ = self.consume_whitespace();
+        // Consume an identifier for the directive name.
+        match self.consume_identifier() {
+            Some(Ok(directive)) => {
+                let directive_kind = match directive.text {
```
ZedThree: I think you can replace this with `#[derive(strum_macros::EnumStr)]` on `CppDirectiveKind`, using `#[strum(default)]` on `Stringification`

crates/fortitude_preprocessor/src/tokens.rs
```diff
@@ -0,0 +1,618 @@
...
+            if i % 2 == 0 {
+                assert_eq!(token.kind, CppTokenKind::Whitespace);
+            } else {
+                assert_eq!(token.kind, CppTokenKind::Newline);
+            }
+        }
+    }
+
+    #[test]
+    fn test_identifier_tokenization() {
```
ZedThree, line 525: Could do pretty much all of these tests as parameterised tests with `test_case`?

crates/fortitude_preprocessor/src/cpp.rs
```diff
@@ -0,0 +1,852 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Provenance {
+    /// A system-defined macro, e.g. __DATE__ or __TIME__.
+    SystemDefined,
+    /// A user-defined macro, e.g. passed in via command line.
+    UserDefined,
+    /// A macro defined in the source file.
+    FileDefined {
+        start: usize,
+        end: usize,
```
ZedThree: Is the plan to switch to `ruff_text_size::TextRange`?

crates/fortitude_preprocessor/src/cpp.rs
```diff
@@ -0,0 +1,852 @@
...
+        path: PathBuf,
+    },
+    /// Plain text in the source file, not from a macro.
+    LocalText { start: usize, end: usize },
+}
+
+/// A snippet of text with provenance information.
+pub struct Snippet {
+    text: String,
```
ZedThree: Is this the transformed text? A possible future optimisation would be to use `Cow<str>` here to avoid unnecessary allocations

crates/fortitude_preprocessor/src/cpp.rs
```diff
@@ -0,0 +1,852 @@
...
+    }
+}
+
+/// A collection of snippets.
+pub struct Snippets {
+    // Collected snippets
+    inner: Vec<Snippet>,
+}
+
+impl Default for Snippets {
```
ZedThree: Is it not possible to do `#[derive(Default)]` for this?

crates/fortitude_preprocessor/src/cpp.rs
```diff
@@ -0,0 +1,852 @@
...
+            .collect::<String>()
+    }
+}
+
+/// A logical line of code, which may span multiple physical lines due to
+/// line continuations. Tracks the byte offset of each location of the
+/// logical line.
+pub struct LogicalLine {
+    /// The text of the logical line.
+    text: String,
```
ZedThree: Another place for `Cow<str>`?

crates/fortitude_preprocessor/src/cpp.rs
```diff
@@ -0,0 +1,852 @@
...
+    /// The byte offsets of each character in the logical line
+    /// relative to the start of the source file.
+    byte_offsets: Vec<usize>,
```
ZedThree: Storing offset of each character individually worries me a little bit. Potential optimisation would be `Vec<TextRange>` storing contiguous chunks?

ZedThree:
A couple of implementations in Python worth looking at for inspiration:

- [`fortls`, the Fortran LSP server](https://github.com/fortran-lang/fortls/blob/78ad649ab1a866a40591badb094c8e8849bb15c5/fortls/parsers/internal/parser.py#L2028)
  - Only 300 lines (!) and only a minimal implementation, but useful *enough* I think
- [`pcpp`](https://github.com/ned14/pcpp/tree/master)
  - Much more feature complete, has some interesting additional features
  - Has a much larger test suite

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: Oops, that's embarrassing, I think that's something Copilot must have generated without me noticing :upside_down_face:

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: Yes, the snippets will be the final text of the preprocessed file. I haven't used `Cow<str>` before, but I'd be happy to look into it if it'll save some copies.

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: I'd been trying to avoid it, as I find `TextSize`, `TextRange`, `OneIndexed`, etc to be really clunky compared to just plain `usize`. I suppose we'll probably have to once we start integrating it into the linter.

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: Yes, I was planning to optimise that at some point. I was thinking it would only be necessary to store the discontinuities, but thought I would implement it the easy way to begin.

ZedThree: :+1: Sounds good! Make it work first, then make it fast :)

crates/fortitude_preprocessor/src/cpp.rs
ZedThree: Yeah, they're definitely a bit more annoying, but `TextRange` is 1/4 the size of `(usize, usize)`, and if we're going to make a whole bunch of these, it's probably worth it.

But definitely do the simple thing and get it working first!

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: It turns out it was actually auto-generated by clippy, and `#[derive(Default)]` requires defining `Default` on `Snippet` and `Provenance`, which doesn't make much sense.

crates/fortitude_preprocessor/src/cpp_tokens.rs
LiamPattinson: Done in https://github.com/PlasmaFAIR/fortitude/pull/510/commits/3fe737b2ffebf3cb31dc0d6821d3a894b86eef01 :+1:

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: I tried this with `Cow<'a, str>` and ran into a host of lifetime issues. For now I've left it as a `String`, though I was able to use `Cow<'a, str>` in `LogicalLine`.

crates/fortitude_preprocessor/src/cpp.rs
LiamPattinson: Done in https://github.com/PlasmaFAIR/fortitude/pull/510/commits/0b4cbd971639883a9c1ae954cb09007677be47e1

LiamPattinson:
I was trying to get some better tests into the preprocessor, and the ones I could find in the wild seem to make liberal use of `/**/` for concatenation and other tricks , so I thought I'd quickly implement that.

In doing so, I got into a bit of bother trying to figure out how `/**/` interacts with escaped newlines, or how either one of them is supposed to work alongside multiline strings. Turns out, the c-preprocessor doesn't get on well with multiline strings _at all_:

```f90
#define x 12
print *, "hello x &
  & x world!", x
```

After preprocessing:

```f90
print *, "hello x &
  & 12 world!", x
```

It interprets the first `x` on line 3 as an identifier and expands it as a macro, while the second `x` is interpreted as being inside of a string and is left alone, whereas the opposite should be true.

I was then wondering why it didn't interpret the `!` as the start of a comment, and therefore also learned that it doesn't have any special handling for Fortran comments whatsoever:

```f90
#define x 12
! x
! "x
```

becomes:

```f90
! 12
! "x
```

This means I'll have to go back and rethink some of my earlier work, as my tokeniser was written with the expectation that comments and multiline strings were meaningful.

LiamPattinson:
> Also sounds like a good candidate for a rule? `non-portable-C-comment-in-macro`?

Good idea! I don't think it needs to be within a macro necessarily, as you can do concatenation tricks with them anywhere, and yet `pcpp` will always replace them with a space. The best way to handle it would probably be to find the pattern `/*`, checking that it isn't inside of a string, and iterate until finding `*/` or EOF. I suppose you might also need to be careful not to match `/*` inside other C-comments too.

crates/fortitude_preprocessor/src/definitions.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+    }
+
+    pub fn get(&self, key: &str) -> Option<&Definition> {
+        self.inner.get(key)
+    }
+
+    fn expand_macro_tokens(&self, tokens: &[CppToken]) -> anyhow::Result<String> {
+        // Due to the removal of comments in replacement text, it is possible
+        // to encounter
```
ZedThree: Think something's happened to this comment!

crates/fortitude_preprocessor/src/definitions.rs
```diff
@@ -0,0 +1,313 @@
...
+                                if let Some(next) = iter.peek() {
+                                    if next.kind == CppTokenKind::Punctuator && next.text == ")" {
+                                        iter.next(); // Consume ')'
+                                        // Empty argument list
+                                        let (_, replacement) =
+                                            self.expand_function_macro(&token.text, &arglist)?;
+                                        result.push_str(&replacement);
+                                        continue;
+                                    }
+                                }
```
ZedThree: Do we need special handling for empty arguments? What if we move the `arglist.push(Vec::new())` to the `_` branch of the below `match`?

crates/fortitude_preprocessor/src/definitions.rs
```diff
@@ -0,0 +1,313 @@
...
+        // Get token list of replacement text
+        let mut replacement = Vec::new();
+        for token in iter.by_ref() {
+            if token.kind == CppTokenKind::Newline {
+                break;
+            }
+            replacement.push(token.to_owned());
+        }
+        let (start, end) = line.offset_range();
+        // TODO: handle redefines properly
```
ZedThree, line 237: It would be nice to raise a violation for redefines, but I'm not sure if it's a good idea to mix up rule checking with the preprocessing. Might be the only way to do some kinds of rules?

crates/fortitude_preprocessor/src/lib.rs
```diff
@@ -0,0 +1,790 @@
...
+    }
+}
+
+/// a C preprocessor implementation that preserves provenance information
+/// when expanding macros and includes.
+pub struct CPreprocessor {
+    #[allow(dead_code)]
+    path: PathBuf,
+    snippets: Snippets,
+    #[allow(dead_code)]
```
ZedThree: Do we still need the `allow`s?

crates/fortitude_preprocessor/src/lib.rs
```diff
@@ -0,0 +1,790 @@
...
+              print *, foo, foo(5), foo(foo + 2), foo(foo(7) + W)
+              print *, bar("hello, ","world!")
+              print *, baz, baz(), foo(baz())
+            end program p
+        "#
+        );
+        let (output, snippets) = preprocess(code)?;
+        let expected = dedent!(
+            r#"
+            program p
```
ZedThree, line 521: One big difference with `cpp` is we consume the directive lines and `cpp` replaces them with blank lines (or `#line`/`# <line> <file>` directives). It obviously won't affect the resulting AST, but might be a good idea to minimise diffs with other preprocessors? Probably just an idea for future work after getting it in?

crates/fortitude_preprocessor/src/definitions.rs
LiamPattinson: I've been thinking it might be a good idea to build up a vec of violations as the preprocessor does its thing, as a lot of preprocessor-related issues won't be visible in the final text and also won't be parseable in the original. This might result in a circular dependency though, so some bits of the project would need to be reorganised.

crates/fortitude_preprocessor/src/definitions.rs
LiamPattinson: That's worrying, whatever it must be possible to encounter must be pretty dangerous.

I was planning to implement something to merge adjacent identifier/number tokens, which should only be possible if they're originally separated by a C-comment, as this would possibly replicate gfortran's behaviour. Maybe I was half way through writing this when I decided against it?

crates/fortitude_preprocessor/src/definitions.rs
LiamPattinson: I might have to rethink that whole algorithm. I wrote it that way because it's possible to omit arguments to function macros, so in case `FUNCTION(,X)` comes up, you need to have an empty token list in the first position before adding a new token list to capture `X`. To handle the case of a function macro that takes no arguments, you need to exit before adding that initial empty list. However, my current implementation would fail in this case:

```f90
#define FUNC(X) X
print *, FUNC() FUNC(Y)
```
It would incorrectly interpret the first call as a function macro that takes no arguments, whereas in reality it's just that the first arg was omitted.

LiamPattinson: I _think_ I have this fixed up now in https://github.com/PlasmaFAIR/fortitude/pull/510/commits/b73b0924442fba6f2a921a918bf0b23720fe8497

PR #518, opened 2025-10-01
Add unsafe fix for `implicit-typing`, and check non-module procedures

### LINKED ISSUE #516: Add (unsafe) fix for `implicit-typing`?
We should be able to automatically insert `implicit none` into `module` or `program` scopes, although this may technically be an unsafe fix. The only "tricky" bit should be finding the end of any `use` statements, unless I'm missing something?

### LINKED ISSUE #517: Check for `implicit-typing` in non-module procedures
We only check `implicit-typing` in `module`, `submodule`, and `program`. Should we also check in `subroutine` and `function` when they _aren't_ in a module or program?

ZedThree:
Fixes #516 
Fixes #517

The indentation isn't perfect, it would be easier once we get something like ruff's `Stylist` struct


LiamPattinson (approved):
Thanks @ZedThree, all looks good to me. I agree that it would be nice to fix up the indentation, but it might be tricky to achieve in general without something like a `Stylist` struct.

PR #522, opened 2025-10-02
Add rule for nonconforming tab characters

ZedThree:
Rendering currently is not great:

```
crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90:4:1: PORT031 Invalid tab character
  |
2 | ! the horrible whitespace!
3 | program invalid_tab
4 |     implicit none
  | ^ PORT031
5 |     print*, "Don't flag this:    "
6 |     ! Or this:|    |
  |
  = help: Replace with spaces

crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90:7:7: PORT031 Invalid tab character
  |
5 |     print*, "Don't flag this:    "
6 |     ! Or this:|    |
7 |     if    (.true.)    then
  |       ^ PORT031
8 |         print*, "Mixed tab/space"
9 |     else
  |
  = help: Replace with spaces

crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90:7:16: PORT031 Invalid tab character
  |
5 |     print*, "Don't flag this:    "
6 |     ! Or this:|    |
7 |     if    (.true.)    then
  |               ^ PORT031
8 |         print*, "Mixed tab/space"
9 |     else
  |
  = help: Replace with spaces

crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90:8:5: PORT031 Invalid tab character
   |
 6 |     ! Or this:|    |
 7 |     if    (.true.)    then
 8 |         print*, "Mixed tab/space"
   |     ^ PORT031
 9 |     else
10 |         print*, "Two tabs"
   |
   = help: Replace with spaces

crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90:10:1: PORT031 Invalid tab character
   |
 8 |         print*, "Mixed tab/space"
 9 |     else
10 |         print*, "Two tabs"
   | ^ PORT031
11 |     end if
12 | end program invalid_tab
   |
   = help: Replace with spaces

crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90:10:2: PORT031 Invalid tab character
   |
 8 |         print*, "Mixed tab/space"
 9 |     else
10 |         print*, "Two tabs"
   | ^ PORT031
11 |     end if
12 | end program invalid_tab
   |
   = help: Replace with spaces
```

I think this is maybe at least partly an `annotate-snippets` issue. Might need to either bump the version or switch to `ruff-annotate-snippets`.

Possibly something even worse, and fix something our side.

The unsafe fix could also be improved with:

a) getting that `Stylist` struct in to improve the tab size,
b) work out if we're indentation and if not, replace with a single space

### DIFF
--- a/crates/fortitude_linter/src/rules/mod.rs
+++ b/crates/fortitude_linter/src/rules/mod.rs
@@ -124,6 +124,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Portability, "011") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKind),
         (Portability, "012") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKindSuffix),
         (Portability, "021") => (RuleGroup::Stable, Ast, Default, portability::star_kinds::StarKind),
+        (Portability, "031") => (RuleGroup::Preview, None, Optional, portability::invalid_tab::InvalidTab),
 
         // style
         (Style, "001") => (RuleGroup::Stable, Text, Default, style::line_length::LineTooLong),
--- a/crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90
+++ b/crates/fortitude_linter/resources/test/fixtures/portability/PORT031.f90
@@ -1,0 +1,12 @@
+! If you touch this flie, be careful your editor doesn't auto-correct
+! the horrible whitespace!
+program invalid_tab
+	implicit none
+    print*, "Don't flag this:	"
+    ! Or this:|	|
+    if	(.true.)	then
+    	print*, "Mixed tab/space"
+    else
+		print*, "Two tabs"
+    end if
+end program invalid_tab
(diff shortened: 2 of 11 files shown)


ZedThree:
> Would it not be possible to implement this as an `AstRule` that triggers on `translation_unit`?

Yes, but if we want to eventually switch away from the entrypoints system, we'll need to handle it like this anyway.

crates/fortitude_linter/src/rules/mod.rs
```diff
@@ -124,6 +124,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Portability, "011") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKind),
         (Portability, "012") => (RuleGroup::Stable, Ast, Default, portability::literal_kinds::LiteralKindSuffix),
         (Portability, "021") => (RuleGroup::Stable, Ast, Default, portability::star_kinds::StarKind),
+        (Portability, "031") => (RuleGroup::Preview, None, Optional, portability::invalid_tab::InvalidTab),
```
LiamPattinson: I'd argue that this should be on by default, though that might upset some people!

LiamPattinson (approved):
Cheers, all looks good! I changed the title just so it would be clear in the changelog that two rules have been added.

ZedThree:
Yeah, I'm surprised it's not available even as an extension. I'm inclined to move it to `error` and have it on by default. I do also agree that probably `invalid-tab` should be on by default

LiamPattinson (approved):

Issue #513, opened 2025-09-23
Better config file finding


ZedThree:
Currently we only look in the directory where `fortitude` is run and its parent directories for config files. This can give different behaviour when running in- or outside a project:

```bash
# Ignores config files inside project/
$ fortitude check project/
# Uses project config files
$ cd project; fortitude check
```

Instead, we should port [ruff's behaviour](https://docs.astral.sh/ruff/configuration/#config-file-discovery), which by default will consider the directory hierarchy of the files being checked, with some customisation options. They also support user config files.

We have a bunch of the machinery to support hierarchical config files if we want too. This might be useful for projects to ignore certain rules within bundled third-party dependencies.

PR #509, opened 2025-09-17
Add language server

ZedThree:
Closes #38

Add LSP language server. This is a fairly straight port of ruff's server, I just had to do it in chunks in order to get my head around it, and which bits we do and don't need. We get diagnostics, hover-text on `allow` comments, and auto-fixes (either to a specific violation, or the whole document).

I've not done the docs yet, there's a few links still to `astral.sh` in a couple of files, and there's one or two tests I've not ported yet either, but I wanted to open the PR so people could start playing with it.

I've tested it works in a bunch of editors: emacs, neovim, helix, kate. VS Code annoyingly needs a whole extension (sorting of defeating the purpose of LSP?! what's going on, microsoft?), and sublime and pycharm aren't available in my package manager.

PR #534, opened 2025-11-10
Add top-level `include` option; deprecate `files/file_extensions`

ZedThree:
This change is a step towards supporting ruff-style hierarchical settings files, and makes it easier to check if a given file should be included or not.

Also making it a top-level option in preparation for the format command

This is groundwork for #509 and #513


LiamPattinson (approved):
Thanks for this, I had a read through and no issues jumped out at me.

I'm surprised to see the `get_files` function has survived yet another refactor. Is it the case that you're building towards Ruff's `Resolver` by increments?

PR #535, opened 2025-11-10
Add `--isolated` option, and user-specific config files

ZedThree:
Builds on #534, another step towards #509 and #513. This now has most of the machinery to also implement extending other config files.

Adds options to ignore all config files, and to find user-specific config files (if no others are found)

Next step is to change how we find files to discover the closest config file at the same time.

### DIFF
--- a/crates/fortitude/tests/check.rs
+++ b/crates/fortitude/tests/check.rs
@@ -10,6 +10,11 @@ fn fortitude_cmd() -> Command {
 }
 
 /// Builder for `fortitude check` commands
+///
+/// This is used to ensure we always set either:
+///   a) `--isolated`, or
+///   b) `--config-file`
+/// to make sure we don't get any interference from user-specific config files
 #[derive(Debug, Default)]
 struct FortitudeCheck<'a> {
     config: Option<&'a Path>,


crates/fortitude_workspace/src/resolver.rs
```diff
@@ -0,0 +1,202 @@
...
+use itertools::Itertools;
+use path_absolutize::path_dedot;
+
+use fortitude_linter::{fs, settings::Settings};
+
+use crate::configuration::{Configuration, ConfigurationTransformer, load_options};
+
+/// The configuration information from a `fortitude.toml` file.
+#[derive(Debug)]
+pub struct FortitudeConfig {
```
ZedThree: Ruff calls this `PyprojectConfig`, maybe this is more like `ConfigFile`?

crates/fortitude_workspace/src/resolver.rs
```diff
@@ -0,0 +1,202 @@
...
+pub enum FortconfigDiscoveryStrategy {
+    /// Use a fixed `fortitude.toml` file for all Python files (i.e., one
+    /// provided on the command-line).
+    Fixed,
+    /// Use the closest `fortitude.toml` file in the filesystem hierarchy, or
+    /// the default settings.
+    Hierarchical,
+}
+
+impl FortconfigDiscoveryStrategy {
```
ZedThree: ...and this `ConfigFileDiscoveryStrategy`?

crates/fortitude_workspace/src/resolver.rs
LiamPattinson: Yes, I think saying `File` in the name will better distinguish it from `Settings`/`Options`/`Configuration`

crates/fortitude_workspace/src/resolver.rs
LiamPattinson: Agreed

docs/configuration.md
```diff
@@ -1,24 +1,34 @@
...
+`fpm.toml` file in the current directory, or one of its parents.
 
+If there is no configuration file in the current directory or its parents,
+Fortitude will look in `${config_dir}/fortitude/` for a user-specific file (see
+[`etcetera`'s base
+strategy](https://docs.rs/etcetera/latest/etcetera/#native-strategy) for how
+`${config_dir}` is determined).
 
-=== "`fortitude.toml` or `.fortitude.toml`"
+If no configuration file is found, Fortitude will fallback to a default configuration.
```
LiamPattinson: A very petty grammar thing:
```suggestion
If no configuration file is found, Fortitude will fall back to a default configuration.
```

crates/fortitude_workspace/src/resolver.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Python file.
+#[derive(Debug, Clone, Copy, is_macro::Is)]
+pub enum ConfigFileDiscoveryStrategy {
+    /// Use a fixed `fortitude.toml` file for all Python files (i.e., one
+    /// provided on the command-line).
+    Fixed,
+    /// Use the closest `fortitude.toml` file in the filesystem hierarchy, or
+    /// the default settings.
+    Hierarchical,
+}
```
LiamPattinson, line 53:
```suggestion
/// The strategy used to discover the relevant `fortitude.toml` file for each
/// Fortran file.
#[derive(Debug, Clone, Copy, is_macro::Is)]
pub enum ConfigFileDiscoveryStrategy {
    /// Use a fixed `fortitude.toml` file for all Fortran files (i.e., one
    /// provided on the command-line).
    Fixed,
    /// Use the closest `fortitude.toml` file in the filesystem hierarchy, or
    /// the default settings.
    Hierarchical,
}
```

crates/fortitude/src/check.rs
```diff
@@ -137,20 +135,13 @@ enum CheckStatus {
...
+    let (cli, config_arguments) = args.partition(global_options)?;
 
-    // Now, we can override settings from the config file with options
-    // from the CLI
-    let config = config_arguments.transform(file_configuration);
-    let settings = config.into_settings(&project_root)?;
+    // Construct the "default" settings. These are used when no `fortitude.toml`
+    // files are present, or files are injected from outside of the hierarchy.
+    let file_configuration = resolve::resolve(&config_arguments, cli.stdin_filename.as_deref())?;
+    let settings = file_configuration.settings;
```
LiamPattinson, line 144: Unless I'm misunderstanding, I don't think this comment is accurate, as `resolve(...)` does look for a `fortitude.toml` relative to the cwd.

crates/fortitude/src/resolve.rs
```diff
@@ -0,0 +1,112 @@
...
+) -> Result<ConfigFile> {
+    let Ok(cwd) = std::env::current_dir() else {
+        bail!("Working directory does not exist")
+    };
+
+    // First priority: if we're running in isolated mode, use the default settings.
+    if config_arguments.isolated {
+        let config = config_arguments.transform(Configuration::default());
+        let settings = config.into_settings(&cwd)?;
+        debug!("Isolated mode, not reading any pyproject.toml");
```
LiamPattinson:
```suggestion
        debug!("Isolated mode, not reading any TOML file");
```

crates/fortitude/src/resolve.rs
```diff
@@ -0,0 +1,112 @@
...
+            settings,
+            Some(fortconfig.to_path_buf()),
+        ));
+    }
+
+    // Third priority: find a `fortitude.toml` file in either an ancestor of
+    // `stdin_filename` (if set) or the current working path all paths relative to
+    // that directory. (With `Strategy::Hierarchical`, we'll end up finding
+    // the "closest" `fortitude.toml` file for every Fortran file later on,
+    // so these act as the "default" settings.)
```
LiamPattinson, line 64: I'm having trouble parsing 'the current working path all paths relative to that directory'. Should it be 'the current working path. All paths are determined relative to that directory'?

crates/fortitude/src/resolve.rs
```diff
@@ -0,0 +1,112 @@
...
+            &pyproject,
+            config_arguments,
+            ConfigurationOrigin::UserSettings,
+        )?;
+        return Ok(ConfigFile::new(
+            ConfigFileDiscoveryStrategy::Hierarchical,
+            settings,
+            Some(pyproject),
+        ));
+    }
```
LiamPattinson:
```suggestion
    if let Some(user_config) = find_user_settings_toml() {
        debug!(
            "Using configuration file (via cwd) at: {}",
            user_config.display()
        );
        let settings = resolve_root_settings(
            &user_config,
            config_arguments,
            ConfigurationOrigin::UserSettings,
        )?;
        return Ok(ConfigFile::new(
            ConfigFileDiscoveryStrategy::Hierarchical,
            settings,
            Some(user_config),
        ));
    }
```

crates/fortitude_workspace/src/resolver.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    let configuration = resolve_configuration(fortconfig, transformer, origin)?;
+    let project_root = relativity.resolve(fortconfig);
+    let settings = configuration.into_settings(project_root)?;
+    Ok((project_root, settings))
+}
+
+/// Extract the [`Settings`] from a given `fortitude.toml` and process the
+/// configuration with the given [`ConfigurationTransformer`].
+pub fn resolve_root_settings(
```
LiamPattinson, line 183: I'm confused about the distinction between `resolve_scoped_settings` and `resolve_root_settings` and why only the latter is used in `fortitude/src/resolve`.

In what circumstance is the former used?

When a path is passed to `resolve_root_settings`, is it assumed to already be the project root?

crates/fortitude/tests/check.rs
LiamPattinson, line 1: With the exception of the new test for `--isolated` at the end, am I right that these changes are just a 'drive-by' update to call `fortitude check` a bit more cleanly throughout the tests?

crates/fortitude/src/resolve.rs
LiamPattinson, line 64: To clarify, does this mean that the immediate closest ancestor config to the cwd will be the 'default', and once hierarchical configs are in, the chain of configs above any given file is applied on top of that?

So in this scenario, if the user runs `fortitude check ../dir_1`, the configs are applied in the numbered order:

```
root toml (2) --+----- dir_1
                |      |
                |      +--- dir_1 toml (3)
                |      |
                |      +--- dir_1.f90
                |
                +----- dir_2 (cwd)
                       |
                       +--- dir_2 toml (1)
```

That doesn't seem right to me, so I think I've probably misunderstood something.

crates/fortitude/tests/check.rs
ZedThree: Ah, no, because we now also look for user-specific config files (e.g. in `~/.config/fortitude`, we need to make sure when we run `fortitude` as a process in our tests we either:

a) use `--isolated`
b) use `--config-file`

to make sure we don't pick up unintended settings from the user-specific file. I'll add this detail as a comment

crates/fortitude/src/resolve.rs
ZedThree: Yes, I think the ruff folk just missed a comma or word there

ZedThree: The hierarchical config stuff coming later will require an explicit `extend = "path/to/base/config.toml"` setting. Config files are never implicitly merged, it's only the CLI settings that get merged in.

So in your scenario, `1` gets picked up as the default settings, and then `3` will be used for `dir_1.f90`. Both `1` and `2` will be completely ignored.

If `3` didn't exist, then `1` would be used instead, and `2` again will be ignored.

Only if neither `1` nor `3` exist will `2` get picked up.

In any case, anything set on the CLI will override (or extend) the actual config file used.

The docs in #537 should hopefully explain this better: https://github.com/PlasmaFAIR/fortitude/blob/c4530503a7f06be5a52bc29a8bc31477d6956af2/docs/configuration.md?plain=1#L32-L58

crates/fortitude/src/check.rs
ZedThree: I think it means when there's no config files are found "alongside" the files to be checked

crates/fortitude/src/resolve.rs
LiamPattinson: Thanks, that really clears things up!

crates/fortitude_workspace/src/resolver.rs
ZedThree: > I'm confused about the distinction between `resolve_scoped_settings` and `resolve_root_settings` and why only the latter is used in `fortitude/src/resolve`.
> 
> In what circumstance is the former used?

It's used in more places in #537 to get the actual settings for a given file. It was just easier to introduce it here

> When a path is passed to `resolve_root_settings`, is it assumed to already be the project root?

Yes, although it's more like "what scope do these settings apply to" -- in #537, we don't resolve the settings for each file individually, we cache the created `Settings` for each found config file for the scope it applies to (something like as if each config file was its own project root)

crates/fortitude/tests/check.rs
LiamPattinson: Thanks, that makes sense

PR #536, opened 2025-11-12
Access all rules through `Settings`

ZedThree:
This is another step towards #513, by removing the need to pass around the extra collections of rules and just using `CheckSettings::rules` directly, we're prepared to have multiple `Settings`, which might have different rules enabled.

This also gets rid of the `PathRule` and `TextRule` traits and explicitly checks if those rules are enabled -- we only had four total of those rules, so this isn't such a big deal, but also this is where we want to end up with the enormous visitor of death for the AST rules.

This looks like a big change, but it's mostly changing which type is used in a lot of places. The change to passing `CheckSettings` instead of `Settings` into `AstRule::check` isn't strictly required, but I think is sensible.


CONTRIBUTING.md
```diff
@@ -168,6 +166,20 @@ There are several steps required to add a new rule to Fortitude:
 tree-sitter build
 tree-sitter parse /path/to/fortran/file.f90
 ```
+    - If it doesn't need the AST for whatever reason (for example, it's
+      checking the filename), then you just need to implement some
+      function that returns a `Diagnostic` containing your new
+      `Violation`.
+5. For non-AST rules, you'll have to also specifically call the
```
LiamPattinson, line 173: The list item after this is also numbered '5'

ZedThree: I'll fix this in #537 -- but markdown will render this correctly, it doesn't actually matter what you number them! Ruff just uses `1.` for instance

PR #537, opened 2025-11-13
Improve config file discovery

### LINKED ISSUE #513: Better config file finding

ZedThree:
Closes #513 

Includes #536 and #535, so probably best wait to review this until after they've been merged

Config files are now fully discovered following ruff's behaviour. For example:

```console
$ pwd
/tmp/foo
$ ls project/
fortitude.toml foo.f90
# Previous behaviour ignores project/fortitude.toml
$ fortitude check project/ --show-settings | grep check.project_root
check.project_root = "/tmp/foo"
# New behaviour uses project/fortitude.toml as expected
$ fortitude check project/ --show-settings | grep check.project_root
check.project_root = "/tmp/foo/project"
```

Extending other files will be another PR.

One bug fix which changes behaviour: previously `exclude` just extended `exclude` including the built in excludes. Now `exclude` _overrides_ previous settings, including the built ins. However, most of the built ins should be in project's `.gitignore` which will still be respected, so this probably won't be noticeable. Additionally, the recommended setting to use is `extend-exclude`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,7 +180,7 @@ tree-sitter parse /path/to/fortran/file.f90
        }
    }
    ```
-5. Map the `Violation` struct to a rule code in `fortitude/src/rules/mod.rs`.
+6. Map the `Violation` struct to a rule code in `fortitude/src/rules/mod.rs`.
     - `code_to_rule` is never called directly, but the match statement within is
       analysed by the macro `fortitude_macros::map_codes` to define a `Rule` enum
       and many associated utilities.
--- a/crates/fortitude/src/check.rs
+++ b/crates/fortitude/src/check.rs
@@ -137,7 +137,8 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
     let (cli, config_arguments) = args.partition(global_options)?;
 
     // Construct the "default" settings. These are used when no `fortitude.toml`
-    // files are present, or files are injected from outside of the hierarchy.
+    // files are present alongside source files, or files are injected from
+    // outside of the hierarchy.
     let file_configuration = resolve::resolve(&config_arguments, cli.stdin_filename.as_deref())?;
 
     let stdin_filename = cli.stdin_filename;
--- a/crates/fortitude/src/resolve.rs
+++ b/crates/fortitude/src/resolve.rs
@@ -58,10 +58,10 @@ pub fn resolve(
     }
 
     // Third priority: find a `fortitude.toml` file in either an ancestor of
-    // `stdin_filename` (if set) or the current working path all paths relative to
-    // that directory. (With `Strategy::Hierarchical`, we'll end up finding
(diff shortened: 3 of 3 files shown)


crates/fortitude_workspace/src/resolver.rs
```diff
@@ -188,3 +287,689 @@ pub fn resolve_root_settings(
...
+#[derive(Debug, Copy, Clone)]
+pub enum ExclusionKind {
+    /// The exclusion came from the `exclude` setting.
+    Exclude,
+    /// The exclusion came from the `extend-exclude` setting.
+    ExtendExclude,
+    /// The exclusion came from the `lint.exclude` setting.
+    LintExclude,
+    /// The exclusion came from the `lint.extend-exclude` setting.
+    FormatExclude,
```
LiamPattinson, line 691: I think the comments and enum options here are incorrect. Ruff's `lint` settings are under `check` for us, and we don't have a `format` section yet.

Should it be that we eventually have `CheckExclude`, `CheckExtendExclude`, `FormatExclude`, and `FormatExtendExclude`?

(I've noticed that Ruff has this exact same 'error', and I'm not sure what's going on with it)

ZedThree: Ah, indeed. The enum names here are correct (for Ruff), there's a top-level `exclude` and `extend-exclude`, and then `lint.exclude` and `format.exclude` which both act like `extend-exclude` for those subcommands specifically.

We currently only have `check.exclude` and `check.extend-exclude`, but when we get the format command in, we'll want to move those to the top-level and add the subcommand-specific options.

For now, we can remove `LintExclude` and `FormatExclude`

ZedThree:
> Something I don't understand is why `resolve_scoped_settings` is called so often. If I'm reading it right, there's a first pass that works out the settings for each path provided, and then it walks down any directories.

Yep, the first pass is for the paths directly provided: `fortitude file1 file2 dir1 dir2`. We resolve the settings for each path by search its ancestors for config files (stopping when we find one).

> For each file in the directories, it then calls `resolve_scoped_settings` for each file, without first checking if that scope has already been cached. I think this is something Ruff does too, so I'd be happy to include it, but it feels like an oversight.

Not quite, we call `resolve_scoped_settings` for each directory rather than each file, but yes it doesn't appear to check if it's already been resolved, good catch! This is Ruff's behaviour too. I'll see how easy it is to fix, and if it is, I'll make a PR upstream too.

PR #538, opened 2025-11-13
Require only minimal python version

ZedThree:
This matches the version required by ruff, which has the same
build/distribution strategy

### DIFF
--- a/python/fortitude/__main__.py
+++ b/python/fortitude/__main__.py
@@ -3,30 +3,91 @@
 Adapted from the ``ruff`` launcher script.
 """
 
-import subprocess
+from __future__ import annotations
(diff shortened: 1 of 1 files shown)


LiamPattinson:
I think the Python `__main__.py` will need to be adjusted to accommodate this, as it uses some features from later Python versions than 3.7.

ZedThree:
Good point! I've just taken Ruff's one (of course). Unfortunately, 3.7 is so old, `uv` doesn't have it available to install

ZedThree:
Yeah, I think it's a combination of missing out on some much nicer recent features and checking a bunch more cases than we were doing

PR #540, opened 2025-11-13
Add `fortitude config` command

ZedThree:
Gives command-line help for the various config options

### DIFF
--- a/crates/fortitude/src/cli.rs
+++ b/crates/fortitude/src/cli.rs
@@ -114,7 +114,9 @@ pub enum SubCommands {
     },
     /// List or describe the available configuration options.
     Config {
-        /// Config key to show
+        /// Config key to show. Running the command with no key will
+        /// show the top-level options. Nested options should be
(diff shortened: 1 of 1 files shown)


crates/fortitude/src/cli.rs
```diff
@@ -110,6 +112,18 @@ pub enum SubCommands {
         #[arg(long, value_enum, default_value = "text")]
         output_format: HelpFormat,
     },
+    /// List or describe the available configuration options.
+    Config {
+        /// Config key to show
```
LiamPattinson:
```suggestion
        /// Config key to show. Running the command with no key will 
        /// show the top-level options. Nested options should be
        /// separated with '.', such as 'check.fix'.
```

ZedThree: Good shout. I'll add this manually and regenerate the docs

ZedThree: Oh, this isn't something that appears in the docs, nevermind

PR #542, opened 2025-11-14
Add more capabilities to `fortitude explain`

ZedThree:
- `--summary` to give one line summaries of the output
- `--list-categories` to list the categories
- `--output-format` to get the output as json

I was also thinking of using [termimad](https://github.com/Canop/termimad/tree/main) to render the output, but decided that might be a bit too far

### DIFF
--- a/crates/fortitude/src/explain.rs
+++ b/crates/fortitude/src/explain.rs
@@ -187,6 +187,12 @@ fn print_rule_explanation(rules: &[Rule]) {
             }
         }
 
+        // Replace links to the docs
+        body = body.replace(
+            "../settings.md",
+            "https://fortitude.readthedocs.io/en/stable/settings",
+        );
+
         let code = rule.noqa_code().to_string();
         let name = rule.as_ref();
         let title = format!("# {code}: {name}\n");


crates/fortitude/src/explain.rs
```diff
@@ -77,14 +190,82 @@ pub fn explain(args: ExplainCommand) -> Result<ExitCode> {
...
     for (code, desc) in outputs {
-        println!("{code}\n{desc}");
+        let string = format!("{code}\n{desc}");
+        if let Some(skin) = &skin {
+            skin.print_text(&string);
+        } else {
+            println!("{}", string);
+        }
+    }
+}
```
LiamPattinson: I found that using markdown rendering, there would be no space between rules when printing multiple at once. This tweak gave me some nicer results:
```suggestion
    for (idx, (code, desc)) in outputs.iter().enumerate() {
        let string = format!("{code}\n{desc}\n");
        if let Some(skin) = &skin {
            skin.print_text(&string);
            if idx < outputs.len() - 1 {
                skin.print_text("---");
            }
        } else {
            print!("{}", string);
        }
    }
```
I couldn't include the top line of the for loop in the suggestion, so I wouldn't recommend merging this directly!

ZedThree:
Now we're using `termimad`, I had a stab at putting the `--list-category` output into a markdown table, but it doesn't play well with `Colored`, and I wasn't super happy with how it looks with plainer text.

Could also put the `--summary` output into a table, but it again it needs some fiddling to make it look good.

Issue #353, opened 2025-03-07
Add `--target-std` flag


LiamPattinson:
In PR https://github.com/PlasmaFAIR/fortitude/pull/337, the rule `implicit-external-procedures` was moved out of preview and is now on by default. As discussed in https://github.com/PlasmaFAIR/fortitude/pull/337#issuecomment-2690260242, the targeted functionality has been in GCC since about 2015, so most projects should be able to use it by now, but not those pinned to an earlier standard (e.g. with `gfortran -std=f2008`).

An easy solution for this is for users to simply disable the rule, but as more rules are added to the `modernise` category it could become onerous to switch off every single rule targeting a newer standard. A way around this could be to add a `--target-std` flag similar to `--target-version` in ruff.

ZedThree:
Yep, great idea!

imciner2:
Yes, I was actually thinking about this also because there are requests for the intrinsic replacements, and that is another point where it would be nice to say I want to target a specific version and not recommend intrinsics from newer standards.

PR #551, opened 2025-11-21
Add `--target-std` option

LiamPattinson:
Resolves https://github.com/PlasmaFAIR/fortitude/issues/353

So far it only turns off `external-implicit-procedures`.

TODO:

- [x] Survey of when certain features were added.
- [x] Update relevant rules.
- [x] Tests.

So far this works like Ruff in that the relevant rules exit early if the standard is set too low/high. It might be cleaner to filter rules before checking.

### DIFF
--- a/crates/fortitude_linter/src/settings.rs
+++ b/crates/fortitude_linter/src/settings.rs
@@ -405,30 +405,18 @@ impl From<GitignoreMode> for bool {
     Debug,
     Hash,
     Default,
+    strum_macros::Display,
     clap::ValueEnum,
 )]
 #[serde(rename_all = "lowercase")]
+#[strum(serialize_all = "lowercase")]
 pub enum FortranStandard {
     F95,
     F2003,
     F2008,
     #[default]
     F2018,
-}
-
-impl fmt::Display for FortranStandard {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "{}",
-            match self {
-                Self::F2018 => "f2018",
-                Self::F2008 => "f2008",
-                Self::F2003 => "f2003",
-                Self::F95 => "f95",
-            }
-        )
-    }
+    F2023,
 }
 
 /// Toggle for progress bar
--- a/crates/fortitude/src/cli.rs
+++ b/crates/fortitude/src/cli.rs
@@ -203,7 +203,7 @@ pub struct CheckCommand {
     pub no_preview: Option<bool>,
 
     /// Set minimum Fortran standard to check files against.
-    /// Options are "f2018" (default), "f2008", "f2003", "f95"
+    /// Options are "f2023", "f2018" (default), "f2008", "f2003", "f95"
     #[arg(long, value_enum)]
     pub target_std: Option<FortranStandard>,
 
--- a/crates/fortitude_workspace/src/options.rs
+++ b/crates/fortitude_workspace/src/options.rs
@@ -131,10 +131,10 @@ pub struct CheckOptions {
     pub preview: Option<bool>,
 
     /// Minimum Fortran standard to check files against.
-    /// Options are "f2018" (default), "f2008", "f2003", and "f95".
+    /// Options are "f2023", "f2018" (default), "f2008", "f2003", and "f95".
     #[option(
         default = "f2018",
-        value_type = r#""f2018" | "f2008" | "f2003" | "f95""#,
+        value_type = r#""f2023" | "f2018" | "f2008" | "f2003" | "f95""#,
         example = r#"
           # Set standard to Fortran 2008
           target-std = "f2008"
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -197,7 +197,7 @@ Options:
       --preview
           Enable preview mode; checks will include unstable rules and fixes. Use `--no-preview` to disable
       --target-std <TARGET_STD>
-          Set minimum Fortran standard to check files against. Options are "f2018" (default), "f2008", "f2003", "f95" [possible values: f95, f2003, f2008, f2018]
+          Set minimum Fortran standard to check files against. Options are "f2023", "f2018" (default), "f2008", "f2003", "f95" [possible values: f95, f2003, f2008, f2018, f2023]
       --progress-bar <PROGRESS_BAR>
           Progress bar settings. Options are "off" (default), "ascii", and "fancy" [possible values: off, fancy, ascii]
       --show-settings
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -585,11 +585,11 @@ Whether to show an enumeration of all fixed lint violations
 <span id="target-std"></span>
 
 Minimum Fortran standard to check files against.
-Options are "f2018" (default), "f2008", "f2003", and "f95".
+Options are "f2023", "f2018" (default), "f2008", "f2003", and "f95".
 
 **Default value**: `f2018`
 
-**Type**: `"f2018" | "f2008" | "f2003" | "f95"`
+**Type**: `"f2023" | "f2018" | "f2008" | "f2003" | "f95"`
 
 **Example usage**:
 


LiamPattinson:
- [x] C001: implicit-typing: {entity} missing 'implicit none'.
  - Since Fortran 90
- [x] C002: interface-implicit-typing: interface '{name}' missing 'implicit none'.
  - I think since Fortran 90?
- [x] C003: implicit-external-procedures: 'implicit none' missing 'external'.
  - Fortran 2018 standard
- [x] C011: missing-default-case: Missing default case may not handle all values.
  - Since Fortran 90
- [x] C051: trailing-backslash: Trailing backslash.
  - N/A
- [x] C061: missing-intent: {entity} argument '{name}' missing 'intent' attribute.
  - Since Fortran 90
- [ ] C071: assumed-size: '{name}' has assumed size.
  - Before Fortran 2003 you couldn't have assumed shape pointers. We should add a check for a `pointer` attribute and exit early for Fortran 95.
- [ ] C072: assumed-size-character-intent: character '{name}' has assumed size but does not have `intent(in)`.
  - This rule recommends `allocatable`, which wasn't allowed for dummy arguments until F2003. I'm not sure how to handle this in F95. Any advice?
- [x] C081: initialisation-in-declaration: '{name}' is initialised in its declaration and has no explicit `save` or `parameter` attribute.
  - I think this is F77
- [x] C082: pointer-initialisation-in-declaration: Pointer '{name}' is initialized in its declaration and has no explicit `save` attribute.
  - Similar
- [x] C091: external-procedure: '{name}' declared as `external`.
  - F77, I think
- [x] C092: procedure-not-in-module: {procedure} not contained within (sub)module or program.
  - F90
- [ ] C101: missing-default-pointer-initalisation: pointer component '{var}' does not have a default initialiser.
  - I think this is a F2003 feature
- [x] C121: use-all: 'use' statement missing 'only' clause.
  - F90
- [ ] C122: missing-intrinsic: 'use' for intrinsic module missing 'intrinsic' modifier.
  - F2003, I think.
- [x] C131: missing-accessibility-statement: module '{}' missing default accessibility statement.
  - F95, I think.
- [x] C141: missing-exit-or-cycle-label: '{name}' statement in named 'do' loop missing label '{label}'.
  - F90
- [x] C143: missing-end-label: '{end_name}' statement in named '{start_name}' block missing label.
  - F77
- [x] C151: misleading-inline-if-semicolon: Semicolon following inline if-statement is misleading.
  - F90
- [x] C152: misleading-inline-if-continuation: Line continuation in inline if-statement is misleading.
  - F90
- [x] C161: nonportable-shortcircuit-inquiry: variable inquiry `{function}({arg})` and use in same logical expression.
  - F77
- [x] C181: unchecked-stat: {stat} argument '{name}' is {check_status} in this scope.
  - F90
- [x] C182: multiple-allocations-with-stat: 'stat' parameter used with multiple {allocations}.
  - F90
- [x] E000: io-error: {message}.
  - N/A
- [x] E001: syntax-error: Syntax error.
  - N/A
- [x] E011: invalid-character: Invalid character '{character}'.
  - N/A
- [x] FORT001: invalid-rule-code-or-name: Unknown rule or code `{rule}` in allow comment.
  - N/A
- [x] FORT002: unused-allow-comment: Unused rule `{rule}` in allow comment.
  - N/A
- [x] FORT003: redirected-allow-comment: `{original}` has been redirected to '{redirect}'.
  - N/A
- [x] FORT004: duplicated-allow-comment: Duplicated rule `{rule}` in allow comment.
  - N/A
- [x] FORT005: disabled-allow-comment: Disabled rule `{rule}` in allow comment.
  - N/A
- [x] MOD011: old-style-array-literal: Array literal uses old-style syntax: prefer `[...]`.
  - F90
- [x] MOD021: deprecated-relational-operator: deprecated relational operator '{symbol}', prefer '{new_symbol}' instead.
  - F90
- [x] OB001: statement-function: statement functions are obsolescent, prefer internal functions. Rule has been removed.
  - Obsolecent since F95, but rule is inactive anyway
- [ ] OB011: common-block: common blocks are obsolescent, prefer modules or derived types.
  - Obsolescent only since F2018
- [ ] OB021: entry-statement: entry statements are obsolescent, use module procedures with generic interface.
  - Since F2008
- [ ] OB031: specific-name: deprecated type-specific function '{func}'.
  - Since F2018
- [x] OB041: computed-go-to: computed go to statements are obsolescent, use a select case statement.
  - Since F95
- [x] OB051: pause-statement: `pause` statements are a deleted feature.
  - Since F95
- [x] OB061: deprecated-character-syntax: '{original}' uses deprecated syntax.
  - Since F95
- [ ] OB201: deprecated-mpi-include: mpif.h include is deprecated, use mpi or mpi_f08 module instead.
  - Should recommend mpi_f08 when at a standard after F2008, or mpi before
- [x] OB211: deprecated-omp-include: omp_lib.h include is deprecated, use the omp_lib module instead.
  - Probably fine for all standards? Older compilers won't have it, but people using them can
    just switch off the rule.
- [x] PORT011: literal-kind: {dtype} kind set with number literal '{literal}'.
  - N/A
- [x] PORT012: literal-kind-suffix: '{literal}' has literal kind suffix '{suffix}'.
  - N/A
- [x] PORT021: star-kind: '{dtype}{size}' uses non-standard syntax.
  - N/A
- [x] PORT031: invalid-tab: Invalid tab character.
  - N/A
- [x] S001: line-too-long: line length of {actual_length}, exceeds maximum {max_length}.
  - N/A
- [x] S061: unnamed-end-statement: end statement should be named.
  - N/A
- [x] S071: missing-double-colon: variable declaration missing '::'.
  - N/A
- [x] S081: superfluous-semicolon: unnecessary semicolon.
  - N/A
- [x] S091: non-standard-file-extension: file extension should be '.f90' or '.F90'.
  - N/A
- [x] S101: trailing-whitespace: trailing whitespace.
  - N/A
- [x] S231: keywords-missing-space: Missing space in '{keywords}'.
  - N/A
- [x] S232: keyword-has-whitespace: Whitespace included in '{keywords}'.
  - N/A
- [x] S241: bad-quote-string: String uses single quotes but double quotes preferred.
  - N/A

crates/fortitude_linter/src/rules/obsolescent/common_blocks.rs
```diff
@@ -60,6 +60,9 @@ use tree_sitter::Node;
 /// end subroutine s
 /// ```
 ///
+/// Common blocks were officially declared obsolescent in Fortran 2018,
+/// so this rule only triggers if the target standard is Fortran 2018 or later.
```
ZedThree: While this is _true_, common blocks have been discouraged since F90 with the introduction of modules, so I'm not very keen on gating on F2018!

crates/fortitude_linter/src/rules/obsolescent/entry_statement.rs
```diff
@@ -17,6 +17,10 @@ use tree_sitter::Node;
 /// Multiple entry procedures can be replaced with modules to share data, and
 /// private module procedures to reuse code.
 ///
+/// ## Notes
+/// Entry statements were officially declared obsolescent in Fortran 2008, so
+/// this rule only triggers if the target standard is Fortran 2008 or later.
```
ZedThree: Again, although `entry` was _officially_ deprecated in F2008, they are still bad practice even in earlier standards

crates/fortitude_linter/src/rules/obsolescent/specific_names.rs
```diff
@@ -57,13 +57,28 @@ fn map_specific_intrinsic_functions(name: &str) -> Option<&'static str> {
...
+/// ! Not recommended
+/// result = ALOG(x)
+///
+/// ! Better
+/// result = LOG(x)
+/// ```
+///
+/// ## Notes
+/// Specific names were officially declared obsolescent in Fortran 2018, so
+/// this rule only triggers if the target standard is Fortran 2018 or later.
```
ZedThree: The generic names are possible in F90, so I don't think we should gate these, especially not as late as F2018

crates/fortitude_linter/src/settings.rs
```diff
@@ -389,6 +392,45 @@ impl From<GitignoreMode> for bool {
...
+    Default,
+    clap::ValueEnum,
+)]
+#[serde(rename_all = "lowercase")]
+pub enum FortranStandard {
+    F95,
+    F2003,
+    F2008,
+    #[default]
+    F2018,
```
ZedThree, line 418: We should also add `F2023`!

LiamPattinson: Can do, although I can't think of anything that would be gated by it right now (it also doesn't help that my version of gcc doesn't support it yet).

PR #552, opened 2025-11-26
Add rule `missing-newline-at-end-of-file`

LiamPattinson:
Similar to https://docs.astral.sh/ruff/rules/missing-newline-at-end-of-file/

I had half an hour to kill so bashed this out, but in hindsight I should have just copied Ruff's implementation!

https://github.com/astral-sh/ruff/blob/main/crates/ruff_linter/src/rules/pycodestyle/rules/missing_newline_at_end_of_file.rs

Raising as a draft PR for now, I'll make the changes in a little while

### DIFF
--- a/crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__missing-newline-at-end-of-file_S002.f90.snap
+++ b/crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__missing-newline-at-end-of-file_S002.f90.snap
@@ -2,11 +2,11 @@
 source: crates/fortitude_linter/src/rules/style/mod.rs
 expression: diagnostics
 ---
-./resources/test/fixtures/style/S002.f90:2:1: S002 [*] missing newline at end of file
+./resources/test/fixtures/style/S002.f90:2:75: S002 [*] missing newline at end of file
   |
 1 | program p
 2 | end program p ! Missing newline, be careful with tools that might add one!
-  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ S002
+  |                                                                           ^ S002
   |
   = help: Add newline at end of file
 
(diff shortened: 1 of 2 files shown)


crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__missing-newline-at-end-of-file_S002.f90.snap
```diff
@@ -0,0 +1,16 @@
+---
+source: crates/fortitude_linter/src/rules/style/mod.rs
+expression: diagnostics
+---
+./resources/test/fixtures/style/S002.f90:2:1: S002 [*] missing newline at end of file
+  |
+1 | program p
+2 | end program p ! Missing newline, be careful with tools that might add one!
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ S002
```
ZedThree: Might be nicer if it only highlights the last character? This is ruff's behaviour:

```
  |
1 | def fn() -> None:
2 |     pass
  |         ^
  |
```

LiamPattinson: Agreed! I couldn't copy Ruff wholesale as we're missing things like `Locator` and `Stylist`, but I copied what I could, and we now replicate that behaviour.

crates/fortitude_linter/src/rules/style/whitespace.rs
```diff
@@ -51,6 +51,58 @@ impl TrailingWhitespace {
...
+        // Check that the last character is a newline
+        if !text.ends_with(['\r', '\n']) {
+            // Determine if the file is using Windows-style line endings
+            let newline = if text.contains("\r\n") {
+                "\r\n"
+            } else if text.contains('\r') {
+                "\r"
+            } else {
+                "\n"
+            };
```
ZedThree, line 95: This is ruff's `Stylist`, right? We should think about getting that in one of these days!

PR #553, opened 2025-11-27
Add rules for `useless-return` and `superfluous-else-return/cycle/exit/stop`

### LINKED ISSUE #546: `useless-return`
This pattern:

```f90
subroutine foo
  ! do stuff
  return
end subroutine foo
```

the `return` can be safely removed. I think this is very easy to detect: trigger on `return`, and check if the next statement is `end_{subroutine,function}_statement`

ZedThree:
Five rules for the price of two!

Closes #546

Because the `superfluous-else-*` rules are all very similar, I've not implemented them as `AstRule`s, but added some special handling to the AST rule loop so they can all be checked in a single function.

They should also be able to have a (safe?) fix -- I just haven't done that yet because it was non-trivial (and could maybe benefit from the `Stylist` -- which should probably be next on the to-do!)

They could plausibly be two rules (as they have different descriptions), but ruff has them as four separate ones so :shrug:

### DIFF
--- a/crates/fortitude_linter/src/lib.rs
+++ b/crates/fortitude_linter/src/lib.rs
@@ -33,6 +33,7 @@ use rules::error::invalid_character::check_invalid_character;
 use rules::error::syntax_error::SyntaxError;
 use rules::style::file_extensions::NonStandardFileExtension;
 use rules::style::line_length::LineTooLong;
+use rules::style::useless_return::check_superfluous_returns;
 use rules::style::whitespace::{MissingNewlineAtEndOfFile, TrailingWhitespace};
 #[cfg(any(feature = "test-rules", test))]
 use rules::testing::test_rules::{self, TEST_RULES, TestRule};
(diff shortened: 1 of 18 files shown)


crates/fortitude_linter/src/rules/style/superfluous_else_return.rs
LiamPattinson, line 1: I think this was left in by accident, it isn't used anywhere

crates/fortitude_linter/src/lib.rs
```diff
@@ -213,6 +214,18 @@ pub(crate) fn check_path(
...
+            Rule::SuperfluousElseReturn,
+            Rule::SuperfluousElseCycle,
+            Rule::SuperfluousElseExit,
+            Rule::SuperfluousElseStop,
+        ]) && matches!(node.kind(), "keyword_statement" | "stop_statement")
+        {
+            if let Some(violation) = check_superfluous_returns(settings, &node, file) {
+                violations.push(violation);
+            }
+        }
```
LiamPattinson, line 247: Activating any of these rules effectively results in all of them being activated, as there's no filtering based on which rule is active within `check_superfluous_returns`. If you try running `check --select=S252 --preview ./crates/fortitude_linter/resources/tests/fixtures/style/S253.f90`, you get `S253` violations.

ZedThree: I knew there was something I'd forgotten to do!

LiamPattinson (approved):
Thanks, looks good.

PR #556, opened 2025-12-09
Add basic symbol table

ZedThree:
Adds a very basic scoped symbol table. Currently only stores variables, and not all information is available as native types, but it's something to build on.

Unfortunately a large chunk of this PR is just passing it to `AstRule::check`, and most rules don't/can't make use of it yet.

There are a couple of rules it is already very useful for though -- `missing-intent` is the best example, where it about halves the size of the implementation.

### DIFF
--- a/crates/fortitude_linter/src/symbol_table.rs
+++ b/crates/fortitude_linter/src/symbol_table.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, str::FromStr};
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result, anyhow};
 use itertools::Itertools;
 use ruff_text_size::TextRange;
 use strum_macros::{EnumIs, EnumString, IntoStaticStr};
@@ -214,9 +214,11 @@ impl<'a> AttributeKind<'a> {
 
         match attr {
             AttributeKind::Intent(_) => Ok(AttributeKind::Intent(Intent::from_node(value))),
-            AttributeKind::Dimension(_) => {
-                Ok(AttributeKind::Dimension(Dimension::try_from_node(value.child(1).context("expected more than one child for 'dimension'")?)?))
-            }
+            AttributeKind::Dimension(_) => Ok(AttributeKind::Dimension(Dimension::try_from_node(
+                value
+                    .child(1)
+                    .context("expected more than one child for 'dimension'")?,
+            )?)),
             _ => Ok(attr),
         }
     }
@@ -289,7 +291,10 @@ impl<'a> VariableDeclaration<'a> {
             return Err(anyhow!("wrong node type"));
         }
 
-        let type_ = Type::try_from_node(&node.child_by_field_name("type").context("expected type")?, src)?;
+        let type_ = Type::try_from_node(
+            &node.child_by_field_name("type").context("expected type")?,
+            src,
+        )?;
 
         let attributes: Result<Vec<_>> = node
             .children_by_field_name("attribute", &mut node.walk())


crates/fortitude_linter/src/rules/correctness/init_decls.rs
```diff
@@ -81,27 +82,24 @@ impl Violation for InitialisationInDeclaration {
...
-            .find(|parent| parent.kind() == "variable_declaration")?;
-
-        // Init in declaration ok for save and parameter
-        if declaration
-            .children_by_field_name("attribute", &mut declaration.walk())
-            .filter_map(|attr| attr.to_text(src))
-            .any(|attr_name| ["save", "parameter"].contains(&attr_name.to_lowercase().as_str()))
-        {
+        let name = get_name_node_of_declarator(node).to_text(src)?.to_string();
+        let decl = _symbol_table.get(name.as_str())?;
```
LiamPattinson: `_symbol_table` should only have an underscore prefix if it's unused.

crates/fortitude_linter/src/rules/correctness/init_decls.rs
```diff
@@ -208,35 +206,31 @@ impl Violation for PointerInitialisationInDeclaration {
...
-
-        // Init in declaration ok for save (in this rule)
-        if declaration
-            .children_by_field_name("attribute", &mut declaration.walk())
-            .filter_map(|attr| attr.to_text(src))
-            .any(|attr_name| ["save"].contains(&attr_name.to_lowercase().as_str()))
-        {
+        let var = get_name_node_of_declarator(node);
+        let name = var.to_text(src)?.to_string();
+        let decl = _symbol_table.get(name.as_str())?;
```
LiamPattinson: Similarly, remove underscore

crates/fortitude_linter/src/rules/correctness/intent.rs
```diff
@@ -48,93 +50,45 @@ impl Violation for MissingIntent {
...
+        src: &SourceFile,
+        _symbol_table: &SymbolTables,
+    ) -> Option<Vec<Diagnostic>> {
+        let entity = node.parent()?.kind().to_string();
+        Some(
+            node.child_by_field_name("parameters")?
+                .named_children(&mut node.walk())
+                .filter_map(|param| {
+                    // Get variable declaration
+                    _symbol_table.get(param.to_text(src.source_text())?)
```
LiamPattinson: Similarly, remove underscore.

Really liking the improvement to this function though!

crates/fortitude_linter/src/rules/style/strings.rs
```diff
@@ -147,7 +153,9 @@ impl AstRule for AvoidableEscapedQuote {
         _settings: &CheckSettings,
         node: &'a Node,
         src: &'a SourceFile,
+        _symbol_table: &SymbolTables,
     ) -> Option<Vec<Diagnostic>> {
+        let _ = _symbol_table;
```
LiamPattinson: I'm a bit confused by this line

crates/fortitude_linter/src/symbol_table.rs
```diff
@@ -0,0 +1,556 @@
...
+    Texture,
+    Value,
+    Volatile,
+    Unknown,
+}
+
+impl AttributeKind {
+    pub fn from_node(value: &Node, src: &str) -> Self {
+        let first_child = value.child(0).unwrap().to_text(src).unwrap_or("<unknown>");
+        // TODO: handle intent, dimension, codimension properly
```
LiamPattinson: Please could you sketch out what currently isn't handled properly?

crates/fortitude_linter/src/rules/style/strings.rs
ZedThree: Ah, I think that's an lsp auto-fix to ignore the unused variable warning, I probably hit too many buttons and applied two different fixes.

crates/fortitude_linter/src/symbol_table.rs
ZedThree: This currently only handles the `intent` and `dimension` part of the attribute, but not the rest of it, so `intent(inout)` will just become `AttributeKind::Intent`. This is minimally useful, but we need to do a bit more work to convert this properly

LiamPattinson: Oh right, that seems obvious now you've explained it, cheers!

LiamPattinson (approved):

PR #558, opened 2025-12-10
Add `out-of-line-attribute` rule

ZedThree:
Includes #556 

~~Still a WIP while I try to get fixes working. Currently working for `parameter` only:~~

Fixes now working for all attributes, including multiple attributes:

```diff
-    integer :: a, b
-    parameter(a=1, b=2)
+    integer, parameter :: a = 1
+    integer, parameter :: b = 2
-    integer :: x
-    allocatable x
-    dimension x(:, :)
+    integer, allocatable, dimension(:, :) :: x
```

This is an unsafe fix because it might clobber comments and line continuations (potentially putting a line longer than the maximum allowed). Also, whitespace might be a little funky.

This adds some helper functions for making these kinds of edits:

- removing an item from a comma-separated list
- removing variable from a declaration
- adding an attribute to a declaration

These could also be useful for removing unused variables and imports

## What it does
Checks for variable attributes which are specified separately to the
variable declaration.

## Why is this bad?
Using separate attribute specification statements (or "out-of-line
attributes") makes the code harder to read by splitting up the important
information about a variable. Instead, give attributes in-line with the
variable declaration. This way, readers only need to look in one place.

## Example
```f90
integer :: nx
real :: x_grid
parameter (nx = 42)
dimension x_grid(nx)
```

Use instead:
```f90
integer, parameter :: nx = 42
real, dimension(nx) :: x_grid
```


crates/fortitude_linter/src/rules/modernisation/out_of_line_attribute.rs
```diff
@@ -0,0 +1,391 @@
...
+        _settings: &crate::settings::CheckSettings,
+        node: &Node,
+        src: &SourceFile,
+        symbol_table: &SymbolTables,
+    ) -> Option<Vec<Diagnostic>> {
+        let code = src.to_source_code();
+
+        if node.kind() == "parameter_statement" {
+            // Parameter statements have slightly different syntax, and have
+            // different nodes in the AST, so handle separat
```
LiamPattinson: Honestly might not be worth fixing, the meaning is clear regardless!
```suggestion
            // different nodes in the AST, so handle separately
```

crates/fortitude_linter/src/rules/mod.rs
```diff
@@ -118,6 +118,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Modernisation, "011") => (RuleGroup::Stable, Ast, Default, modernisation::old_style_array_literal::OldStyleArrayLiteral),
         (Modernisation, "021") => (RuleGroup::Stable, Ast, Default, modernisation::relational_operators::DeprecatedRelationalOperator),
         (Modernisation, "031") => (RuleGroup::Preview, Ast, Optional, modernisation::include_statement::IncludeStatement),
+        (Modernisation, "041") => (RuleGroup::Preview, Ast, Optional, modernisation::out_of_line_attribute::OutOfLineAttribute),
```
LiamPattinson: I'd argue that this should be part of the default set, I would hope that this would be fairly uncontroversial.

```suggestion
        (Modernisation, "041") => (RuleGroup::Preview, Ast, Default, modernisation::out_of_line_attribute::OutOfLineAttribute),
```

LiamPattinson (approved):

PR #559, opened 2025-12-12
Move AST types to separate module

ZedThree:
Includes #558 

I fixed the lifetime issues that meant I couldn't have the AST types in a separate file to `SymbolTable`

Also adds a couple of traits to reduce duplication and hopefully make edit helper functions easier to make generic


LiamPattinson:
I think this needs a quick rebase

PR #562, opened 2025-12-12
Add rules for declaring arrays

### LINKED ISSUE #43: Prefer `var_name(N)` to `dimension(N)` (or vice versa?)
```f90
real, dimension(N) :: array  ! version 1
real :: array(N)             ! version 2

! Similarly...

real, dimension(:) :: array  ! version 1
real :: array(:)             ! version 2
```

Fortran developers often code as if they're paid per keystroke, so I would assume version 1 to be the preferred method, but the [Fortran-Lang page on learning arrays](https://fortran-lang.org/en/learn/best_practices/arrays/) uses version 2 throughout. Either way, it should be kept consistent throughout a project, and a rule should exist to enforce this. I'd lean towards version 2 as it makes it much easier to stay within line-length limits, especially when you also start writing in kinds, intents, allocatable, pointer, target, contiguous, etc.

### LINKED ISSUE #149: Warn about misleading declarations with dimension
The following code is valid, but I think fortitude should warn about the strange declaration of y.

```
implicit none
integer, parameter :: wp = kind(1.0d0)
real(kind=wp), dimension(3) :: x, y(2)
x = [10.0_wp, 20.0_wp, 30.0_wp]
y = x(:2)
print*,y
end
```
A better way to write it is

```
implicit none
integer, parameter :: wp = kind(1.0d0)
real(kind=wp), dimension(3) :: x
real(kind=wp) :: y(2)
x = [10.0_wp, 20.0_wp, 30.0_wp]
y = x(:2)
print*,y
end
```

ZedThree:
Closes #149
Closes #43 

Adds three rules:
- `inconsistent-array-declaration`, using both `dimension` and `foo(N)` in a declaration
- `mixed-scalar-array-declaration` for `real :: a, b(2)`
- `bad-array-declaration`, for either using or not using `dimension` to declare arrays

~~Still trying to get the fixes for `bad-array-declaration` working properly.~~

~~This will take us up to 100 rules!~~
Oops, I was counting the test rules as well, this will bring us to 89!

Currently, the option `check.inconsistent-dimensions.prefer-attribute` only applies to `bad-array-declaration`, but it could in principle be used for the two other rules' fixes too


crates/fortitude_linter/src/rules/mod.rs
```diff
@@ -155,6 +155,9 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Style, "253") => (RuleGroup::Preview, None, Optional, style::useless_return::SuperfluousElseCycle),
         (Style, "254") => (RuleGroup::Preview, None, Optional, style::useless_return::SuperfluousElseExit),
         (Style, "255") => (RuleGroup::Preview, None, Optional, style::useless_return::SuperfluousElseStop),
+        (Style, "261") => (RuleGroup::Preview, None, Optional, style::inconsistent_dimension::InconsistentArrayDeclaration),
```
LiamPattinson: Could this one be on by default? I'd consider it to be very poor style, bordering on a 'correctness' issue.

crates/fortitude_linter/src/rules/style/inconsistent_dimension.rs
```diff
@@ -0,0 +1,393 @@
...
+}
+
+pub mod settings {
+    use crate::display_settings;
+    use ruff_macros::CacheKey;
+    use std::fmt::Display;
+
+    #[derive(Debug, Clone, Default, CacheKey)]
+    pub struct Settings {
+        pub prefer_attribute: bool,
```
LiamPattinson: As we discussed that this rule might be a little too opinionated to appear when the user asks for `style`/`ALL`, would it be possible to instead turn this into a ternary choice of `Some(true)`, `Some(false)`, and `None`, with `None` meaning 'either'? That way, the rule wouldn't do anything unless the user sets this up in their settings file.

If that won't work well with the TOML file, maybe string options of 'always', 'never', 'sometimes'/'either'/'any'?

crates/fortitude_linter/src/rules/style/mod.rs
```diff
@@ -131,4 +134,28 @@ mod tests {
         assert_snapshot!(snapshot, diagnostics);
         Ok(())
     }
+
+    // #[test_case(Rule::InconsistentArrayDeclaration, Path::new("S261.f90"), true)]
+    // #[test_case(Rule::InconsistentArrayDeclaration, Path::new("S261.f90"), false)]
+    // #[test_case(Rule::MixedScalarArrayDeclaration, Path::new("S262.f90"), true)]
+    // #[test_case(Rule::MixedScalarArrayDeclaration, Path::new("S262.f90"), false)]
+    #[test_case(Rule::BadArrayDeclaration, Path::new("S263.f90"), true)]
+    #[test_case(Rule::BadArrayDeclaration, Path::new("S263.f90"), false)]
```
LiamPattinson: Should these comments be removed?

ZedThree: Oh, I had some initial attempt at getting these other rules to also work with `prefer_attribute`. Still not sure if they should or not. This might be another reason to have the setting be a ternary, `always`/`never`/`keep`?

crates/fortitude_linter/src/rules/mod.rs
ZedThree: Yes, I wasn't quite sure where to put these ones. I think they are (especially this one) "readability" rules, which given the current categories, feels like it comes under `style` more than `correctness`, even if this is quite a strong style rule.

On by default is a good idea though, wherever we put it

crates/fortitude_linter/src/rules/style/inconsistent_dimension.rs
ZedThree: Sure, I'll add an enum so it's nicely typed, with `always`/`never`/`keep`, and making `bad-array-declaration` only do anything if it's not `keep`

crates/fortitude_linter/src/rules/mod.rs
LiamPattinson: I'd be happy to keep it under style.

crates/fortitude_linter/src/rules/style/mod.rs
LiamPattinson: Given S263 will correct any fixes from S261 and S262 on a subsequent pass, I don't think it's a high priority for those rules to get it right the first time, but it would be nice if it isn't too much work.

LiamPattinson (approved):
Thanks for all of the changes, I think the ternary enum is a really elegant solution! I just switched S261 to default, and I'm now happy to merge.

Issue #547, opened 2025-11-21
`unreachable-statement`


LiamPattinson:
Similar to https://github.com/PlasmaFAIR/fortitude/issues/546, if a statement immediately after a return isn't something like `end if` or `end do`, raise an error:

```f90
! Okay
subroutine s1(x)
  integer, intent(inout) :: x
  if (x == 100) then
    return
  end if
  x = x + 1
end subroutine s1

! Also okay
subroutine s2(x)
  integer, intent(inout) :: x
  if (x == 100) return
  x = x + 1
end subroutine s2

! Not okay
subroutine s3(x)
  integer, intent(inout) :: x
  x = x + 1
  return
  print *, x
end subroutine s3
```

PR #563, opened 2025-12-15
Add rule `unreachable-statement`

LiamPattinson:
Resolves https://github.com/PlasmaFAIR/fortitude/issues/547

Doesn't handle a few edge cases such as `return` at the end of `block`/`associate`, and it isn't smart enough to figure out something like the following:

```f90
do i=1, 10
  return
end do
print *, "do something"
```

I think a full branch analysis would be needed to do this in full.

### DIFF
--- a/crates/fortitude_linter/src/rules/correctness/unreachable_statement.rs
+++ b/crates/fortitude_linter/src/rules/correctness/unreachable_statement.rs
@@ -1,5 +1,4 @@
-use crate::ast::FortitudeNode;
-use crate::rules::style::useless_return::{BlockExit, next_non_comment_sibling};
+use crate::ast::{FortitudeNode, types::BlockExit};
 use crate::settings::CheckSettings;
 use crate::symbol_table::SymbolTables;
 use crate::{AstRule, FromAstNode};
@@ -77,7 +76,7 @@ impl AstRule for UnreachableStatement {
         // This will require going up the tree before finding the next statement.
         let text = node.child(0)?.to_text(src.source_text())?;
         let _ = BlockExit::try_from(text).ok()?;
-        let sibling = next_non_comment_sibling(node)?;
+        let sibling = node.next_non_comment_sibling()?;
         if EXECUTABLE_STATEMENTS.contains(&sibling.kind()) {
             some_vec!(Diagnostic::from_node(
                 UnreachableStatement {
--- a/crates/fortitude_linter/src/rules/style/useless_return.rs
+++ b/crates/fortitude_linter/src/rules/style/useless_return.rs
@@ -1,4 +1,4 @@
-use crate::ast::FortitudeNode;
+use crate::ast::{FortitudeNode, types::BlockExit};
 use crate::settings::CheckSettings;
 use crate::symbol_table::SymbolTables;
 use crate::{AstRule, FromAstNode, Rule};
@@ -6,32 +6,6 @@ use ruff_diagnostics::{AlwaysFixableViolation, Diagnostic, Fix, Violation};
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 use ruff_source_file::SourceFile;
 use tree_sitter::Node;
-
-/// Get the next sibling that isn't a comment
-pub(crate) fn next_non_comment_sibling<'a>(node: &Node<'a>) -> Option<Node<'a>> {
-    let mut sibling = node.next_named_sibling();
-    while let Some(next_sibling) = sibling {
-        if next_sibling.kind() != "comment" {
-            return Some(next_sibling);
-        }
-        sibling = next_sibling.next_named_sibling();
-    }
-    None
-}
-
-/// Get the next statement (either the next sibling, or the next sibling of
-/// the parent) which isn't a comment
-fn next_non_comment_statement<'a>(node: &'a Node) -> Option<Node<'a>> {
-    if let Some(next) = next_non_comment_sibling(node) {
-        return Some(next);
-    }
-    for node in node.ancestors() {
-        if let Some(next) = next_non_comment_sibling(&node) {
-            return Some(next);
-        }
-    }
-    None
-}
 
 /// ## What it does
 /// Checks for unnecessary `return` statements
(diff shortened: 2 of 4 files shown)


crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__unreachable-statement_C191.f90.snap
```diff
@@ -0,0 +1,52 @@
+---
+source: crates/fortitude_linter/src/rules/correctness/mod.rs
+expression: diagnostics
+---
+./resources/test/fixtures/correctness/C191.f90:10:5: C191 code following `return` is unreachable
+   |
+ 8 |     integer :: i
+ 9 |
+10 |     return  ! Error
+   |     ^^^^^^ C191
```
ZedThree: Could we instead put the diagnostic on the unreachable statement, rather than on the `return`? This is what [clang's `-Wunreachable-code`](https://godbolt.org/z/nY3M9YYM1) does:

```
<source>:7:5: warning: code will never be executed [-Wunreachable-code]
    7 |     num += 2;
      |     ^~~
```

Note that it only does it on the first unreachable statement, not on all of them

crates/fortitude_linter/src/rules/correctness/unreachable_statement.rs
```diff
@@ -0,0 +1,86 @@
...
+
+impl AstRule for UnreachableStatement {
+    fn check<'a>(
+        _settings: &CheckSettings,
+        node: &'a Node,
+        src: &'a SourceFile,
+        _symbol_tables: &SymbolTables,
+    ) -> Option<Vec<Diagnostic>> {
+        // TODO: Not catching returns at the end of block, associate, etc.
+        // This will require going up the tree before finding the next statement.
```
ZedThree, line 76: Would it be simpler to instead check that the next statement is an `end_*` statement?

crates/fortitude_linter/src/rules/style/useless_return.rs
```diff
@@ -8,7 +8,7 @@ use ruff_source_file::SourceFile;
 use tree_sitter::Node;
 
 /// Get the next sibling that isn't a comment
-fn next_non_comment_sibling<'a>(node: &Node<'a>) -> Option<Node<'a>> {
+pub(crate) fn next_non_comment_sibling<'a>(node: &Node<'a>) -> Option<Node<'a>> {
```
ZedThree: Maybe move this and `BlockExit` to `ast/types.rs`?

docs/rules/unreachable-statement.md
```diff
@@ -0,0 +1,11 @@
...
+This rule is unstable and in [preview](../preview.md). The `--preview` flag is required for use.
+
+This rule is turned on by default.
+
+## What it does
+Checks for `return`, `exit`, `cycle`, and `stop` statements that result in
+unreachable code.
+
+## Why is this bad?
+Unreachable code can never be executed, and is almost certainly a mistake.
```
ZedThree, line 11: One piece of feedback we had from BttFF was more examples -- could we add one here?

crates/fortitude_linter/src/rules/correctness/unreachable_statement.rs
LiamPattinson: I'd considered doing it this way -- rather than checking 'is the next statement executable?', instead check 'is the next statement the end of an executable block?'. I found an edge case in `internal_procedures`, which should be easy to catch:

```f90
subroutine s()
  return
contains  ! not an 'end_*' node, but behaves like one
  subroutine s2()
    ...
  end subroutine s2
end subroutine s
```

Another edge case that occurred to me was something cursed like:

```f90
subroutine s()
  return
  #define X
end subroutine s
```

In the end I decided I'd prefer to go for the safe route that might result in false negatives than the smart route that might result in false positives.

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__unreachable-statement_C191.f90.snap
LiamPattinson: Sure, can do. Do you think it would be okay for redundant violations to be raised in the case that a return/exit/etc appears within an unreachable block? E.g.

```f90
subroutine s()
  return
  x = 2  ! Raises violation here
  return
  y = 5  ! Also raises violation here, but ideally shouldn't
end subroutine s
```

docs/rules/unreachable-statement.md
LiamPattinson: Sure, will do.

LiamPattinson: Resolved in https://github.com/PlasmaFAIR/fortitude/pull/563/commits/fddb1935549fa7d8b4808fb9d34521a949ace063

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__unreachable-statement_C191.f90.snap
ZedThree: Oh, yeah, that case is fine -- if you do weird things, you should expect weird results!

LiamPattinson: Resolved in https://github.com/PlasmaFAIR/fortitude/pull/563/commits/46622090066d311baf6b2ec0dfad7269a9bbb3c7

crates/fortitude_linter/src/rules/style/useless_return.rs
LiamPattinson: Happy to do so. I've just approved https://github.com/PlasmaFAIR/fortitude/pull/559, so to avoid a messy merge conflict I'll wait until that's in and I've rebased on `main`.

LiamPattinson: Done in https://github.com/PlasmaFAIR/fortitude/pull/563/commits/8de57727c51623680da1ac58791c9c7f4068fcb3, and I moved the `next_non_comment_*` functions into `FortitudeNode` while I was at it.

PR #585, opened 2026-02-02
[server] Find config for files outside of open workspaces

### LINKED ISSUE #565: Hierarchical directory structure not picked up by server
Mentioned at the end of https://github.com/PlasmaFAIR/fortitude/pull/509

I have this directory structure:

```
---+---dir_0
   |   |
   |   +---fortitude.toml
   |   |
   |   +---test_0.f90
   |   |
   |   +---dir_0_0
   |   |   |
   |   |   +---test_0_0.f90
   |   |
   |   +---dir_0_1
   |       |
   |       +---fortitude.toml
   |       |
   |       +---test_0_1.f90
   |
   +---dir_1
       |
       +---fortitude.toml
       |
       +---test_1.f90
```

The `fortitude.toml` in `dir_0` looks like this:

```toml
[check]
select = ["E", "C", "S"]
preview = true
```

While those in `dir_0_1` and `dir_1` look like:

```toml
[check]
select = ["E", "S"]
preview = false
```

All `test_*.f90` look like:

```f90
program p
end program p;
```

I expect to see the lint warnings `program missing implicit none` and `unnecessary semicolon` in `test_0.f90` and `test_0_0.f90`, but not in `test_0_1.f90` or `test_1.f90`, and this is indeed the case when using `fortitude check`. However, when sitting in `dir_0` and opening all files in neovim, I find:

- `test_0.f90`: expects both, reports both
- `test_0_0.f90`: expects both, reports both
- `test_0_1.f90`: expects neither, reports both
- `test_1.f90`: expects neither, reports only `program missing implicit none`

I get the same result if I `cd` into `dir_0_0`, but if I `cd` into `dir_0_1` I find:

- `test_0.f90`: expects both, reports only `program missing implicit none`
- `test_0_0.f90`: expects both, reports only `program missing implicit none`
- `test_0_1.f90`: expects neither, reports neither
- `test_1.f90`: expects neither, reports only `program missing implicit none`

The behaviour seems to be that the server searches up from the current working directory to find the closest `fortitude.toml`, and then applies those settings to everything in child directories from that point, ignoring any other `fortitude.toml` files. When opening a file above that location, the server instead uses default settings and ignores `fortitude.toml` files relative to the file being opened.

ZedThree:
Fixes #565

There's two different routes we can go down for finding settings for such files:
- the file is below an open workspace, so we can add any settings we find to
  that workspace
- the file is not below an open workspace, in which case we can't cache them

Given this filesystem layout:

```
.
├── dir_0
│   ├── fortitude.toml
│   └── test_0.f90
└── dir_1
    ├── fortitude.toml
    └── test_1.f90
```

This fixes:

- cwd at top-level, opening `dir_0/test_0.f90`
- cwd at top-level, opening `dir_0/test_0.f90`, then `../dir_1/test_1.f90` in same session
- cwd in `dir_0`, opening `../dir_1/test_1.f90`
- cwd in `dir_0`, opening `test_0.f90` first, then `../dir_1/test_1.f90` in same session

without negatively affecting non-default workspaces (such as opening the folder
in VS Code or `lsp-mode` in Emacs).

The main downside to this approach is the lack of workspace for files in `dir_1`
-- we can't share already parsed settings.

---

This doesn't work for the situation where we're opening a file in a nested
directory below the default workspace, where we _do_ have a config file:

```
.
├── subdir
│   ├── fortitude.toml
│   └── test_1.f90
├── fortitude.toml
└── test_0.f90
```

Opening `test_0.f90` first (in single file mode), and then opening
`subdir/test_1.f90` still uses the settings from `./fortitude.toml` instead of `subdir/fortitude.toml`

---

This is the fix I've opened a PR for upstream in ruff, but which hasn't been merged there yet. I propose we get this in and release 0.8.0 rather than waiting on ruff

### DIFF
--- a/crates/fortitude_server/src/session/index.rs
+++ b/crates/fortitude_server/src/session/index.rs
@@ -235,7 +235,7 @@ impl Index {
     ) {
         // Check if we have settings for the given url, for example if we're
         // opening a file outside of the existing workspaces.
-        // See https://github.com/astral-sh/fortitude/issues/17944
+        // See https://github.com/astral-sh/ruff/issues/17944
         let has_settings = match self.settings_for_url(&url) {
             None => {
                 tracing::debug!("{url} not in index");


crates/fortitude_server/src/session/index.rs
```diff
@@ -230,7 +227,41 @@ impl Index {
...
-    pub(super) fn open_text_document(&mut self, url: Url, document: TextDocument) {
+    pub(super) fn open_text_document(
+        &mut self,
+        url: Url,
+        document: TextDocument,
+        global: &GlobalClientSettings,
+    ) {
+        // Check if we have settings for the given url, for example if we're
+        // opening a file outside of the existing workspaces.
+        // See https://github.com/astral-sh/fortitude/issues/17944
```
LiamPattinson: Likely a find-and-replace error:
```suggestion
        // See https://github.com/astral-sh/ruff/issues/17944
```

docs/editors/setup.md
```diff
@@ -44,6 +44,8 @@ following to your `init.lua`:
     vim.lsp.config('fortitude', {
       cmd = { "fortitude", "server" },
       filetypes = { "fortran" },
+      root_markers = { 'fpm.toml', 'fortitude.toml', '.fortitude.toml', '.git' },
+      single_file_support = true,
```
LiamPattinson, line 48: I saw some of these settings mentioned in a Ruff issue. Is it the case that these resolve some of the problems already?

ZedThree: They definitely help, but they don't resolve everything (and only apply to neovim, and not other editors!)

PR #588, opened 2026-02-05
Prepare for v0.8.0

LiamPattinson:


CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
 # Changelog
 
+## 0.8.0
```
ZedThree, line 3: Might be worth duplicating the 8-rc1 changelog here as well? Unsure

LiamPattinson: I had a think about about it, and I think it would make sense to do so.

PR #591, opened 2026-02-10
Add ignore-comment-length setting

oyarzun:
Add setting to allow comments to exceed the line length limit


oyarzun:
@ZedThree thanks for your work on the fortitude and taking the time to look at this PR. 

I was aware that the implementation would also ignore the line length if the there was a exclamation mark in a string literal. I figured I would push it up to get some additional eyes on the code (my first time coding in rust). Using the AST makes sense, thanks for the pointer.

I agree, I'll remove the CLI flags.

ZedThree (dismissed):
Thanks @oyarzun! Sorry it's taken so long to get around to reviewing this, but it looks ready to go in.

I'm just going to do two things first:
1. Rename the option and move to a subgroup, so from `check.ignore-comment-length` to `check.line-too-long.ignore-comments`. This keeps it out of the global options, as it's only relevant for this rule.
2. Add a link to this at the end of the `line-too-long` docstring (~L56 in `line_length.rs`)

I'll do that myself rather than hold this up any longer!

ZedThree (approved):

PR #592, opened 2026-02-12
Add rules for obsolescent `equivalence`, `block data`, and `forall`

ZedThree:
Do we have a way of automatically capturing or displaying the standard that these rules apply from?


ZedThree:
I'm not super crazy about the numbering scheme for these rules. Should they actually be `OB011`, `OB012`, `OB013` etc?

ZedThree:
The standard groups `equivalence`, `common` blocks, and `block data` together, so they make sense to group together.

Someone's been helping get non-block labelled `do` working better in the grammar, so we can get that in too. Hopefully even with a fix.

I'm just working on arithmetic `if`, as I believe we can have a fix for that in a lot of cases too.

ZedThree:
There are still at least a couple of variations on the obsolete `do` loops I want to get in, but with the fixes for arithmethic `if` this is already quite big. @LiamPattinson do you want me to break it up? I could put the `if` and `do` rules in their own PRs

PR #603, opened 2026-03-05
Add rule for obsolete `arithmetic-if`

ZedThree:
~~Includes #592~~

This bumps the grammar to get some recent features, including handling non-block (labelled) `do` loops better, which is required for getting good fixes. There seems to be a small side-effect of changing some of the context for one rule :shrug:


crates/fortitude_linter/src/ast.rs
```diff
@@ -218,6 +235,15 @@ impl<'tree1> FortitudeNode<'tree1> for Node<'tree1> {
...
     }
 
+    fn indentation_ignore_stmt_label(&self, source_file: &SourceFile) -> String {
+        let src = source_file.to_source_code();
+        let start_byte = self.start_textsize();
+        let start_index = src.line_index(start_byte);
+        let start_line = src.line_start(start_index);
+        let width = (start_byte - start_line).to_usize();
+        format!("{:width$}", " ")
+    }
```
LiamPattinson, line 245: My reading of this is that it only gives indentation if the node is the first one on the line. Okay if that's the case, but it should be stated in the docs.

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unnamed-end-statement_S061.f90.snap
LiamPattinson, line 1: Yeah, the changes here are very strange. Hopefully nothing to worry about?

crates/fortitude_linter/src/rules/obsolescent/arithmetic_if.rs
```diff
@@ -0,0 +1,430 @@
...
+/// 1, 2, or 3 blocks
+///
+/// More precisely, we can have:
+/// - 1 block: `if` -- 2 labels
+/// - 2 blocks: `if/else` -- 2 labels
+/// - 2 blocks: `if/else if` -- 3 labels
+/// - 3 blocks: `if/else if/else` -- 3 labels
+///
+/// We get an `else` block if the (n-1)th block has a `goto` to a later sibling node.
+/// In all other cases, we don't need an `else:
```
LiamPattinson, line 144:
```suggestion
/// In all other cases, we don't need an `else`:
```

crates/fortitude_linter/src/rules/obsolescent/arithmetic_if.rs
```diff
@@ -0,0 +1,430 @@
...
+            first.0.prev_non_comment_sibling(),
+        );
+        return None;
+    }
+
+    // TODO: this is too crude, need to just lop off brackets?
+    let condition = node
+        .child_with_name("parenthesized_expression")?
+        .named_child(0)?
+        .to_text(src)?;
```
LiamPattinson, line 222: Not sure I understand the issue with this code. Is it likely to be error prone?

LiamPattinson (commented):
Sorry for taking a while to get around to this one! I only spotted minor issues, but to be honest I'm struggling to follow the auto-fix. The included test indicate that it's working, but given the number of early-exit opportunities, I'm guessing it's with the caveat that the arithmetic ifs have to be structured in a very particular way. It's hard for me to judge how robust it'll be when faced with the horrors of real-world Fortran, but I'd be happy to leave it as-is and fix things if/when any bug reports roll in.

crates/fortitude_linter/src/rules/obsolescent/arithmetic_if.rs
ZedThree: Err, I don't remember now! Will need to double check against the corpus

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unnamed-end-statement_S061.f90.snap
ZedThree: I think it's fine, it's just a change in the context, although I'm not really sure why

crates/fortitude_linter/src/ast.rs
ZedThree: Good point. It's fine for the purposes I'm using it for here, but ideally this wouldn't be a separate function to `indentation`, but it will need to be pretty clever to handle statement labels.

Maybe just changing the name to something like `node_line_offset` or similar

LiamPattinson:
Besides the merge conflict and some very minor edits to comments, I think this PR is basically finished. Please could you take a look over it before I move on to https://github.com/PlasmaFAIR/fortitude/pull/604?

LiamPattinson (approved):
Though there are still some TODO comments, I'm happy to merge this so I can start having a look at https://github.com/PlasmaFAIR/fortitude/pull/604

PR #604, opened 2026-03-06
Add rules for non-block `do` statements

ZedThree:
Includes #603 

Add rules and fixes for a few different obsolete or deleted non-block `do` features.

The rules have some overlaps, but splitting them up makes applying fixes a lot easier. Unfortunately, there's still a couple of edge cases where we can't apply nice fixes. It will be nice to get in an `unused-statement-label` rule+fix to completely tidy up these obsolete constructs


LiamPattinson (approved):
All looks good! There are some tricky bits of logic in the fixes of `non_block_do.rs` that I haven't got time to comb through, but the tests look pretty convincing.

PR #606, opened 2026-03-06
Add rule `too-many-arguments`

LiamPattinson:
Inspired by https://github.com/PlasmaFAIR/fortitude/pull/602, I thought I'd expand the set of complexity rules a little by adding something easy. This could be easily expanded for `too-many-locals`, `too-many-nested-blocks`, `too-many-variables-in-derived-type`, etc.

Renames the `too-complex` option category to `complexity` so that we can reuse the namespace.

I numbered it `S911`, as I know there are other general complexity measures beyond cyclomatic complexity that could fall under `S90*`.


crates/fortitude_linter/src/rules/style/complexity.rs
```diff
@@ -165,6 +165,112 @@ fn cyclomatic_complexity(node: &Node) -> usize {
...
+/// end subroutine update_position
+/// ```
+///
+/// where `vector` is a derived type defined as:
+/// ```f90
+/// type :: vector
+///  real :: x
+///  real :: y
+///  real :: z
+/// end type vector
```
ZedThree, line 218: Missing closing backticks:

```suggestion
/// end type vector
/// ```
```

crates/fortitude_linter/src/rules/mod.rs
```diff
@@ -158,7 +158,8 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Style, "261") => (RuleGroup::Preview, None, Default, style::inconsistent_dimension::InconsistentArrayDeclaration),
         (Style, "262") => (RuleGroup::Preview, None, Optional, style::inconsistent_dimension::MixedScalarArrayDeclaration),
         (Style, "263") => (RuleGroup::Preview, None, Optional, style::inconsistent_dimension::BadArrayDeclaration),
-        (Style, "901") => (RuleGroup::Preview, Ast, Optional, style::too_complex::TooComplex),
+        (Style, "901") => (RuleGroup::Preview, Ast, Optional, style::complexity::TooComplex),
+        (Style, "911") => (RuleGroup::Preview, Ast, Optional, style::complexity::TooManyArguments),
```
ZedThree: Not `902`?

crates/fortitude_linter/src/rules/style/complexity.rs
```diff
@@ -165,6 +165,112 @@ fn cyclomatic_complexity(node: &Node) -> usize {
...
+
+impl Violation for TooManyArguments {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let Self {
+            actual_args,
+            max_args,
+            procedure_name,
+        } = self;
+        format!("too many arguments in procedure {procedure_name} ({actual_args} > {max_args})")
```
ZedThree: We're not super consistent with our message formats, but maybe this instead:

```suggestion
        format!("Too many arguments in procedure `{procedure_name}` ({actual_args} > {max_args})")
```

ZedThree (changes requested):
Looks good, thanks @LiamPattinson! The docstring needs fixing, and there's a style question too.

Also the numbering -- should this actually be `S902`? I mean, in some ways, I think we should push to prefer the names for individual rules

crates/fortitude_linter/src/rules/mod.rs
LiamPattinson: I was thinking of leaving `90*` free for other general complexity measures, such as [Sonar's 'cognitive complexity'](https://www.sonarsource.com/blog/cognitive-complexity-because-testability-understandability/). I'm not sure how likely we are to implement those however, so I'd also be happy to change it if you think it would be best.

docs/rules/too-many-arguments.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Why is this bad?
+Procedures with many arguments are harder to understand, maintain, call,
+and test. They often indicate that a procedure is doing too much and should
+be refactored into smaller, more focused procedures, or that a derived type
+should be used to group related arguments together.
+
+The equivalent rule in Pylint has a default threshold of 5, but we set it to
+10 for Fortran to account for the fact that Fortran procedures often have
+more arguments than Python functions.
```
cyrilgandon: I understand the rationale for raising the threshold to account for Fortran's verbosity, but I'd argue that 10 is quite high as a default.

Looking at real-world Fortran codebases, procedures with 10+ arguments are almost always a sign of missing derived types or poor separation of concerns, not a language necessity. The fact that Fortran *tends* to have more arguments than Python doesn't mean we should normalise it.

I'd personally keep the default at 5, consistent with Pylint, and let users raise it in their config if their codebase genuinely requires it. That way the rule is actionable out of the box rather than silent on the most problematic cases.

crates/fortitude_linter/src/rules/mod.rs
ZedThree: That looks neat, but I don't see a problem with grouping that under `S9` if we do implement it

docs/rules/too-many-arguments.md
LiamPattinson: That's a fair point, I normally see it being used for numerical physics codes where one equation in a paper gets written as one function in the code, but just because it's common doesn't mean it's a good idea, and there's always `! allow(too-many-arguments)` for the few routines for which it's unavoidable.

I've bumped it down to 5 by default, but it'll be easy to change it in future if feedback shows it's too low.

LiamPattinson:
Something I missed on my first pass: should we have an allowance for `self` in type-bound procedures? I'm not sure how we can test for this in general besides just checking if the first argument is called `self` or something similar.

imciner2:
> Something I missed on my first pass: should we have an allowance for self in type-bound procedures? I'm not sure how we can test for this in general besides just checking if the first argument is called self or something similar.

Does pylint include or exclude that? Python also has to pass self (unlike C++ that doesn't have that, and clang-tidy appears to have this rule disabled by default so there isn't much to learn there), so we could follow their lead here. (I will note that if we go with a argument name-based version, both `self` and `this` should probably be included).

LiamPattinson:
It seems pylint/ruff does subtract `self`/`cls` arguments, but does so with some clever semantic analysis that we can't yet replicate. For now I'll aim for the simple solution of subtracting the first argument if it's `self` or `this`, but ideally it should be checking derived types and seeing if there's a type-bound procedure with the same name.

LiamPattinson:
Another edge case: procedures in interface blocks should never be counted!

ZedThree (approved):
Thanks @LiamPattinson!

PR #607, opened 2026-03-06
Add `Stylist` struct to detect indentation and quote style

ZedThree:
Not actually used anywhere yet, but we can now start to use it to get quotes and indentation used in files


ZedThree:
This should probably also detect capitalisation style too

LiamPattinson (dismissed):
Thanks @ZedThree, all looks good. Happy to get this in now and work on getting it more deeply integrated in a later PR.

LiamPattinson (approved):

Issue #560, opened 2025-12-12
Rule `unsorted-imports`


LiamPattinson:
E.g. https://docs.astral.sh/ruff/rules/unsorted-imports/

Should follow similar rules to isort: Sort alphabetically, prioritising CamelCase and ALLCAPS.

Intrinsics should go at the top.

Will need special handling to split long `use` lines over multiple lines

Some parts could be achieved using the utilities in https://github.com/PlasmaFAIR/fortitude/pull/558

ZedThree:
Done in #608

LiamPattinson:
The existing rule sorts use statements, but doesn't sort the imported entities. We should probably open a new issue for that.

ZedThree:
Oh, good point, I didn't check that!

PR #608, opened 2026-03-10
Add rule S271 `unsorted-uses`

cyrilgandon:
Fixing https://github.com/PlasmaFAIR/fortitude/issues/560

## Summary

This PR implements the S271 rule for sorting `use` statements, following the issue's reference to isort-style ordering.

### What is implemented

- `use` statements are sorted alphabetically by module name (case-insensitive, normalized to lowercase)
- Intrinsic modules (`use, intrinsic ::`) are always sorted to the top
- Statements are grouped by contiguous blocks — blank lines between groups are preserved and each group is sorted independently
- An autofix is provided

### What is out of scope

Multiline `use` statements (with `&` continuations) are not reordered by this rule. A follow-up rule will handle sorting the `only:` import lists, and that rule will be the right place to also tackle line breaking for long statements. (S272: `unsorted-only-statement`)

### Question on CamelCase / ALLCAPS prioritization

The issue mentions *"prioritising CamelCase and ALLCAPS"* in the sort order — I'm not sure I fully understand the intent here. Should `CamelCase` module names sort before lowercase ones? Is this about Fortran derived type conventions, or something else? I've gone with a simple case-insensitive alphabetical sort for now, which felt like the most predictable behavior. Happy to adjust once the expected ordering is clarified.

### DIFF
--- a/crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unsorted-uses_S271.f90.snap
+++ b/crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unsorted-uses_S271.f90.snap
@@ -1,6 +1,7 @@
 ---
 source: crates/fortitude_linter/src/rules/style/mod.rs
 expression: diagnostics
+snapshot_kind: text
 ---
 ./resources/test/fixtures/style/S271.f90:2:2: S271 [*] `use` statements are not sorted
   |
@@ -13,22 +14,19 @@ expression: diagnostics
   = help: Sort `use` statements
 
 ℹ Safe fix
-1  1  | module test_module
-2     |-	use module_z_with_tab, only: fun_z_with_tab
-   2  |+  use, intrinsic :: iso_fortran_env, ONLY: int32
-   3  |+    USE module_a, only: fun_a !! module_a inline comments
-3  4  |   use module_c, only: fun_c3, &
-4  5  |                       fun_c2, & !! fun_c2 comments
-5  6  |                       fun_c1
-6     |-  use, intrinsic :: iso_fortran_env, ONLY: int32
-7     |-    USE module_a, only: fun_a !! module_a inline comments
-   7  |+  use module_no_only
-8  8  |   USE module_z, only: fun_z
-9     |-  use module_no_only
-   9  |+	use module_z_with_tab, only: fun_z_with_tab
+1 1 | module test_module
+2   |-	use module_z_with_tab, only: fun_z_with_tab
+  2 |+  use, intrinsic :: iso_fortran_env, ONLY: int32
+  3 |+    USE module_a, only: fun_a !! module_a inline comments
+3 4 |   use module_c, only: fun_c3, &
+4 5 |                       fun_c2, & !! fun_c2 comments
+5 6 |                       fun_c1
+6   |-  use, intrinsic :: iso_fortran_env, ONLY: int32
+7   |-    USE module_a, only: fun_a !! module_a inline comments
+  7 |+	use module_z_with_tab, only: fun_z_with_tab
+8 8 |   USE module_z, only: fun_z
+9 9 |   use module_no_only
 10 10 | 
-11 11 |   use module_single, only: fun_single
-12 12 | 
 
 ./resources/test/fixtures/style/S271.f90:13:3: S271 [*] `use` statements are not sorted
    |
@@ -79,27 +77,31 @@ expression: diagnostics
 27 27 | 
 28 28 |     compute_something = helper(x) + util(y) + real(ieee_max(x,y), real64)
 
-./resources/test/fixtures/style/S271.f90:33:5: S271 [*] `use` statements are not sorted
+./resources/test/fixtures/style/S271.f90:32:5: S271 [*] `use` statements are not sorted
    |
-31 |   subroutine test_comments_as_separator()
-32 |     !! fun_c is used for...
-33 |     use module_c, only: fun_c
-   |     ^^^^^^^^^^^^^^^^^^^^^^^^^ S271
-34 |     use module_a, only: fun_a
-35 |     ! fun_b is used for...
+31 |     subroutine test_comments_as_separator()
+32 | /     !! fun_c is used for...
+33 | |     use module_c, only: fun_c
+   | |_____________________________^ S271
+34 |       use module_a, only: fun_a
+35 |       ! fun_b is used for...
    |
    = help: Sort `use` statements
 
 ℹ Safe fix
+29 29 |   end function compute_something
 30 30 | 
 31 31 |   subroutine test_comments_as_separator()
-32 32 |     !! fun_c is used for...
-   33 |+    use module_a, only: fun_a
-33 34 |     use module_c, only: fun_c
-34    |-    use module_a, only: fun_a
-35 35 |     ! fun_b is used for...
-36 36 |     use module_b, only: fun_b
+32    |-    !! fun_c is used for...
+33    |-    use module_c, only: fun_c
+34 32 |     use module_a, only: fun_a
+35 33 |     ! fun_b is used for...
+36 34 |     use module_b, only: fun_b
+   35 |+    !! fun_c is used for...
+   36 |+    use module_c, only: fun_c
 37 37 |   end subroutine test_comments_as_separator
+38 38 | 
+39 39 |   subroutine test_freeform()
 
 ./resources/test/fixtures/style/S271.f90:40:5: S271 [*] `use` statements are not sorted
    |
--- a/crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unsorted-uses_S271.f90.snap
+++ b/crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unsorted-uses_S271.f90.snap
@@ -1,0 +1,79 @@
+---
+source: crates/fortitude_linter/src/rules/style/mod.rs
+expression: diagnostics
+---
+./resources/test/fixtures/style/S271.f90:2:3: S271 [*] `use` statements are not sorted
+  |
+1 |   module test_module
+2 | /   use module_c, only: fun_c3, &
+3 | |                       fun_c2, &
+4 | |                       fun_c1
+  | |____________________________^ S271
+5 |                                   use, intrinsic :: iso_fortran_env, ONLY: int32
+6 |       USE module_a, only: fun_a
+  |
+  = help: Sort `use` statements
+
+ℹ Safe fix
+1 1 | module test_module
+  2 |+                                use, intrinsic :: iso_fortran_env, ONLY: int32
+  3 |+    USE module_a, only: fun_a
+  4 |+					use module_b_with_tab, only: fun_b
+2 5 |   use module_c, only: fun_c3, &
+3 6 |                       fun_c2, &
+4 7 |                       fun_c1
+5   |-                                use, intrinsic :: iso_fortran_env, ONLY: int32
+6   |-    USE module_a, only: fun_a
+7 8 |     use module_no_only
+8   |-					use module_b_with_tab, only: fun_b
+9 9 |     USE module_z_last, only: fun_z
+10 10 | 
+11 11 |   use module_single, only: fun_single
+
+./resources/test/fixtures/style/S271.f90:13:1: S271 [*] `use` statements are not sorted
+   |
+11 |   use module_single, only: fun_single
+12 |
+13 | use, intrinsic :: iso_fortran_env, only: real64
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ S271
+14 |         use aa_non_intrinsic_module_d, only: fun_d
+15 |         use, intrinsic :: iso_c_binding, only: fun_i
+   |
+   = help: Sort `use` statements
+
+ℹ Safe fix
+10 10 | 
+11 11 |   use module_single, only: fun_single
+12 12 | 
+   13 |+        use, intrinsic :: iso_c_binding, only: fun_i
+13 14 | use, intrinsic :: iso_fortran_env, only: real64
+14 15 |         use aa_non_intrinsic_module_d, only: fun_d
+15    |-        use, intrinsic :: iso_c_binding, only: fun_i
+16 16 | 
+17 17 |   implicit none (type, external)
+18 18 | 
+
+./resources/test/fixtures/style/S271.f90:23:5: S271 [*] `use` statements are not sorted
+   |
+22 |   real function compute_something(x, y)
+23 |     use custom_math, only: fun_m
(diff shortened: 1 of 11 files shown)


ZedThree:
I'm not sure about the capitalisation issue myself. I'll check this against the corpus to see if there is some convention that multiple projects use.

LiamPattinson:
Amazing, I was just thinking I should probably get around to this one! It's been on my list for a while.

On the CamelCase/ALLCAPS topic: The idea was just to copy the convention of `ruff`/`isort`, so that if users are using a specific convention for naming derived types or module-scoped parameters, they'd all be grouped together. In hindsight, I'm not sure if this would actually be a good idea, as the case insensitivity of Fortran means we don't know if the choice of case is actually meaningful or not. I've also seen a bunch of scientific codes that will name variables using capitals to match the convention in the equations they're solving (e.g. electron temperature would be `Te`, not `electron_temperature`), so there's no real consistency.

crates/fortitude_linter/src/rules/style/use_statement.rs
```diff
@@ -0,0 +1,227 @@
...
+    let range1 = stmt1.textrange();
+    let range2 = stmt2.textrange();
+
+    // Get the text between the end of stmt1 and start of stmt2
+    let between_range = TextRange::new(range1.end(), range2.start());
+    let between_text = src.slice(between_range);
+
+    // Check if there's only whitespace (no empty lines)
+    !between_text.contains("\n\n") && !between_text.contains("\r\n\r\n")
+}
```
LiamPattinson: I think a more robust solution here would be to use tree-sitter's [`start_position`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.start_position) and [`end_position`](https://docs.rs/tree-sitter/latest/tree_sitter/struct.Node.html#method.end_position), and to check that the row difference is greater than 1.

crates/fortitude_linter/src/rules/style/use_statement.rs
```diff
@@ -0,0 +1,227 @@
...
+
+            // range.start() points to the first significant character of the node, after indentation.
+            // Walk back to find the '\n' preceding this line, then advance one character
+            // to get the true beginning of the line including its indentation.
+            let line_begin = src.source_text()[..range.start().to_usize()]
+                .rfind('\n')
+                .map(|pos| pos + 1)
+                .unwrap_or(0);
+            let line_end = range.end().to_usize();
+            let text = src.source_text()[line_begin..line_end].to_string();
```
LiamPattinson: We have a newly added function to work out indentation, which also shows some Ruff functions that can be used to get line starts/ends

https://github.com/PlasmaFAIR/fortitude/blob/1a576bb2436da638e8c85a5e5d032ce65d6bd56e/crates/fortitude_linter/src/ast.rs#L209-L219

crates/fortitude_linter/src/rules/style/use_statement.rs
```diff
@@ -0,0 +1,227 @@
...
+
+            let module_name = node
+                .module_name(src.source_text())
+                // Fortran is case-insensitive, normalize to lowercase for consistent sorting
+                .map(|s| s.to_lowercase())
+                .unwrap_or_default();
+
+            let is_intrinsic = node
+                .children(&mut node.walk())
+                .any(|child| child.to_text(src.source_text()) == Some("intrinsic"));
```
LiamPattinson: I'm surprised that `tree-sitter-fortran` doesn't have a named node for `intrinsic`. Something we should probably fix!

crates/fortitude_linter/src/rules/style/use_statement.rs
```diff
@@ -0,0 +1,227 @@
...
+    // We need to include the indentation in the replacement range,
+    // so we walk back to the beginning of the line.
+    let block_start = {
+        let start = block.first()?.textrange().start().to_usize();
+        let line_begin = src.source_text()[..start]
+            .rfind('\n')
+            .map(|pos| pos + 1)
+            .unwrap_or(0);
+        ruff_text_size::TextSize::from(line_begin as u32)
+    };
```
LiamPattinson: As before, could use the `indentation` function to extract the indentation.

LiamPattinson (commented):
All looking really good so far! I've left a few comments where tweaks could be made. 

There are also a few edge cases that occurred to me:

- How do we handle comment lines in the middle of a group of `use` statements? Do we assume they're separating two separate blocks, or assume the comment is attached to the following line?
  - I'd lean towards testing block separation by finding the lines between adjacent `use` statements and then checking if any of them contain only whitespace. Comments would then be attached to the following `use` statement and carried around with them.
- How do we handle multiple `use` statements on one line, like `use a; use b; use c`? (besides writing angry emails to the code authors)
  - Perhaps a first pass of the fix could be to remove semicolons using functions similar to those in [`semicolons.rs`](https://github.com/PlasmaFAIR/fortitude/blob/main/crates/fortitude_linter/src/rules/style/semicolons.rs)
- How do we handle C preprocessor statements in the middle of a block of `use` statements?
  - I have no ideas for this one, and I don't know how likely it is to occur in practice.

LiamPattinson:
I think the tests are failing due to mismatching line endings between different operating systems. We had to add something to [`whitespace.rs`](https://github.com/PlasmaFAIR/fortitude/blob/main/crates/fortitude_linter/src/rules/style/whitespace.rs) to handle this, but I'm sure it's come up in several places and I'm forgetting if we ever came to a general solution. It looks like `Stylist` in https://github.com/PlasmaFAIR/fortitude/pull/607 will be of help here, but I haven't gotten around to reviewing it yet.

cyrilgandon:
> * How do we handle comment lines in the middle of a group of `use` statements? Do we assume they're separating two separate blocks, or assume the comment is attached to the following line?
>   
>   * I'd lean towards testing block separation by finding the lines between adjacent `use` statements and then checking if any of them contain only whitespace. Comments would then be attached to the following `use` statement and carried around with them.

Good catch, thanks! I noticed the current implementation actually deletes comments entirely when they appear in the middle of a use block — they fall within the replacement range but are not part of any node in `statements_with_data`, so they just vanish.

imciner2:
> I'd lean towards testing block separation by finding the lines between adjacent use statements and then checking if any of them contain only whitespace. Comments would then be attached to the following use statement and carried around with them.

I think being able to separate blocks via whitespace is crucial, and I think clang-tidy treats white-space only lines as the separator, so I think it makes sense to copy that here.

Thinking about future rules, can the machinery for processing use statements here be used to easily search for duplicated use statements for modules? (e.g., similar to this clang-tidy rule https://clang.llvm.org/extra/clang-tidy/checks/readability/duplicate-include.html)

crates/fortitude_linter/src/rules/style/use_statement.rs
imciner2: Adding the `intrinsic` and `non_intrinsic` parts as named nodes (the standard appears to call them `module-nature` keywords) would also help simplify the intrinsic use rule (https://github.com/PlasmaFAIR/fortitude/blob/1a576bb2436da638e8c85a5e5d032ce65d6bd56e/crates/fortitude_linter/src/rules/correctness/use_statements.rs#L137).

cyrilgandon:
After trying to carry comments along with their associated `use` statement, I ran into way too many corner cases (inline comments, comments at the start of a block, blank lines between comments...) and the code was getting really messy.

Thinking about it more, I think treating a comment line as a block separator is actually the honest and correct behavior. The rule is simple: sort contiguous `use` statements, full stop. If you put comments in the middle of your imports, you're explicitly organizing them into separate groups — and that intent should be respected rather than silently reordered.

Note that the following example:
```f90
    !! fun_c is used for...
    use module_c, only: fun_c
    use module_a, only: fun_a
```
will be sorted to:
```f90
    !! fun_c is used for...
    use module_a, only: fun_a
    use module_c, only: fun_c
```
The comment ends up detached from `module_c` after sorting. I think this is acceptable.

I tried a quick test in Rust.

```rs
/// Contains methods to parse Fortran code into a tree-sitter Tree and utilities to simplify the
/// navigation of a Tree.
use tree_sitter::{Node, TreeCursor};
use ruff_diagnostics::Edit;
```
It gives:
```rs
use ruff_diagnostics::Edit;
/// Contains methods to parse Fortran code into a tree-sitter Tree and utilities to simplify the
/// navigation of a Tree.
use tree_sitter::{Node, TreeCursor};
```
I guess this just means I'm not skilled enough to implement the same behavior here 🙂

cyrilgandon:
> * How do we handle multiple `use` statements on one line, like `use a; use b; use c`? (besides writing angry emails to the code authors)
>   
>   * Perhaps a first pass of the fix could be to remove semicolons using functions similar to those in [`semicolons.rs`](https://github.com/PlasmaFAIR/fortitude/blob/main/crates/fortitude_linter/src/rules/style/semicolons.rs)

For handling multiple `use` statements on the same line, we now filter them so that only the first statement per line is considered. This way, statements like `use a; use b; use c` will only pick `use a` during sorting or linting.  


> * How do we handle C preprocessor statements in the middle of a block of `use` statements?
>   
>   * I have no ideas for this one, and I don't know how likely it is to occur in practice.

Honestly, I don’t have a solution for C preprocessor statements in the middle of `use` blocks either 😅.  
I haven’t seen this happen in practice, so it might not be a common case, but it’s definitely something to keep in mind if it ever shows up.

imciner2:
> I haven’t seen this happen in practice, so it might not be a common case, but it’s definitely something to keep in mind if it ever shows up.

I have seen this in a few codes. One that I have (unable to share unfortunately) has the following in places:

```
#if(MKL==1)
  use LAPACK95, only : gesvd, syev !,dlaqge
  use BLAS95,   only : trmm,gemm
#else
  use F95_LAPACK,  only : gesvd=>LA_GESVD, syev=>LA_SYEV !,dlaqge
  use blas_class,   only : trmm,gemm
#endif
```
```
#if MKL
  use LAPACK95, only : ggevx, geevx, sygvd
#else
  use F95_LAPACK, only : ggevx=>LA_GGEVX,geevx=>LA_GEEVX,sygvd=>LA_SYGVD
#endif
```

ZedThree:
Yep, I know of several codes that have something like:

```f90
#if HAVE_FOO
  use foo
#endif
```

and even the much worse:

```f90
  use foo, only: bar
#if HAVE_ZING
    , zing
#endif
```

but I think currently we can't actually even parse the second one

LiamPattinson:
I think it should be possible to handle the case where `use` statements are within `#if` directives, and I'd argue that they should just be treated as separate blocks. I was thinking something worse, like:

```f90
use mod_a
#define X 3.1415
use mod_b
```

I'm not aware of any other construct that can be interwoven with a block of `use` statements. Maybe they could be handled by defining block separators as any line that isn't a `use` statement or a comment?

imciner2:
Yea, a define in the middle like that could pose an issue if the thing being used is then used further down in the use statements. Probably saying that comments, whitespace and preprocessor statements are block separators is the best option here.

cyrilgandon:
Two follow-up rules would make sense to add after this one.

1.  ~~Disallow inline `use` statements~~

 ~~Something like `no-inline-use` or `no-freeform-use`. This rule would forbid writing multiple `use` statements on a single line separated by semicolons~~
_Already done in S082 (multiple-statements-per-line)_

2. **Sort `only` lists**

A second rule could enforce sorting of imports inside `use ... , only:` clauses.

Example:

```fortran
use my_module, only: z, a, m
```

would become:

```fortran
use my_module, only: a, m, z
```

imciner2:
> Disallow inline use statements

Wouldn't this already be caught by S082 (multiple-statements-per-line)?

LiamPattinson:
> Wouldn't this already be caught by S082 (multiple-statements-per-line)?

Yes, I don't think we need a specific rule for this.

A separate rule to sort the list of things included with `only` is a good idea though, both from an implementation standpoint and to give users more flexibility. Ruff only implements `unsorted-imports`, which is essentially the entirety of the `isort` tool running in `black` mode. I think splitting the rule in two is a sound approach.

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__unsorted-uses_S271.f90.snap
```diff
@@ -0,0 +1,125 @@
...
+source: crates/fortitude_linter/src/rules/style/mod.rs
+expression: diagnostics
+---
+./resources/test/fixtures/style/S271.f90:2:2: S271 [*] `use` statements are not sorted
+  |
+1 | module test_module
+2 |     use module_z_with_tab, only: fun_z_with_tab
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ S271
+3 |   use module_c, only: fun_c3, &
+4 |                       fun_c2, & !! fun_c2 comments
```
ZedThree: Maybe it would be better to highlight the whole unsorted blocked, rather than just the first statement?

crates/fortitude_linter/src/rules/style/use_statement.rs
```diff
@@ -0,0 +1,366 @@
...
+/// use module_b, only: fun_b
+///
+/// ! Better
+/// use, intrinsic :: iso_fortran_env, only: int32
+/// use module_a, only: fun_a
+/// use module_b, only: fun_b
+/// use module_c, only: fun_c
+/// ```
+///
+/// Blocks of `use` statements separated by blank lines are sorted independently.
```
ZedThree: Maybe move this to the `Why is this bad` section?

ZedThree (changes requested):
Looks great, thanks @cyrilgandon! Just a couple of pretty minor things.

I do think it would be better if we could keep comments attached to the `use` statements -- I'll have an investigate to see how that can be done

ZedThree:
I've pushed a branch [`comment-blocks`](https://github.com/cyrilgandon/fortitude/compare/feat/unsorted-use-statement...PlasmaFAIR:fortitude:comment-blocks) which changes the behaviour to keep blocks of comments attached to the `use` statement (and also highlight the whole unsorted block)

@cyrilgandon @LiamPattinson Thoughts?

LiamPattinson:
Thanks @ZedThree, that's looking pretty good to me! I was going to suggest merging as-is and adding a similar fix afterwards, but it would be even better if we could get your work merged into this branch first.

ZedThree:
Ok, I've pushed straight to this branch, fixing the conflicts and my other comment too

ZedThree (approved):
Thanks @cyrilgandon!

ZedThree:
Ok from my side, but maybe someone else wants to double check my changes?

LiamPattinson (approved):
This is all looking good to me!

I think this rule will require some stress testing to ensure we're hitting all the edge cases, and I'd like to handle the case of semi-colon separated `use` statements within this rule if possible. For now, I'm happy to merge this as-is and play around with it later.

Thanks @cyrilgandon, this is a great addition!

cyrilgandon:
> Ok, I've pushed straight to this branch, fixing the conflicts and my other comment too

Thank you ZedThree, awesome job on the CommentBlock structure. I'm sure it will be useful for other rules too.

PR #610, opened 2026-03-16
Added a pre-commit hook definition for fortitude

wiltonloch:
This allows fortitude to be easily invoked in pre-commit runs of other projects.

I added it using the python installation because the rust cargo install shows problems when the pre-commit infrastructure tries to call it.


ZedThree:
Thanks @wiltonloch, but I think this is already covered by [fortitude-pre-commit](https://github.com/PlasmaFAIR/fortitude-pre-commit)?

wiltonloch:
Aah true, sorry! I did not check that in the README file. Thanks for pointing it out! Feel free to close the PR.

PR #611, opened 2026-03-18
Add rule S233 `incorrect_keyword_case`

### LINKED ISSUE #456: Keywords should be lower/upper/title(?) case
We should add a style rule that performs the following conversions:

```f90
! Prefer lowercase
PURE INTEGER FUNCTION   -->   pure integer function
! Prefer uppercase
pure integer function   -->   PURE INTEGER FUNCTION
! Prefer titlecase (I know some style guides use this, but it's rare)
pure integer function   -->   Pure Integer Function
``` 

I'm not sure if should support title case, but I'd be open to it. It probably wouldn't be much more work.

The biggest challenge here will be finding every keyword and the corresponding TreeSitter node!

cyrilgandon:
Fixes https://github.com/PlasmaFAIR/fortitude/issues/456

This PR introduces a new **`IncorrectKeywordCase`** rule, aimed at enforcing consistent keyword casing across Fortran code:

* The rule **targets only leaf nodes** in the AST, ensuring that keywords like `module`, `implicit`, `end`, `type`, etc., are corrected **without touching associated names** (`my_module`, `BASE_TYPE`, etc.).
* Multi-token statements such as `implicit none`, `end type BASE_TYPE`, and `module MY_MODULE` are handled correctly.
* Supports user-configurable **keyword casing** through the new `KeywordCase` setting.

Additionally, the `titlecase` implementation has been optimized:

* Removed the regex-based approach in favor of a single-pass implementation

This significantly reduces the performance impact observed when using the `titlecase` option.

### ~~Breaking change~~

* ~~The previously planned `KeywordWhitespaceOptions` has been **renamed to `KeywordOptions`** to integrate the new casing option.~~
* ~~This is a **breaking change**: existing configs using `KeywordWhitespaceOptions` must be updated to `KeywordOptions`.~~


ZedThree:
Sorry, poor timing, we've just merged #607 which has a `Capitalisation` enum and functionality for converting to title case! Also a trait so we can convert `str` easily.

Instead of making a hard breaking change to the config option, it would be better to deprecate the old one, and have the new name default to reading the old one (you can look how ruff handles this, e.g. `extend-ignore`).

Slightly orthogonal to this PR, we'll also need a config setting for how we capitalise newly inserted keywords, and that might actually want an additional `auto` setting for taking it from the file (using `Stylist`). I'm not sure how that higher level setting should relate to specific rules' settings, including this one -- `invalid-tab.indent-width` for example might be used in more places.

cyrilgandon:
Thanks, that actually works out great.

Even better if I can reuse the existing `Capitalisation` enum from #607 instead of duplicating logic.

I’ll switch to a deprecation approach for the config option rather than introducing a breaking change.

For the  `auto` mode, I assume this rule would follow the same behavior, i.e. detecting casing per file and adapting accordingly?

ZedThree:
> For the `auto` mode, I assume this rule would follow the same behavior, i.e. detecting casing per file and adapting accordingly?

Yes, exactly. This makes sense for other rules in general, but maybe it doesn't make complete sense for this rule in particular. What do you think?

cyrilgandon:
I see the idea, but I’m not entirely convinced it makes sense for this rule.

Keyword casing feels more like a global stylistic choice, not something that should vary per file. Introducing an auto mode here might be a bit overkill, and it feels like the kind of feature that ends up rarely used in practice.

I think a fixed, explicit configuration is simpler and more predictable for this case.

cyrilgandon:
**Regarding the keyword list — scope question**

While working on this rule, I came across a fairly comprehensive Fortran 2018 keyword list: https://raw.githubusercontent.com/cdslaborg/FortranKeywords/refs/heads/main/FortranKeywords2018.txt

It covers 14 categories: `attribute`, `constant`, `function_elemental`, `function_inquiry`, `function_transformational`, `function_void`, `module_intrinsic`, `specifier`, `statement`, `subroutine`, `subroutine_atomic`, `subroutine_collective`, `subroutine_pure`, `type_derived`.

My current implementation covers `attributes` and `statements`. I'm reasonably confident these are the core targets for a keyword casing rule.

The open question is about intrinsic functions like `COS`, `SIN`, `SQRT`, etc.. Should they be in scope? On one hand, they are technically keywords and their casing is inconsistent in the wild. On the other hand, they behave more like identifiers in practice, and linting their casing feels closer to enforcing a naming convention than a structural syntax rule.

What's your take? Should we include intrinsics, or keep the rule focused on structural keywords only?

LiamPattinson:
I'd lean towards implementing the simplest case for now and extending it later, so just the structural keywords. There are a couple of complications with intrinsic functions:

- Different compilers offer different sets of built-in functions that go beyond the Fortran standard. Would the scope include those too? It could be tricky to get the full list.
- I've seen numerous cases of people reusing intrinsic function names for their own variables (`gamma` pops up a lot for any physics equation with $\gamma$ in it). We'll need to implement some additional functionality before we can test whether a given function call is referencing an intrinsic function or a local one.
  - As an aside, we should have a rule that forbids the reuse of intrinsic function names.

cyrilgandon:
I reconsidered the options structure: rather than deprecating `KeywordWhitespaceOptions`, I created a separate `KeywordCaseOptions`. There didn't seem to be much value in reusing the same settings struct for two fairly distinct concerns (keyword casing vs. spacing in compound keywords).

I also double-checked the keyword list. I kept a few intrinsic-like keywords that feel structural enough to warrant casing enforcement — `allocate`, `read`, `write`, `print`, etc. Happy to be corrected on that.

One concern though: running fortitude against my project is noticeably slow. I suspect the large keyword list might be the culprit, but I haven't looked into how fortitude processes it internally yet. This needs investigation before merge.

Marking as ready for review and removing draft status, but I'll keep looking at the performance issue in the meantime.

cyrilgandon:
Quick update on the performance issue I mentioned earlier:

I tracked it down, and the slowdown comes from using the `titlecase` option. Both `lowercase` and `uppercase` don’t seem to impact performance at all, which is reassuring.

So the issue looks isolated rather than inherent to the keyword list size itself. I’ll dig a bit deeper into why `titlecase` behaves differently, but at least we know the scope now.

ZedThree:
Sorry this has languished, @cyrilgandon! Thanks for optimising the titlecase implementation. Do you think this is ready now?

It would be good to add some variables named like keywords to the test, maybe something like fprettify's test:

```f90
program demo
   integer :: endif, if, elseif
   integer, DIMENSION(2) :: function
   endif = 3; if = 2
   if (endif == 2) then
      endif = 5
      elseif = if + 4*(endif + &
                       2**10)
   elseif (endif == 3) then
      FUNCTION(IF) = ENDIF/ELSEIF
      PRINT *, ENDIF
   endif
end program
```

I've just tried this and it unfortunately has some false positives:

```
/tmp/foo.f90:10:16: S233 [*] Keyword 'IF' should be 'if'
   |
 8 |                        2**10)
 9 |    elseif (endif == 3) then
10 |       FUNCTION(IF) = ENDIF/ELSEIF
   |                ^^ S233
11 |       PRINT *, ENDIF
12 |    endif
   |
   = help: Change to 'if'

/tmp/foo.f90:10:22: S233 [*] Keyword 'ENDIF' should be 'endif'
   |
 8 |                        2**10)
 9 |    elseif (endif == 3) then
10 |       FUNCTION(IF) = ENDIF/ELSEIF
   |                      ^^^^^ S233
11 |       PRINT *, ENDIF
12 |    endif
   |
   = help: Change to 'endif'
```

It looks like it's sufficient to just check to see if the parent node is `identifier`, and if so, skip it:

```diff
modified   crates/fortitude_linter/src/rules/style/keywords.rs
@@ -377,6 +377,10 @@ impl AstRule for IncorrectKeywordCase {
             // Filter node that wrap other nodes, we want only leafs
             return None;
         }
+        if node.parent()?.kind() == "identifier" {
+            // This is actually a variable
+            return None;
+        }
         let text = node.to_text(src.source_text())?;
 
         let expected = text.to_capitalisation(settings.incorrect_keyword_case.keyword_case);
```

I'm also wondering if it might be possible to generate the list of keywords from the grammar somehow... will investigate!

ZedThree:
I managed to generate what I think is a pretty robust list of keywords in the grammar Rust binding: https://github.com/stadelmanma/tree-sitter-fortran/pull/192

This means the entrypoints can just be:

```rust
    fn entrypoints() -> Vec<&'static str> {
        tree_sitter_fortran::KEYWORDS.to_vec()
    }
```

@cyrilgandon Your list of keywords also let me spot that the grammar was missing the `non_recursive` procedure attribute, so I've fixed that, thanks!

cyrilgandon:
Ah yes, good catch on variables named as keywords. I’m so used to other languages forbidding that that I completely forgot Fortran allows it!! I’ll add the test and the fix.

Other than that, I think it’s ready. So I guess we’re waiting for the new `tree_sitter_fortran` release to land first?

Really nice work, by the way, finding a way to avoid maintaining a hardcoded keyword list and relying on a dynamic one instead. Much cleaner.

ZedThree:
We can actually just point straight to the commit for the time being:

```toml
tree-sitter-fortran = { git = "https://github.com/stadelmanma/tree-sitter-fortran.git", rev = "32a3e63" }
```

ZedThree (approved):
LGTM, thanks @cyrilgandon! I do worry slightly that the new option is a bit verbose, and your initial idea of combining it with the whitespace option was sensible -- maybe we revisit this idea in a separate PR?

PR #612, opened 2026-03-18
Add fixes for `superfluous-else-*` rules

ZedThree:
This is failing on Windows even though the final diffs look identical, so I'm guessing there's a line ending issue. Unfortunately, the GitHub raw logs don't seem to preserve the Windows line endings, so I'm a bit stuck. @LiamPattinson do you still have access to a Windows dev environment?


LiamPattinson:
It might take me a while to set it up, the Windows machine I have access to hasn't been booted in several years now.

We have a fix for issues like this in the style rule `missing-newline-at-end-of-file` in `whitespace.rs`. It might just be necessary to pass the newline string into your `redent` function.

ZedThree:
I thought we had something for showing non-printable characters in the snapshots, but I can't find it now.

I'm a bit worried it's doing something like splitting `\r\n` when inserting a change. But I had thought all of our snapshots use Unix line-endings, so I don't understand what's going on here

ZedThree:
Think I've solved the line-ending issue here -- basically GitHub's `action/checkout` defaults to the git setting `core.autocrlf = true`, which automatically converts `\n` line endings to `\r\n` on checkout (and vice versa on commits). This includes both the test `.f90` files _and_ the snapshots, so fixes that touch existing lines for whatever reason have to be careful to preserve the line endings.

As annoying as this is, it's probably actually useful so that we can catch these issues!

I had thought we'd bumped into this before, but I couldn't find it.

Also lucky that we'd merged in the `Stylist` struct so this has actually a very easy fix in practice. I'm constructing the `Stylist` right in the fix helper function, but we should actually do it once per file and pass it into the checker functions (possibly just lumping it in with the `Settings` that we pass in already)

crates/fortitude_linter/src/rules/style/useless_return.rs
```diff
@@ -123,11 +130,18 @@ pub(crate) struct SuperfluousElseReturn {
 }
 
 impl Violation for SuperfluousElseReturn {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
     #[derive_message_formats]
     fn message(&self) -> String {
         let Self { branch } = self;
         format!("Unecessary {branch} after `return` statement")
```
LiamPattinson:
```suggestion
        format!("Unnecessary {branch} after `return` statement")
```

This was missed in the first pass! Unfortunately this will require rebuilding all of the docs/snapshots.

crates/fortitude_linter/src/rules/style/useless_return.rs
```diff
@@ -173,11 +187,18 @@ pub(crate) struct SuperfluousElseCycle {
 }
 
 impl Violation for SuperfluousElseCycle {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
     #[derive_message_formats]
     fn message(&self) -> String {
         let Self { branch } = self;
         format!("Unecessary {branch} after `cycle` statement")
```
LiamPattinson:
```suggestion
        format!("Unnecessary {branch} after `cycle` statement")
```

crates/fortitude_linter/src/rules/style/useless_return.rs
```diff
@@ -223,11 +244,18 @@ pub(crate) struct SuperfluousElseExit {
 }
 
 impl Violation for SuperfluousElseExit {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
     #[derive_message_formats]
     fn message(&self) -> String {
         let Self { branch } = self;
         format!("Unecessary {branch} after `exit` statement")
```
LiamPattinson:
```suggestion
        format!("Unnecessary {branch} after `exit` statement")
```

crates/fortitude_linter/src/rules/style/useless_return.rs
```diff
@@ -270,11 +298,18 @@ pub(crate) struct SuperfluousElseStop {
 }
 
 impl Violation for SuperfluousElseStop {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
     #[derive_message_formats]
     fn message(&self) -> String {
         let Self { branch, stop } = self;
         format!("Unecessary {branch} after `{stop}` statement")
```
LiamPattinson:
```suggestion
        format!("Unnecessary {branch} after `{stop}` statement")
```

LiamPattinson (dismissed):
Looks good, except for a typo...

crates/fortitude_linter/src/rules/style/useless_return.rs
ZedThree: Oops! Thanks, fixed!

LiamPattinson (approved):

PR #613, opened 2026-03-20
Add `--git-staged` and `--git-since` CLI arguments

### LINKED ISSUE #584: Option to only run on modified lines
I'm a developer for a large, under-staffed Earth system model that encourages community contributions. As you might expect, over the decades we've had a lot of mess build up in the codebase, so I would love to use Fortitude to clean it up.

One concern I have is that linting the entire codebase all at once will inevitably generate merge conflicts. In a perfect world, where contributors followed best git practices and maintainers had plenty of time, this wouldn't be too much of an issue. But that's unfortunately not where we find ourselves.

As a way to slowly improve our codebase while avoiding merge conflicts, it would be useful to have an option for Fortitude to ignore lines that git doesn't consider modified. In a pre-commit context, this would mean ignoring unstaged lines.

Unless you have some kind of magic I can't imagine, I don't think it would be possible to implement this as a filter before the check gets run, but Fortitude is so dang fast I think it'd be fine to do the complete check and then just filter the results.

ZedThree:
Closes #584 

~~This is the first part of #584, and lets the user supply line ranges to filter, keeping only warnings that are within the supplied ranges.~~

Needs some docs to clarify exactly how this works.

~~Second part will be something like a `--git-staged` flag which will automatically supply the appropriate line ranges.~~

Now adds `--git-staged-only` as well. Currently only seems to work from the top-level of a repo, but really should work from nested directories too

- [x] tests for git flags
- [x] docs
- [x] `--git-since <tree-ish>`
- [x] fix issues around absolute/relative paths


ZedThree:
Hmm, the Windows build was having problems with the filename as an absolute value due to the `\\` separators, and now the MacOS build is having problems with it as a relative path :facepalm:

crates/fortitude/src/cli.rs
```diff
@@ -328,6 +329,20 @@ pub struct CheckCommand {
...
+    ///    {"name": "file1.f90", "lines": [[6, 7], [42, 45]]},
+    ///    {"name": "file2.f90"}
+    ///  ]
+    /// Line ranges include the end.
+    #[arg(long, help_heading = "File selection", value_name = "LINE_FILTER")]
+    pub line_filter: Option<Filter>,
+
+    /// Only run on files that have been staged in a git repository
+    #[arg(long, help_heading = "File selection")]
+    pub git_staged_only: bool,
```
LiamPattinson: Could this be renamed `--staged-only` for brevity? I imagine there may be rare occasions where this might be applied to non-git repos, but hopefully that would be something we could add support for later.

LiamPattinson (commented):
Looks good! Would it be possible to push a test for `--git-staged-only`/`--staged-only`?

Also looks like the docs need regenerating.

LiamPattinson:
Do you think it would be a good idea for `--git-staged-only` to be on by default in our pre-commit hook? Currently it filters on staged files, but that could mean flagging up 300 instances of `line-too-long` when the user only fixed a single typo.

ZedThree:
Added docs, but needs tests for `--git-since` -- which will be a little complicated!

ZedThree:
There's something really annoying going on around absolute/normalised paths on macs:

```
[2026-03-27][11:36:41][fortitude_linter::line_filter][DEBUG] pwd: "/private/var/folders/tb/y368xp_x10s3ty1b_mtl5mxr0000gn/T/.tmpokWiYp\n"
[2026-03-27][11:36:41][fortitude::check][DEBUG] Using line filter: Some(FilterMap { inner: {"/private/var/folders/tb/y368xp_x10s3ty1b_mtl5mxr0000gn/T/.tmpokWiYp/test.f90": [LineRange { start: OneIndexed(10), end: OneIndexed(16) }]} })
[2026-03-27][11:36:41][fortitude_workspace::resolver][DEBUG] Dropping path due to file filter: "/var/folders/tb/y368xp_x10s3ty1b_mtl5mxr0000gn/T/.tmpokWiYp/test.f90"
```

The line filter is getting a path starting `/private/var/...` and the file discovery is getting `/var/...`. I think one of them is a symlink to the other, and I'm sure I'm using the same function to normalise the paths, so I don't understand why we're not getting a consistent result!

ZedThree:
Finally sorted out the tests:

In order to apply the git line filters correctly, we need to _exactly_ match file names, which may be complicated by

a) relative / absolute paths
b) symbolic links
c) non-canonical forms (`a/../b`)

So we need to normalise & canonicalise both the filenames in the filter and in the found file list.

The performance effect should be negligible because we only need to do all this once per file, and only if using these new flags

LiamPattinson:
Just resolved a merge conflict. I went with a merge over a rebase to avoid GitHub giving me credit for all of your work!

LiamPattinson (approved):
All looks good!

PR #614, opened 2026-03-23
Add `--fixable`, `--unfixable`, and `--extend_fixable` options

### LINKED ISSUE #455: Add `--fixable` options
Ruff gives the following options to specify which rules should be considered eligible for fixes:

```
      --fixable <RULE_CODE>                                List of rule codes to treat as eligible for fix. Only applicable when fix itself is enabled (e.g., via `--fix`)
      --unfixable <RULE_CODE>                              List of rule codes to treat as ineligible for fix. Only applicable when fix itself is enabled (e.g., via `--fix`)
      --extend-fixable <RULE_CODE>                         Like --fixable, but adds additional rule codes on top of those already specified
```

This would be very useful to have in combination with `--unsafe-fixes`, where users may prefer to manually fix some of these issues than to trust our auto-fix. Users may also wish to set `--fixable=style` to quickly fix up only style issues etc.

LiamPattinson:
Resolves https://github.com/PlasmaFAIR/fortitude/issues/455

I've added some basic tests, but haven't yet tried every combination of command line vs config file to really put it through its paces.

### DIFF
--- a/crates/fortitude/src/cli.rs
+++ b/crates/fortitude/src/cli.rs
@@ -286,6 +286,39 @@ pub struct CheckCommand {
     )]
     pub extend_per_file_ignores: Option<Vec<PatternPrefixPair>>,
 
+    /// List of rule codes to treat as eligible for fix. Only has an effect if `--fix` is also enabled.
+    #[arg(
+        long,
+        value_delimiter = ',',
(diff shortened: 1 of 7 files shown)


ZedThree (approved):
Thanks @LiamPattinson!

PR #616, opened 2026-03-31
Upgrade implicit typing rules

### LINKED ISSUE #609: Rule C001: also warn about missing `type` in `implicit none (external)`
The message included for rule C003 (#192) highlights that we should be explicit about external procedures and suggests the full statement 

```
implicit none (type, external)
```

If you scan through the first paragraph however, you might end up omitting the `type`, which still passes fortitude and bypasses C001!

Minimal example:
```
$ cat test.f90
module mymod
        implicit none (external)

        public
        integer :: i
end module mymod

$ fortitude check test.f90
fortitude: 1 files scanned.
All checks passed!
```

Is it possible to make C001 check for `type` if the brackets exist?

LiamPattinson:
Resolves https://github.com/PlasmaFAIR/fortitude/issues/609

Our rules for implicit typing are missing a few edge cases. For example:

- As stated in #609, `implicit-typing` misses `implicit none (external)`.
- `superfluous-implicit-none` doesn't check if the `implicit none` in the higher scope is actually equivalent, e.g. `implicit none (type) == implicit none`, but `implicit none (external) != implicit none`.

I've started a refactor of the machinery around these rules so we can better handle each case. I still need to make some changes to a few rules and add all of the edge cases to the tests.

This one affects rules that have been long out of preview, so we'll have to make sure we get this right!

### DIFF
--- a/crates/fortitude_linter/src/ast/types.rs
+++ b/crates/fortitude_linter/src/ast/types.rs
@@ -3,6 +3,7 @@
 use std::str::FromStr;
 
 use anyhow::{Context, Result, anyhow};
+use bitflags::bitflags;
 use itertools::Itertools;
 use ruff_source_file::SourceFile;
 use strum_macros::{Display, EnumIs, EnumString, IntoStaticStr};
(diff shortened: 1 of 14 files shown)


crates/fortitude_linter/src/ast/types.rs
```diff
@@ -484,3 +486,76 @@ pub(crate) enum BlockExit {
...
     Error,
 }
+
+bitflags! {
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    struct ImplicitNoneType: u8 {
+        const TYPE = 0b0001;
+        const EXTERNAL = 0b0010;
+    }
+}
```
ZedThree, line 496: Nice!

crates/fortitude_linter/src/rules/correctness/implicit_typing.rs
```diff
@@ -211,34 +275,34 @@ impl AstRule for ImplicitExternalProcedures {
...
+
+        // If we get here, it's either `implicit none` or `implicit none
+        // (type)`, so we want to add `external` to it.
+        let edit = node
+            .children(&mut node.walk())
+            .find(|child| child.to_text(src.source_text()).unwrap().to_lowercase() == "type")
+            .map_or_else(
+                || Edit::insertion(" (type, external)".to_string(), node.end_textsize()),
+                |type_node| Edit::insertion(", external".to_string(), type_node.end_textsize()),
+            );
```
ZedThree: Can we use `ImplicitTypingEdit` here to save the duplication?

ZedThree (dismissed):
Looks good, thanks @LiamPattinson! Just one question really, but it's not a big deal

crates/fortitude_linter/src/rules/correctness/implicit_typing.rs
LiamPattinson: There's possibly some way to simplify everything that I'm not seeing. I can't see a clean way to give `ImplicitTypingEdit` the ability to add `external` without adding an extra branch in the constructor and a toggle that would only be switched on in `implicit-external-procedures`. Rather than reducing duplication, it would just move it elsewhere.

I've moved that messy edit to a function at the top of the file so that it's closer to similar functions and is hopefully easier to read. Feel free to reorganise anything if you think there's a better to achieve this.

ZedThree: Ah, ok, never mind then!

ZedThree (approved):

PR #619, opened 2026-04-02
Add superfluous save lint

VincentVanlaer:
Since Fortran 2008, `save` statements and attributes are no longer needed for module level variables.

### DIFF
--- a/crates/fortitude_linter/src/rules/modernisation/save.rs
+++ b/crates/fortitude_linter/src/rules/modernisation/save.rs
@@ -1,4 +1,4 @@
-use ruff_diagnostics::{Diagnostic, Edit, Fix, FixAvailability, Violation};
+use ruff_diagnostics::{AlwaysFixableViolation, Diagnostic, Edit, Fix};
 use ruff_macros::{ViolationMetadata, derive_message_formats};
 
 use crate::{
@@ -38,16 +38,14 @@ pub(crate) struct SuperfluousSave {
     entity: &'static str,
 }
 
-impl Violation for SuperfluousSave {
-    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Always;
-
+impl AlwaysFixableViolation for SuperfluousSave {
     #[derive_message_formats]
     fn message(&self) -> String {
         format!("save {} is superfluous at the module level", self.entity)
     }
 
-    fn fix_title(&self) -> Option<String> {
-        Some("Remove `save`".to_string())
+    fn fix_title(&self) -> String {
+        "Remove `save`".to_string()
     }
 }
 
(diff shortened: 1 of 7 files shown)


ZedThree:
This looks great, thanks @VincentVanlaer! Please could you just expand the test to include a `save` attribute in a `subroutine`, just to check that it _doesn't_ get picked up?

ZedThree:
Oh, please could you also run `cargo dev generate-all` to update the docs with the new rule?

ZedThree:
I've just checked this on a subroutine:

```f90
  subroutine count()
    integer, save :: num = 1
    print*, num
    num = num + 1
  end subroutine count
```

which raises a false positive. This is because one of the entrypoints is `variable_declaration` -- we just need to check that if we hit that, that the parent is a `module`/`submodule` (or `program`?), and if not, we can bail.

VincentVanlaer:
Yes, sorry for not replying yet. I ran out of time to work on this. I had this working before, so I think I messed up when refactoring. I'll push a fixed version.

ZedThree:
No worries, we appreciate the contribution! :)

I'm happy to fix this up if you don't have time to work on it any more

ZedThree:
Thanks @VincentVanlaer, looks good! My only question is should this actually be a `safe_edit`, instead of unsafe? It should always be correct to apply this fix, right?

VincentVanlaer:
I think so as well. I just put it on unsafe since it is a new fix. But I'm happy to change it.

ZedThree:
Ah, we have the `--preview` flag for that :) Yes please, let's make it a safe fix. That might encourage people to use it too.

I have one last bit of bikeshedding -- maybe this belongs more under `modernisation` than `style`? There's definitely an argument that it is a style violation, and has never actually been needed in practice. It's not a huge deal to have it in either category, except that we're more likely to have `modernisation` rules on by default than `style`.

Anyone else have thoughts on the appropriate category?

LiamPattinson:
I'd lean towards `modernisation` myself.

VincentVanlaer:
I have moved the lint to modernisation and made the fix safe.

crates/fortitude_linter/src/rules/modernisation/save.rs
```diff
@@ -0,0 +1,123 @@
...
+/// module example
+///     integer :: a
+/// end module example
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct SuperfluousSave {
+    entity: &'static str,
+}
+
+impl Violation for SuperfluousSave {
```
LiamPattinson: There's a closely related trait `AlwaysFixableViolation` that could help reduce the boilerplate here, but it's functionally the same as what you've done here.

LiamPattinson (dismissed):
Thanks for your contribution @VincentVanlaer, this looks great! I have one comment, but it's nothing major.

ZedThree (approved):
I made the switch to `AlwaysFixableViolation`, and also switched removing the standalone `save` statement to use `FortitudeNode::edit_delete`, which will take care of removing the whole line if it will be empty after removing the statement

PR #624, opened 2026-04-23
Add rule `bare-decimal`

### LINKED ISSUE #579: Rule to enforce a digit on each side of decimal points
It would be good to have a rule that would enforce having a digit (i.e., writing out the implied `0`) on each side of the decimal point.

So this would flag both
```
real, save  ::  a = .1
real, save  ::  b = 1.
```

and turn them into 
```
real, save  ::  a = 0.1
real, save  ::  b = 1.0
```

Mentioned in "Modern Fortran" by Clerman and Spector, Rule 24.

LiamPattinson:
Fixes https://github.com/plasmaFAIR/fortitude/issues/579

Happy to workshop the name, and I'll admit my implementation isn't the prettiest!

Also includes a drive-by fix on the docs for `double-precision`.


crates/fortitude_linter/src/rules/style/literals.rs
```diff
@@ -0,0 +1,102 @@
...
+impl AlwaysFixableViolation for ImplicitZeroDigit {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let Self {
+            is_trailing,
+            literal,
+            ..
+        } = self;
+        let trailing = if *is_trailing { "trailing" } else { "leading" };
+        format!("{trailing} decimal point in literal {literal}")
```
ZedThree: This perhaps just bikeshedding, but I think adding the type makes it a bit clearer what the issue is:

```suggestion
        format!("{trailing} decimal point in `real` literal `{literal}`")
```

crates/fortitude_linter/src/rules/style/literals.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    fn fix_title(&self) -> String {
+        let Self { preferred, .. } = self;
+        format!("Prefer '{preferred}'")
```
ZedThree: We should probably write this down, because we're not very consistent, but I prefer to use backticks for quoting code snippets (and use single-quotes for filenames/strings)

```suggestion
        format!("Prefer `{preferred}`")
```

crates/fortitude_linter/src/rules/style/literals.rs
```diff
@@ -0,0 +1,102 @@
...
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for floating point literals that begin or end in a bare decimal point,
+/// such as `.5` or `2.`.
+///
+/// ## Why is this bad?
+/// Floating point literals that begin or end in a bare decimal point can be
+/// difficult to read and may lead to confusion. For example, `.5` could be
+/// misread as `5.`. It is generally recommended to include a leading zero
```
ZedThree: Very small tweak:

```suggestion
/// missed and may lead to confusion. For example, `.5` could be
/// misread as `5`. It is generally recommended to include a leading zero
```

ZedThree (dismissed):
Nice, I'm pretty sure we've had people asking for this rule.

I only have very minor nit-picky comments!

crates/fortitude_linter/src/rules/style/literals.rs
LiamPattinson: Agreed! This might have been a case of me tab-completing whatever Copilot suggested, so apologies for that.

ZedThree:
Just saw you questioned the rule name too. It's definitely not bad, but it is maybe a little vague? 

Other suggestions:

- `real-literal-implicit-zero`
- `real-literal-bare-decimal`
- `literal-number-implicit-zero`

Not sure!

ZedThree:
The short code is going to clash with #619 btw

LiamPattinson:
> The short code is going to clash with #619 btw

Good catch! I'll increment it by 10

LiamPattinson:
I went with just `bare-decimal` to keep things short and snappy. I think the meaning should be fairly unambiguous.

LiamPattinson:
Added tests for kind suffixes: I'd forgotten to catch that case, but luckily the code handled them without modification!

ZedThree (approved):

PR #626, opened 2026-05-01
Add support for `pyproject.toml` config files

### LINKED ISSUE #617: Support `pyproject.toml` files
Some users have non-fpm projects that also include Python components, and therefore would prefer to reuse their `pyproject.toml` file than to add a `fortitude.toml`.

It should be fairly straightforward to extend the existing functionality for supporting `fpm.toml` to also include `pyproject.toml`. The only major difference would be to use `[tool.fortitude]` over `[extra.fortitude]`.

ZedThree:
Closes #617

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,10 +206,10 @@ Run `fortitude explain` to see which rules are in preview mode.
 
 ## Configuration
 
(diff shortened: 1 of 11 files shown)


LiamPattinson (approved):
Thanks @ZedThree, all looks good

PR #627, opened 2026-05-05
Port ruff `Diagnostic`

ZedThree:
Ruff has significantly changed the structure of their `Diagnostic`
implementation since we last updated our use of their crates. In particular,
most of the implementation has moved from `ruff_diagnostics` to `ruff_db`, which
has a bunch of dependencies we don't want, including a Python parser.

In order to get some of the new `Diagnostic` features from them that we really
want (for example, multiple annotations for a violation), we're going to need to
reimplement a bunch of things.

This PR just brings across the ruff utilities that we are using as of `ruff
0.9.6`, which is basically just `Diagnostic` and the `violation_metadata` macro (plus a very minor refactor of converting the `FromAstNode` trait to member functions, which saves us an import). I'll do the upgrades separately to make it easier to review.

About half this PR is literally just fixing imports.

There's a few other bits from the `ruff_diagnostics` crate (`Fix`, `Edit`, etc.) that haven't changed since 0.9.6, and I didn't see any reason to port them, so we just re-export them from `fortitude_linter::diagnostics`. If we do decide to port them later, we won't need to fix the imports at least.


crates/fortitude_linter/src/diagnostics/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+use serde::{Deserialize, Serialize};
+use tree_sitter::Node;
+
+use crate::{fix::FixTable, rules::Rule, settings::CheckSettings, traits::TextRanged};
+
+pub use diagnostic_message::DiagnosticMessage;
+pub use violation::{AlwaysFixableViolation, FixAvailability, Violation, ViolationMetadata};
+
+// Re-export some things from ruff
+pub use ruff_diagnostics::{Applicability, Edit, Fix, IsolationLevel, SourceMap, SourceMarker};
```
LiamPattinson, line 25: I'm a little confused as to how this works, given `ruff_diagnostics` have been removed from the `Cargo.toml` file.

LiamPattinson (approved):
Looks good! Only have one question about the implementation.

ZedThree:
It's only been removed from the individual crates that no longer use it,
it's still in the workspace and `fortitude-linter` crate `Cargo.toml` files

LiamPattinson:
Ah, fair enough, I think I must have misread which `Cargo.toml` I was looking at

PR #635, opened 2026-05-13
Various fixes to settings

LiamPattinson:
Fixes some minor inconsistencies and bugs with settings

### DIFF
--- a/crates/fortitude_linter/src/stylist.rs
+++ b/crates/fortitude_linter/src/stylist.rs
@@ -216,8 +216,11 @@ impl fmt::Display for Quote {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, CacheKey)]
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, CacheKey, strum_macros::Display,
+)]
 #[serde(deny_unknown_fields, rename_all = "kebab-case")]
+#[strum(serialize_all = "kebab-case")]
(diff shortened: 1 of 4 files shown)


crates/fortitude_linter/src/stylist.rs
```diff
@@ -231,6 +229,16 @@ pub enum Capitalisation {
     Titlecase,
 }
 
+impl fmt::Display for Capitalisation {
```
ZedThree: Surprised we need to do this! What was the issue with `strum_macros::Display`?

ZedThree (dismissed):

crates/fortitude_linter/src/stylist.rs
LiamPattinson: It was printing the default as `Lowercase`, ironically. I tried to see if any serde/strum solution existed to force it to lowercase, but I couldn't figure it out.

ZedThree: Oh, I ran into this before, the docs are really hidden, [but it is possible](https://docs.rs/strum/latest/strum/additional_attributes/index.html)

LiamPattinson: Thanks, fixed in [cf662f9](https://github.com/PlasmaFAIR/fortitude/commit/cf662f99522cf8f99b4c19007a420d05226530a7). I probably should have dug deeper into this, I'm sure I've used the same solution elsewhere in Fortitude.

ZedThree (approved):

PR #638, opened 2026-05-15
keyword-has-whitespace.md: fix typo

matthewrmshin:
Fix typo: `in` should be `is`.


LiamPattinson (commented):
Nice catch, thanks @matthewrmshin!

The docs are normally generated from the code itself, so to make sure your fix won't be overwritten the change needs to be made in the doc-comment in https://github.com/PlasmaFAIR/fortitude/blob/main/crates/fortitude_linter/src/rules/style/keywords.rs. You can then generate the docs using `cargo dev generate-all`.

matthewrmshin:
@LiamPattinson Thanks. Typo should now be fixed in both places.

LiamPattinson (approved):
Thanks, all looks good to me!

PR #640, opened 2026-05-18
Merge `DiagnosticMessage` into `Diagnostic`

ZedThree:
Moves all of `DiagnosticMessage`'s capabilities into `Diagnostic`. This was something we took from Ruff, and I think the main reason to split this into two structs was so we didn't have to store a `SourceFile` in `Diagnostic`. However, the multiple annotations feature in Ruff uses `Span`, which is a struct of the `TextRange` and `SourceFile`, so we'll be storing the file anyway


LiamPattinson (approved):
Thanks, this looks much cleaner than before!

PR #643, opened 2026-05-20
Pass context struct into rule check functions

ZedThree:
`CheckContext` bundles up a bunch of things that we might want to use in `check` functions:

- `SourceFile`
- `Stylist`
- `Settings`
- `SymbolTable`

and provides a `.create_diagnostic()` method. At the moment this isn't super useful, but when changing `Diagnostics` to allow multiple annotations, creating a "normal" diagnostic will be a bit more complicated, and this will wrap that complexity up and keep the call sites clean.

We also now have `Stylist` available in every rule, so we can start modifying fixes to make them match the local formatting.

### DIFF
--- a/crates/fortitude_linter/src/traits.rs
+++ b/crates/fortitude_linter/src/traits.rs
@@ -15,6 +15,12 @@ where
     }
 }
 
+impl<'a> HasNode<'a> for &'a Node<'a> {
+    fn node(&self) -> &Node<'a> {
+        self
+    }
+}
+
 #[macro_export]
 macro_rules! impl_has_node {
     ($ty: ty) => {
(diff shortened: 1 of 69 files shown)


crates/fortitude_linter/src/traits.rs
LiamPattinson, line 1: I'm liking these changes! Being able to supply a custom text range or a node to the same function is pretty sweet.

LiamPattinson (approved):
Thanks @ZedThree, this is much cleaner and more extensible!

crates/fortitude_linter/src/traits.rs
ZedThree: I'm a bit worried the `TextRange` version makes an additional copy, but I'm pretty sure the compiler will elide it in practice

PR #644, opened 2026-05-20
Bump ruff dependencies to 0.15

ZedThree:
Main change is to `SourceLocation` accounting for BOM characters, and needing to
switch to the new `LineColumn` which uses `line` instead of `row`. This latter
change has an impact on a couple of the snapshots.

The other change is `ruff_options_metadata` is now a separate crate, and we can
remove our vendored copy of `options_base.rs`.


LiamPattinson (approved):
Thanks @ZedThree, all looks good.

I've taken a look at the changes in Ruff, and I can't quite figure out what the difference is supposed to be between `LineColumn` and `SourceLocation`. They both contain a line and a column both account for UTF character sizes, and both are serialisable with serde. The only difference seems to be that one calls it `column` and the other `character_offset`, and otherwise I don't know why one should be used over the other.

ZedThree:
It's basically to do with [BOM characters](https://en.wikipedia.org/wiki/Byte_order_mark), which are zero-width, non-displaying characters. `LineColumn` doesn't count them and gives a visual location, `SourceLocation` does and gives a byte location. Or something like that

PR #645, opened 2026-05-21
Add rule `keyword-reuse`

### LINKED ISSUE #39: Variable name aliases 'reserved' word
Fortran doesn't have keywords, so truly cursed code like the following is unfortunately valid:

```f90
integer function else(if, elseif, endif, then)
  integer, intent(in) :: if
  integer, intent(in) :: elseif
  integer, intent(in) :: endif
  integer, intent(in) :: then

  if(if == elseif) then
    else = then
  elseif(then == if) then
    else = endif
  else
    else = if
  endif
end function else
```

This can make the code extremely difficult to understand, and even in less egregious cases it can mess up syntax highlights:

```f90
subroutine write(arg1, arg2)
  ...
end subroutine write

call write(...)
call some_other_subroutine(...)
```

This rule should warn when the user has named something after a 'reserved' word in the standard. It should trigger on the following nodes, and probably a few others too:

- variable_declaration
- function_statement
- subroutine_statement
- module_statement
- submodule_statement
- program_statement
- use_alias
- derived_type_definition
- binding_name (type-bound procedures)

It may be difficult to make this rule performant, as it would have to check every new definition against every 'reserved' word.

LiamPattinson:
Closes https://github.com/PlasmaFAIR/fortitude/issues/39

Labelled `S301` rather than `S234`, as `S23*` rules are dealing with poorly formatted keywords, while this rule is about identifiers with inappropriate names. I'd like `S302` to be a similar rule `intrinsic-reuse`, which will catch things like `sin` etc.

Primarily uses the symbol table to catch keyword clashes, but needs special handling to deal with block labels, so the rule is mostly checked via the `check_path` function right after a new symbol table is created, but is separately implemented as an AST rule just to catch `block_label_start_expression`.

It's currently missing identifiers in a function `result`. I'll have to do a little extra work on `SymbolTable` to ensure we're catching that too.

This has also raised a few `tree-sitter-fortran` bugs for me:

- do loops labelled `do`
- modules called `module` (but weirdly _not_ programs called `program`)

There are probably a bunch more cases like that worth checking.


LiamPattinson:
> It's currently missing identifiers in a function `result`. I'll have to do a little extra work on `SymbolTable` to ensure we're catching that too.

I could quite figure out how to do it robustly, so I've implemented a workaround for now on the `keyword-reuse` rule. @ZedThree Is this something you think could be solved on `SymbolTable`?

crates/fortitude_linter/resources/test/fixtures/style/S301_keyword_reuse.f90
```diff
@@ -0,0 +1,70 @@
+module program  ! Couldn't call this 'module', syntax error
```
ZedThree: This (and the other things marked as syntax errors) work for me -- what was giving you the error?

LiamPattinson: Whoops! They were flagged as syntax errors in my editor, but I think that was FortLS rather than Fortitude. I'll fix it up.

LiamPattinson:
I've rebased this to take into account the changes to `SymbolTable`, with the further modification of adding type definitions as a symbol type. It hasn't yet been fleshed out, and so far only captures the name.

Some issues remain:

- The full nodes are used for diagnostic ranges, while it should really just target the name nodes. I'm not sure how to fix this in general, but I'll take another look -- maybe I'm missing something obvious.
- Programs and modules aren't captured in symbol tables, so need to be checked manually.

ZedThree:
- `VariableDeclaration` has something like this, `NameDecl` and `get_name_node_of_declarator`, maybe all the `Symbol` variants should have a `.name() -> NameDecl` method (and we add `NameDecl::as_str()`)
- We should add `translation_unit` to `BEGIN_SCOPE_NODES` and then we can capture `program`/`module`/`submodule` in `SymbolTable::new`

Also, it might be useful to have something like `.expect_child_with_name() -> Result<Node>` in `FortitudeNode`

LiamPattinson:
I've made the changes requested. The structs for type definitions, modules, and programs are just stubs for now.

One unintended change is that symbols in the symbol table now retain their original capitalisation when you call `symbol.name()`. Hopefully this is desirable?

The new changes do lead to some confusion for me about the difference between:

- `self.name()`
- `self.node()`
- `self.name_node()`
- `self.name.name()`
- `self.name.node()`
- `self.name.name_node()`

so it might require some further clean-up.

LiamPattinson:
I had another go and added a distinction between `NameDecl`, which contains things like sizes and initialisation, and `Name`, which is just a name. `NameDecl` now contains a `Name` rather than a `String`.

LiamPattinson:
Updated so that the rule code is now `S311` (and always was throughout the git history).

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__keyword-reuse_S311_keyword_reuse.f90.snap
```diff
@@ -0,0 +1,316 @@
...
+./resources/test/fixtures/style/S311_keyword_reuse.f90:52:42: S311 Keyword `write` used as an identifier
+   |
+51 |     block: block
+52 |         integer :: allocate, deallocate, write
+   |                                          ^^^^^ S311
+53 |         print *, allocate, deallocate, write
+54 |     end block block
+   |
+
+./resources/test/fixtures/style/S311_keyword_reuse.f90:51:5: S311 Keyword `block` used as an identifier
```
ZedThree: This might be better if it said "used as a label"?

ZedThree (dismissed):
LGTM! Minor quibble over the message, but otherwise great. The symbol table stuff looks particular good!

crates/fortitude_linter/src/rules/style/snapshots/fortitude_linter__rules__style__tests__keyword-reuse_S311_keyword_reuse.f90.snap
LiamPattinson: Agreed, I think that would be more intuitive. Fixed in https://github.com/PlasmaFAIR/fortitude/pull/645/commits/fcdc1ea38b8eb7f15ac5485f466d87ddc6617de4

ZedThree: Ah, sorry, I was not super clear! I was thinking just `label` for the `block` labels. `identifier` is probably fine for most other uses?

LiamPattinson: Ah, I see. Fixed in https://github.com/PlasmaFAIR/fortitude/pull/645/commits/61aed8d25d51ec572ead9ceba6c4ec31cfb7f9e2. It looks like it works for the labels on do loops too.

LiamPattinson:
Rebased again to handle changes to diagnostics. Turns out I needn't have bothered as there were no changes, but I may as well keep things on top of `main`.

ZedThree (approved):
Looks great, thanks @LiamPattinson!

PR #648, opened 2026-05-24
Document fixed-form Fortran support status

### LINKED ISSUE #500: Docs need to make clear that we don't (currently) handle fixed-form

thuantip:
## Summary

- Add an FAQ entry clarifying that Fortitude currently targets free-form Fortran source.
- Note that fixed-form files such as `.f`, `.for`, and `.ftn` are not included by default.
- Clarify that adding fixed-form extensions to `include` does not make Fortitude parse them as fixed-form source.

Fixes #500.

## Verification

- git diff --check


ZedThree (approved):
Thanks @thuantip, this looks good! Thanks for your contribution! :)

Issue #621, opened 2026-04-04
Exit zero for selected rules


LiamPattinson:
> Thank you! I definitely missed it, sorry. Is there also a way to make the command terminate with exit status 0 only when the selected rules are violated but still output the entire diagnostic? This again is something that is useful to add fortitude to an already established project, for example in the GitLab CI, to make it blocking only on selected rules. 

 _Originally posted by @eduardz1 in [#620](https://github.com/PlasmaFAIR/fortitude/issues/620#issuecomment-4187175590)_

This could be a way for users to set some rule violations as 'errors' while others are simply 'warnings'. By default, I'd lean towards all violations being treated as errors, as I would find it difficult to decide how seriously somebody should take any given rule, but I can see the value of giving people the option to customise this.

eduardz1:
As a simple workaround, we can currently just execute fortitude twice, similar to:

```shell
fortitude check --select ALL --output-format="gitlab" --output-file fortitude-report.json || true
fortitude check
```

To integrate, for example, in GitLab's CI.

This has the downside that all rules are still treated as 'warnings' in GitLab's codequality, it would be better to separate the rules between the various 'severity' levels,a s described in https://docs.gitlab.com/ci/testing/code_quality/#code-quality-report-format

ZedThree:
This seems like a good idea, and I agree that everything as an error is the sensible default.

Given we already have `--exit-zero`, maybe `--exit-zero-only=RULES` is the way to go? Rather than introducing a severity concept and having something like `--warnings=RULES`

eduardz1:
> This seems like a good idea, and I agree that everything as an error is the sensible default.
> 
> Given we already have `--exit-zero`, maybe `--exit-zero-only=RULES` is the way to go? Rather than introducing a severity concept and having something like `--warnings=RULES`

I am not sure that maybe it's too specific to GitLab, but I think a better solution would be something like:

```toml
[check]
severity-default = "info"

[check.severity]
C022 = "blocker"
PORT011 = "blocker"
```

which would override the hardcoded warning level in `gitlab.rs`

eduardz1:
But I think both solutions can be implemented, they serve different use cases

PR #650, opened 2026-06-02
Add options to set manually (no-)exit-zero on specific rules)

eduardz1:
This PR addresses https://github.com/PlasmaFAIR/fortitude/issues/621 and allows to exit with zero on specific rules or, vice-versa, exit with an error on specific rules. This is useful in CI to make the script fail only on specified rules

### DIFF
--- a/crates/fortitude/src/check.rs
+++ b/crates/fortitude/src/check.rs
@@ -202,6 +202,7 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
         unsafe_fixes,
         show_fixes,
         output_format,
+        ref non_zero_on,
         ..
     } = file_configuration.settings.check;
 
@@ -273,11 +274,20 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
     }
 
     let diagnostics = results.diagnostics;
+    let has_blocking_diagnostics = diagnostics
+        .messages
+        .iter()
+        .any(|message| non_zero_on.contains(message.rule()));
+    let has_blocking_fixes = diagnostics
+        .fixed
+        .values()
+        .any(|fix_table| fix_table.keys().any(|rule| non_zero_on.contains(*rule)));
+
     if !cli.exit_zero {
         if fix_only {
             // If we're only fixing, we want to exit zero (since we've fixed all fixable
             // violations), unless we're explicitly asked to exit non-zero on fix.
-            if cli.exit_non_zero_on_fix && !diagnostics.fixed.is_empty() {
+            if cli.exit_non_zero_on_fix && has_blocking_fixes {
                 return Ok(ExitCode::FAILURE);
             }
         } else {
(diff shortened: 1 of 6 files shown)


crates/fortitude/tests/check.rs
```diff
@@ -1339,9 +1509,9 @@ program test
   write (*, '("╔════════════════════════════════════════════╗")')
   write (*, '("║  UTF-8 LOGO BOX                            ║")')
   write (*, '("╚════════════════════════════════════════════╝")')
-  !-- transform into g/cm³   
+  !-- transform into g/cm³
```
LiamPattinson, line 1512: I haven't had a chance to go through everything yet, but just wanted to flag that the trailing whitespace in this test and similar tests was intentional.

eduardz1: Ah sorry, I think it was removed due to the pre-commit

eduardz1: I see that I ran into the same problem in my other PR as well. If that's okay, instead of adding the spaces again, I can add `\x20` so that it's more explicit and it's not removed automatically by the pre-commit hooks

crates/fortitude/src/check.rs
```diff
@@ -273,22 +274,31 @@ pub fn check(args: CheckCommand, global_options: GlobalConfigArgs) -> Result<Exi
...
 
     let diagnostics = results.diagnostics;
+    let has_blocking_diagnostics = diagnostics
+        .messages
+        .iter()
+        .any(|message| non_zero_on.contains(message.rule()));
+    let has_blocking_fixes = diagnostics
+        .fixed
+        .values()
+        .any(|fix_table| fix_table.keys().any(|rule| non_zero_on.contains(*rule)));
```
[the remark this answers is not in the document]
eduardz1: I haven't thought of that, I think it might make more sense to keep the old behavior in this specific instance

crates/fortitude/tests/check.rs
ZedThree: Oh, that's a good idea. Would you mind doing that in a separate PR?

eduardz1: > Oh, that's a good idea. Would you mind doing that in a separate PR?

Sure, this breaks slightly in practice because in `crates/fortitude/tests/check.rs` the results of the snapshots are saved inline in the rust source. I think the easiest solution would be to also save those in `.snap` files, this way in the rust sources we can have `\x20` (for raw strings we actually need to do something like `[SPACE]` and `.replace([SPACE], " ")`) and in the pre-commit hook we can just add `exclude: \.snap(\.new)?$` to the `trailing-whitespace` and `end-of-file-fixer` hooks

ZedThree:
I think this is now replaced by #693, or at least you can achieve basically the same thing with `severity-overrides`

eduardz1:
> I think this is now replaced by #693, or at least you can achieve basically the same thing with `severity-overrides`

Yes I agree, or at least I don't need this level of granularity for our use case.

PR #652, opened 2026-06-02
Expand `SymbolTable` to cover procedures

ZedThree:
This captures procedure names in a scope, including `function` `result` variables, moving us towards a more complete semantic parsing. The `function` name (or `result` name) appears as a `Variable` in its own scope, and as a `Procedure` otherwise.

`SymbolTable` gets a few methods to iterate over everything in a scope -- might be useful to have these on `SymbolTables` as well? And/or something to get the current scope?

I also reworked `HasNode` to be a normal `derive` macro.

Thinking ahead to e.g. #645, maybe `SymbolTables` should hold `Vec<Scope>` with `Scope` something like an enum with each variant holding a `SymbolTable`:

```rust
enum Scope {
  Function(Function),
  ..
}

struct Function {
  name: String,
  locals: OnceCell<SymbolTable>,
  args: OnceCell<SymbolTable>,
  attrs: .. // and so on
}
```

Using `OnceCell` here so that we a) don't need multiple passes over the AST, and b) we only need to create them if needed.

### DIFF
--- a/crates/fortitude_linter/src/ast/symbol_table.rs
+++ b/crates/fortitude_linter/src/ast/symbol_table.rs
@@ -54,6 +54,19 @@ impl<'a> SymbolTable<'a> {
             .filter_map(|decl| VariableDeclaration::try_from_node(&decl, src).ok())
             .for_each(|line| new_table.insert_from_decl_line(line));
 
+        // The `function` statement itself _may_ also be the declaration line if
+        // it has a type as a procedure attribute. If it doesn't, then it will
+        // either have an explicit decl line, which is handled above, or it's
+        // implicitly typed, which we don't currently handle here at all
+        if scope.is_named() && scope.kind() == "function" {
+            let stmt = scope
+                .child(0)
+                .expect("`function` must have `function_statement` as zeroth child");
+            if let Ok(decl) = VariableDeclaration::try_from_fn_stmt(&stmt, src) {
+                new_table.insert_from_decl_line(decl);
+            }
+        }
+
         new_table
     }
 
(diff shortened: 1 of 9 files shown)


crates/fortitude_linter/src/ast/symbol_table.rs
```diff
@@ -54,38 +78,80 @@ impl<'a> SymbolTable<'a> {
             .filter_map(|decl| VariableDeclaration::try_from_node(&decl, src).ok())
             .for_each(|line| new_table.insert_from_decl_line(line));
 
+        // The `function` statement itself _may_ also be the declaration line if
+        // it has a type as a procedure attribute. If it doesn't, then it will
+        // either have an explicit decl line, which is handled above, or it's
+        // implicitly typed, which we don't currently handle here at all
```
LiamPattinson, line 84: Does this imply that Fortitude will crash, or just that it'll miss the function name in the symbol table?

LiamPattinson (approved):
Thanks @ZedThree, this looks great! Obviously this will be useful for `keyword-reuse`, but it will also greatly simplify the implementation of a bunch of new rules. Just off the top of my head:

- `unused-variable`
- `undeclared-variable` (would need to check for bare-use within scope first)
- First declaration lines should be procedure dummy parameters
- Procedure dummy parameters should be declared in the same order as they are listed
- `intent(in)` should be declared first, `intent(out)` at the end.

crates/fortitude_linter/src/ast/symbol_table.rs
ZedThree: Just that there won't be an entry in the symbol table -- but this is true of all implicitly typed variables

Issue #659, opened 2026-06-09
`FortitudeNode::edit_delete` should be line-continuation aware


ZedThree:
If we delete `bar` here:

```f90
foo &
  bar &
  zing
```

then currently we'll leave the line-continuation:

```f90
foo &
   &
  zing
```
which is fine, but we could also delete the `&`.

Similarly, if we delete `bar` here:

```f90
foo &
  bar
```

we'll leave the preceding `&`:

```f90
foo &
```

which is invalid.

We should be able to work out which line-continuation characters we can/must remove

PR #660, opened 2026-06-09
Add new SuperfluousWhileTrue rule.

### LINKED ISSUE #605: `do while (.true.)` should be `do`
An LLM generated

```
   do while (.true.)
      xn = x + alpha * p
      if (real(f(xn), kind=dp) <= (fx + ((c1_opt * alpha) * gtp))) then
         return
      end if
      alpha = alpha * rho_opt
      if (alpha < 1e-16_dp) then
         return
      end if
   end do
```

for which Fortitude was silent, but I think it should suggest replacing `do while (.true.)` with `do`

avadean:
@LiamPattinson @ZedThree 

**Add new SuperfluousWhileTrue rule.**

- Checks if a `do while` statement is only evaluating the boolean literal `.true.`. If so, will recommend to reduce the statement to `do`.
- Put in "Style" rule with number 301 - **let me know if I should alter this**.
- Handles the case where additional, redundant parentheses are added by the user e.g `do while ((.true.))`.
- **Can** handle a line return ampersand (&) when it is placed A) inside the parenthesised expression (see test `fail_7`), or B) between the `while` and the parenthesised expression (see test `fail_8`).
- **Cannot** handle a line return ampersand (&) when it is placed between the `do` and the `while`.

Closes #605.


crates/fortitude_linter/resources/test/fixtures/style/S301.f90
```diff
@@ -0,0 +1,163 @@
...
+        end do
+    end subroutine pass_3
+
+    ! PASS 4 -> with FALSE.
+    subroutine pass_4()
+        do while (.false.)
+            x = x + 1
+            if (x > 10) exit
+        end do
+    end subroutine pass_4
```
LiamPattinson, line 40: Not for this PR: Should this be caught by `unreachable-statement`, or a new rule that checks for `if(.false.)` etc?

crates/fortitude_linter/resources/test/fixtures/style/S301.f90
```diff
@@ -0,0 +1,163 @@
...
+    ! PASS 6 -> with expression.
+    subroutine pass_6()
+        do while (x <= 10)
+            x = x + 1
+        end do
+    end subroutine pass_6
+
+    ! PASS 7 -> with TRUE and expression.
+    subroutine pass_7()
+        do while (.true. .and. x <= 10)
```
LiamPattinson, line 59: Not for this PR: We could fix `.true./.false. .and./.or. ...`

crates/fortitude_linter/src/rules/style/mod.rs
```diff
@@ -60,6 +60,7 @@ mod tests {
     #[test_case(Rule::SuperfluousElseStop, Path::new("S255.f90"))]
     #[test_case(Rule::UnsortedUses, Path::new("S271.f90"))]
     #[test_case(Rule::BareDecimal, Path::new("S291.f90"))]
+    #[test_case(Rule::SuperfluousWhileTrue, Path::new("S301.f90"))]
```
LiamPattinson, line 63: I've used `S301` for `keyword-reuse` in https://github.com/PlasmaFAIR/fortitude/issues/659. Not necessarily a problem, we've had clashes like this before, whichever gets merged second will just have to bump the number.

LiamPattinson (approved):
Thanks @avadean, welcome to Fortitude!

All looks good to me. I'm unsure if we should merge this now or wait until after https://github.com/PlasmaFAIR/fortitude/issues/659 is sorted, but I'd lean towards merging now.

crates/fortitude_linter/resources/test/fixtures/style/S301.f90
ZedThree: Probably `unreachable-statement` is reasonable?

ZedThree:
I'm almost done fixing #659 but I found a couple of edge cases I want to check first, so maybe let's just merge this now

LiamPattinson:
I guess uncommenting the unhandled test case could form part of the test suite for #659

PR #663, opened 2026-06-10
Add helpers for handling line continuations in deletions

### LINKED ISSUE #659: `FortitudeNode::edit_delete` should be line-continuation aware

ZedThree:
Closes #659 

Adds a new function `delete_stmt_part` for deleting a node that's part of a statement and handling line continuations (and surrounding whitespace) correctly, and uses it in `superfluous-while-true`

I decided not to make this part of `FortitudeNode::edit_delete` to avoid fun with comments.

This also doesn't handle trailing line continuations because I was struggling to come up with an example where we want to delete the first node that isn't handled by `remove_from_comma_sep_stmt`.

Also as a drive-by, simplifies a couple of other methods, because I realised ruff has a trait `LineRanges` for `str` that lets us get the offset of the start of a line just from the `str`

### DIFF
--- a/crates/fortitude_linter/resources/test/fixtures/style/S301.f90
+++ b/crates/fortitude_linter/resources/test/fixtures/style/S301.f90
@@ -150,14 +150,13 @@ contains
         end do
     end subroutine fail_8
 
-    ! ** auto-fix not currently implemented **
     ! FAIL X -> line continuation inbetween "do" and "while".
-    ! subroutine fail_X()
-    !    do &
-    !            while (.true.)
-    !        x = x + 1
-    !        if (x > 10) exit
-    !    end do
-    !end subroutine fail_X
+    subroutine fail_X()
+       do &
+               while (.true.)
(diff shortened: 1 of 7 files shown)


LiamPattinson (dismissed):
Thanks @ZedThree, all looks good

LiamPattinson:
Did anything change in the rebase, or was it just getting ahead of `main`?

ZedThree:
Just the new test snapshot of `superfluous-while-true` due to the change in the diff output

LiamPattinson (approved):

PR #664, opened 2026-06-10
Update diff format to match ruff's new style

ZedThree:
Ruff's new diff style only includes a single line of column numbers, and is probably easier to read?

This is partly in advance of the incoming diagnostic changes, but also ruff now displays the diff in its normal output too.

Old format:

```text
ℹ Safe fix
22 22 |       "Hello world!"; end if
23 23 | 
24 24 |   ! Raise an error if the body is on a new line.
25    |-  if (condition) &
26    |-      print *, "Hello"
   25 |+  if (condition) then
   26 |+    print *, "Hello"
   27 |+  end if
27 28 |       print *, "World!"
28 29 | 
29 30 |   ! Misleading semicolons: the second statement in the body should be placed
```

new format:
```
22 |       "Hello world!"; end if
23 |
24 |   ! Raise an error if the body is on a new line.
   -   if (condition) &
   -       print *, "Hello"
25 +   if (condition) then
26 +     print *, "Hello"
27 +   end if
28 |       print *, "World!"
29 |
30 |   ! Misleading semicolons: the second statement in the body should be placed
```

### DIFF
--- a/crates/fortitude_linter/src/diagnostics/message/diff.rs
+++ b/crates/fortitude_linter/src/diagnostics/message/diff.rs
@@ -10,7 +10,7 @@ use colored::{Color, ColoredString, Colorize, Styles};
 use ruff_text_size::{Ranged, TextRange, TextSize};
 use similar::{ChangeTag, TextDiff};
 
-use crate::diagnostics::{Applicability, Fix};
+use crate::diagnostics::Fix;
 use ruff_source_file::{OneIndexed, SourceFile};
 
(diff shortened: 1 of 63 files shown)


LiamPattinson (approved):
All looks good, I imagine it'll make it easier to interpret snapshot updates.

(You had me scared with the number of lines changed here!)

ZedThree:
Unfortunately touching how any of the diagnostics get rendered touches masses of lines!

PR #668, opened 2026-06-11
Port Ruff support for multiple annotations in diagnostics

ZedThree:
Includes a few recent open PRs, so they probably want to be merged first.

Changes:
- Default output format puts the filename on a second line
- Preview mode adds the severity to the output (this is currently `error` for everything!)
- Preview mode also shows the diff for each diagnostic individually by default
- Support for adding additional help/info messages, additional annotations in snippets, additional snippets in diagnostics
- Many of the other output formats have also been updated

Example of new preview output:

```
error[OB094][*]: `goto` points to `end do`
  --> crates/fortitude_linter/resources/test/fixtures/obsolescent/labelled_do.f90:60:20
   |
58 |     ! This only has a `goto` that could be a `cycle`
59 |     do i = 1, 10
60 |         if (i > 5) goto 91
   |                    ^^^^^^^
61 |         foo(i) = 2 * i
62 | 91  end do
   |
help: Use `cycle` instead of `goto` to `end do`
57 |
58 |     ! This only has a `goto` that could be a `cycle`
59 |     do i = 1, 10
   -         if (i > 5) goto 91
60 +         if (i > 5) cycle
61 |         foo(i) = 2 * i
62 | 91  end do
63 |
```
(but with colour in the terminal!)

This is a really big change, but most of it is in the test snapshots. I've tried to cut out a few things I'm pretty sure we don't need, but this new system is very general and can create some quite complex diagnostics. Astral use the same system for both Ruff and Ty, which have different needs, so I've also tried to strip out a bunch of that generality.

@eduardz1 This adds severity levels, but doesn't expose any user configuration for changing them. #651 should be built on top of this.

A few points:

- There are `Annotations` -- a diagnostic can have multiple annotations:

```
error[OB094]: `goto` points to `end do`
  --> crates/fortitude_linter/resources/test/fixtures/obsolescent/labelled_do.f90:79:24
   |
77 |     do 110, i = 1,10
78 |       do 120, j = 1,10
79 |         if (i + j > 5) goto 110
   |                        ^^^^^^^^
80 |         A(i, j) = i + j
81 | 120    end do
82 | 110 end do
   | --- `goto` points here
83 |
84 | end program test
   |
help: Use `cycle` instead of `goto` to `end do`

error[OB094][*]: `goto` points to `end do`
   --> FLEUR/src/tools/inpgen2/kpoints/tetcon.f90:360:10
    |
359 | 2000     END DO
360 |          goto 1100
    |          ^^^^^^^^^
361 |
362 |          ! HERE, WE COULD NOT FIND ANY NEW TETRAHEDRON USING J-TH FACE.
    |
   ::: FLEUR/src/tools/inpgen2/kpoints/tetcon.f90:373:1
    |
371 |          voltet(nt) = abs(volnew)
372 |
373 | 1100  END DO
    | ---- `goto` points here
374 |       if (it < nt) goto 1000
375 |       IF (nt > ndiv3) CALL juDFT_error(" nt>ndiv3", calledby="tetcon")
    |
help: Use `cycle` instead of `goto` to `end do`
```

Here I've added a secondary annotation to a rule (not included in this PR, just experimenting!). You can see that if the annotations are close enough together, they get merged into a single one, otherwise there's a `::: <filename>` break.

- There are also `SubDiagnostics` -- a diagnostic can have multiple sub-diagnostics, each of which can have multiple `Annotations` (or none). Here's the same example, but using a `SubDiagnostic` instead:

```
error[OB094]: `goto` points to `end do`
  --> crates/fortitude_linter/resources/test/fixtures/obsolescent/labelled_do.f90:79:24
   |
77 |     do 110, i = 1,10
78 |       do 120, j = 1,10
79 |         if (i + j > 5) goto 110
   |                        ^^^^^^^^
80 |         A(i, j) = i + j
81 | 120    end do
   |
help: Use `cycle` instead of `goto` to `end do`
info: Target of `goto`
  --> crates/fortitude_linter/resources/test/fixtures/obsolescent/labelled_do.f90:82:1
   |
80 |         A(i, j) = i + j
81 | 120    end do
82 | 110 end do
   | ^^^
83 |
84 | end program test
   |
```

Note that the `^^^` underlining, because the annotation is a primary one vs a secondary in the first example.

- If a `SubDiagnostic` has any `Annotations`, the snippets are _not_ merged with other `SubDiagnostic`s

- The `info` sub-diagnostic above appears *after* the `help` suggestion -- that's because that's now implemented as a `SubDiagnostic` without any `Annotations`. The `help` sub-diagnostic gets added automatically via `context.create_diagnostic()` and sub-diagnostics are displayed in the order they were added, so without some fiddling, all further sub-diagnostics will appear after the `help`. This is important, because `--preview` now displays the fix diff:

```
error[OB094][*]: `goto` points to `end do`
  --> crates/fortitude_linter/resources/test/fixtures/obsolescent/labelled_do.f90:60:20
   |
58 |     ! This only has a `goto` that could be a `cycle`
59 |     do i = 1, 10
60 |         if (i > 5) goto 91
   |                    ^^^^^^^
61 |         foo(i) = 2 * i
62 | 91  end do
   |
help: Use `cycle` instead of `goto` to `end do`
info: Target of `goto`
  --> crates/fortitude_linter/resources/test/fixtures/obsolescent/labelled_do.f90:62:1
   |
60 |         if (i > 5) goto 91
61 |         foo(i) = 2 * i
62 | 91  end do
   | ^^
63 |
64 |     do i = 1,10
   |
57 |
58 |     ! This only has a `goto` that could be a `cycle`
59 |     do i = 1, 10
   -         if (i > 5) goto 91
60 +         if (i > 5) cycle
61 |         foo(i) = 2 * i
62 | 91  end do
63 |
```

Note that the `info` snippet appears to merge into the fix diff!

Ruff has a very fancy way of solving this involving a struct with a custom `impl Drop` that is probably overkill for our needs. 

- I've identified a couple of rules that could benefit from these changes, but let's add those in a separate PR


crates/fortitude_linter/src/diagnostics/message/mod.rs
```diff
@@ -1,136 +1,2719 @@
...
+            .secondary_code("PORT021")
+            .documentation_url("https://fortitude.readthedocs.io/rules/star-kind")
+            .build(),
+        ];
+
+        (env, diagnostics)
+    }
+
+    /// Create Ruff-style syntax error diagnostics for testing the various output formats.
+    pub(crate) fn create_syntax_error_diagnostics(
```
LiamPattinson, line 2693: There's a lot for me to work through, but something that jumped out from the snapshot tests: Should this function be rewritten to use Fortitude-specific syntax errors? Or is writing it Ruff-style the point of the test? If it's the latter, I think it should be made clearer that that's the intent.

ZedThree: This isn't really needed for us, I don't think. Ruff has it because they generate diagnostics via a different mechanism than us, not via  a `Rule` so they don't get a `SecondaryCode`. The contents of the test file here don't matter at all, it's really just testing the rendering of `DiagnosticId::InvalidSyntax` rather than `DiagnosticId::Lint`.

`DiagnosticId` is complete overkill for us at the moment, so we could just pare it down to `Lint` for the time being and remove this test. Most of the variants are for raising errors from parsing the config file (e.g. empty `includes` section).

eduardz1:
Hi, I haven't had time to look at the PR in detail yet. Just wanted to comment to acknowledge it. I will wait for this to be merged to update my other PRs :)

LiamPattinson:
Whoops, didn't mean to do that, sorry. Please feel free to undo that commit and rebase on `main` properly! I'd do it myself but that would give me credit over your PR.

Edit: Never mind, fixing it via rebase felt too dodgy so I just fixed the broken tests instead.

crates/fortitude_linter/src/diagnostics/mod.rs
```diff
@@ -106,197 +111,1495 @@ impl AddAssign for FixMap {
...
-        self.fix = Some(fix);
+    /// Adds an "info" sub-diagnostic with the given message.
+    ///
+    /// If callers want to add an "info" sub-diagnostic with annotations, then
+    /// create a [`SubDiagnostic`] manually and use [`Diagnostic::sub`] to
+    /// attach it to a parent diagnostic.
+    ///
+    /// An "info" diagnostic is useful when contextualizing or otherwise
+    /// helpful information can be added to help end users understand the
+    /// main diagnostic message better. For example, if a the main diagnostic
```
LiamPattinson:
```suggestion
    /// main diagnostic message better. For example, if the main diagnostic
```

crates/fortitude_linter/src/diagnostics/mod.rs
```diff
@@ -106,197 +111,1495 @@ impl AddAssign for FixMap {
...
+
+impl PartialEq<&str> for LintName {
+    fn eq(&self, other: &&str) -> bool {
+        self.0 == *other
+    }
+}
+
+/// Uniquely identifies the kind of a diagnostic.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, PartialOrd, Ord, Hash)]
+pub enum DiagnosticId {
```
LiamPattinson, line 988: I'm not sure if we strictly need all of these, but also can't see the harm in keeping them around just in case.

crates/fortitude_linter/src/diagnostics/mod.rs
```diff
@@ -106,197 +111,1495 @@ impl AddAssign for FixMap {
...
+            Severity::Warning => AnnotateLevel::Warning,
+            Severity::Error => AnnotateLevel::Error,
+            // NOTE: Should we really collapse this to "error"?
+            //
+            // After collapsing this, the snapshot tests seem to reveal that we
+            // don't currently have any *tests* with a `fatal` severity level.
+            // And maybe *rendering* this as just an `error` is fine. If we
+            // really do need different rendering, then I think we can add a
+            // `Level::Fatal`. ---AG
+            Severity::Fatal => AnnotateLevel::Error,
```
LiamPattinson, line 1202: I'm not sure if we need this level. What's the difference between this and `Error`?

LiamPattinson (commented):
Sorry for taking so long with this, there was a lot to get through! I have only very minor comments, so we should be good to merge very soon.

crates/fortitude_linter/src/diagnostics/mod.rs
ZedThree: There's definitely a few that I think would be useful to start using

crates/fortitude_linter/src/diagnostics/mod.rs
ZedThree: `Error` is the usual level for rules -- #651 would allow for turning some down to `Warning` instead.

`Fatal` is for internal errors: ruff/ty use it if they catch a `panic` or if applying a fix introduces a syntax error

LiamPattinson (approved):

PR #673, opened 2026-06-12
Add new ReturnInProgram rule.

### LINKED ISSUE #548: `return-in-program`
Some compilers let you `return` from a `program` body as an alternative to `stop`. This is non-standard, and should be a `portability` rule

avadean:
@LiamPattinson @ZedThree

**Add new ReturnInProgram rule.**
- Checks if a `return` statement is inside the body of a `program`. If so, will recommend to replace the statement with `stop`.
- Put in "Portability" rule with number 041.
- Allows `return` statement inside `function`, `subroutine`, `module`, `submodule`, `block_data`. (Latter 3 I don't think are possible but added as safeguards (?))

Closes #548

### DIFF
--- a/crates/fortitude_linter/src/rules/portability/return_in_program.rs
+++ b/crates/fortitude_linter/src/rules/portability/return_in_program.rs
@@ -1,0 +1,68 @@
+use crate::ast::FortitudeNode;
+use crate::diagnostics::{AlwaysFixableViolation, Diagnostic, Fix};
+use crate::{AstRule, CheckContext};
+use fortitude_macros::ViolationMetadata;
+use ruff_macros::derive_message_formats;
+use tree_sitter::Node;
+
+/// ## What it does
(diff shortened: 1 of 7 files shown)


crates/fortitude_linter/src/rules/portability/return_in_program.rs
```diff
@@ -0,0 +1,68 @@
...
+            .to_text(context.source_text())?
+            .eq_ignore_ascii_case("return")
+        {
+            return None;
+        }
+
+        for ancestor in node.ancestors() {
+            if matches!(
+                ancestor.kind(),
+                "function" | "subroutine" | "module" | "submodule" | "block_data"
```
ZedThree, line 55: Technically, `module` and `submodule` can't have executable statements, so we could safely exclude those, but we would then have to have `module_procedure` instead, so this is probably fine

ZedThree (approved):
LGTM, thanks @avadean !

PR #674, opened 2026-06-12
Refactor CLI test fixture

ZedThree:
Another one to reduce the size of the annotations PR.

- Refactor the CLI test fixture to take care of creating the temp directory and making file creation easier
- Refactor the filters to take out the temp directories more reliably on all three platforms
- Keep the filename in the filter, so we see `test.f90:4:3` instead of just `[TEMP_FILE]`
- Default to concise output to cut down the snapshots
- Add explicit CLI tests for the output formats

### DIFF
--- a/crates/fortitude/tests/check.rs
+++ b/crates/fortitude/tests/check.rs
@@ -1,115 +1,203 @@
+use anyhow::{Context, Result};
+use insta::internals::SettingsBindDropGuard;
 use insta_cmd::{assert_cmd_snapshot, get_cargo_bin};
 use std::path::{Path, PathBuf};
 use std::{fs, process::Command};
 use tempfile::TempDir;
+use textwrap::dedent;
 
 const BIN_NAME: &str = "fortitude";
+const STDIN_BASE_OPTIONS: &[&str] = &["--output-format", "concise"];
 
 fn fortitude_cmd() -> Command {
     Command::new(get_cargo_bin(BIN_NAME))
 }
 
-/// Builder for `fortitude check` commands
-///
-/// This is used to ensure we always set either:
-///   a) `--isolated`, or
-///   b) `--config-file`
-/// to make sure we don't get any interference from user-specific config files
-#[derive(Debug, Default)]
-struct FortitudeCheck<'a> {
-    config: Option<&'a Path>,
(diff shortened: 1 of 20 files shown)


ZedThree:
Currently failing on Windows:

```
        3       │-test/file/doesnt/exist.f90:1:1: E000 Error opening file: No such file or directory (os error 2)
              3 │+test\file\doesnt\exist.f90:1:1: E000 Error opening file: The system cannot find the path specified. (os error 3)
```

and I don't understand why _only_ this test is failing. The filters work on every other test

crates/fortitude/tests/check.rs
```diff
@@ -240,62 +304,27 @@ unknown-key = 1
 
 #[test]
 fn check_all() -> anyhow::Result<()> {
```
LiamPattinson: Could it be that this test should be running wit `--select=ALL`? I'm not sure how the test name lines up with the body.

crates/fortitude/tests/check.rs
```diff
@@ -2434,9 +1790,10 @@ fn preview_enabled_prefix() -> anyhow::Result<()> {
 #[test]
 fn preview_disabled_direct() -> anyhow::Result<()> {
     // All the FORT99XX test rules should be triggered
```
LiamPattinson: I think this comment is incorrect, probably a copy-paste error

LiamPattinson (dismissed):
Thanks, this is much cleaner, good to see a few hundred lines removed! I agree with using concise format in these tests, as the detailed output should be checked in the per-rule tests.

I've left two comments pointing at possible errors in the tests that are unrelated to the lines changed here.

crates/fortitude/tests/check.rs
ZedThree: Looks like it originally was testing just `fortitude check`, but then we added specific rules at some point, probably for stability. Given that this doesn't really test `ALL` rules (and is the wrong test for that!) and is basically just the same as the next test (`check_select_cli`), I'm just going to delete this one.

crates/fortitude/tests/check.rs
ZedThree: Yes, good catch, will fix

LiamPattinson (approved):

PR #675, opened 2026-06-15
Add rule `shadowed-variable`

### LINKED ISSUE #672: Rule: `shadowed-variable`
Using `SymbolTable`, it should be fairly straightforward to check if a variable has been declared in a higher scope. I've worked on a code before that had a few bugs caused by unintentional shadowing of module-scope variables, so I think this should fall under `correctness`.

I personally wouldn't want this to be switched on by default, as I frequently shadow variables on purpose as a way of indicating intent, e.g.:

```f90
subroutine sub(x, y, z)
  integer, intent(in) :: x
  integer, intent(inout) :: y
  integer, intent(out) :: z

  call helper(x, y)
  ...  

contains

  ! By calling the args x and y, it's clear which variables I intend to
  ! pass to this from the higher scope
  subroutine helper(x, y)
    integer, intent(in) :: x
    integer, intent(out) :: y
    ...
  end subroutine helper
end subroutine sub
```

LiamPattinson:
Closes https://github.com/PlasmaFAIR/fortitude/issues/672

Also upgrades some AST stuff along the way.

TODO:

- [x] Better testing, try uppercase vs lowercase and more obscure scenarios. 
- [x] Expand list of 'allowed integers', I've certainly missed a few.
- [x] Capture `use`'d variables in `SymbolTable` and check those.
  - We'll need to capture aliases too
- [x] Perhaps we should also check against non-variables in higher scopes?
- [x] Add user option:

```toml
[check.shadowed_variables]
allow = ["foo", "bar"]
```

### DIFF
--- a/crates/fortitude_linter/src/ast/symbol_table.rs
+++ b/crates/fortitude_linter/src/ast/symbol_table.rs
@@ -8,7 +8,8 @@ use crate::{ast::types::ProcedureKind, traits::HasNode};
 use super::{
     FortitudeNode,
     types::{
-        HasName, Module, Name, Procedure, Program, TypeDefinition, Variable, VariableDeclaration,
+        HasName, Module, Name, Procedure, Program, TypeDefinition, UseStatement, UsedItem,
+        Variable, VariableDeclaration,
     },
 };
 
@@ -39,6 +40,7 @@ pub enum Symbol<'a> {
     Type(TypeDefinition<'a>),
     Module(Module<'a>),
     Program(Program<'a>),
+    UsedItem(UsedItem<'a>),
 }
 
 impl<'a> Symbol<'a> {
@@ -49,6 +51,7 @@ impl<'a> Symbol<'a> {
             Self::Type(typedef) => typedef.name(),
             Self::Module(module) => module.name(),
             Self::Program(program) => program.name(),
+            Self::UsedItem(item) => item.name(),
         }
     }
 }
@@ -61,6 +64,7 @@ impl<'a> HasNode<'a> for Symbol<'a> {
             Self::Type(typedef) => typedef.node(),
             Self::Module(module) => module.node(),
             Self::Program(program) => program.node(),
+            Self::UsedItem(item) => item.node(),
         }
     }
 }
@@ -76,6 +80,7 @@ impl<'a> HasNode<'a> for Symbol<'a> {
 pub struct SymbolTable<'a> {
     inner: HashMap<String, Symbol<'a>>,
     decl_lines: Vec<Rc<VariableDeclaration<'a>>>,
+    use_statements: Vec<Rc<UseStatement<'a>>>,
 }
 
 impl<'a> SymbolTable<'a> {
@@ -104,6 +109,12 @@ impl<'a> SymbolTable<'a> {
                 }
             }
         }
+
+        scope
+            .named_children(&mut scope.walk())
+            .filter(|child| child.kind() == "use_statement")
+            .filter_map(|stmt| UseStatement::try_from_node(&stmt, src).ok())
+            .for_each(|stmt| new_table.insert_from_use_statement(stmt, src));
 
         scope
             .named_children(&mut scope.walk())
@@ -187,6 +198,22 @@ impl<'a> SymbolTable<'a> {
         self.decl_lines.push(decl);
     }
 
+    /// Insert all symbols found in a single use statement
+    pub fn insert_from_use_statement(&mut self, stmt: UseStatement<'a>, src: &str) {
+        let stmt = Rc::new(stmt);
+        if let Some(items) = stmt.included_items() {
+            for item in items.named_children(&mut items.walk()) {
+                // Other nodes such as comments can be found in the list
+                if let Some(item) = UsedItem::try_from_node(item, src, stmt.clone()) {
+                    let symbol = Symbol::UsedItem(item);
+                    let name = symbol.name().as_str().to_ascii_lowercase();
+                    self.inner.insert(name, symbol);
+                }
+            }
+        }
+        self.use_statements.push(stmt);
+    }
+
     /// Return the symbol with the given name if it exists
     pub fn get(&self, name: &str) -> Option<&Symbol<'_>> {
         self.inner.get(name)
@@ -195,6 +222,11 @@ impl<'a> SymbolTable<'a> {
     /// Iterator over the variable declaration lines
     pub fn iter_decl_lines(&self) -> impl Iterator<Item = &Rc<VariableDeclaration<'a>>> {
         self.decl_lines.iter()
+    }
+
+    /// Iterator over the use statements
+    pub fn iter_use_statements(&self) -> impl Iterator<Item = &Rc<UseStatement<'a>>> {
+        self.use_statements.iter()
     }
 
     /// Iterator over symbols in this scope
@@ -627,6 +659,85 @@ end program foo
     }
 
     #[test]
+    fn used_items() -> Result<()> {
+        let mut parser = Parser::new();
+        parser
+            .set_language(&tree_sitter_fortran::LANGUAGE.into())
+            .context("Error loading Fortran grammar")?;
+
+        let code = r#"
+program foo
+  use, intrinsic :: iso_fortran_env, only: i8 => int8, real32
+  use :: my_module, only: my_subroutine
+  use another_module
+end program foo
+"#;
+        let tree = parser.parse(code, None).context("Failed to parse")?;
+        let root = tree.root_node().child(0).context("Missing child")?;
+
+        let mut symbol_table = SymbolTables::default();
+        symbol_table.push_table(SymbolTable::new(&root, code));
+
+        let i8 = symbol_table.get("i8");
+        assert!(i8.is_some());
+        let i8 = i8.unwrap();
+        assert!(i8.is_used_item());
+        assert!(i8.name().as_str() == "i8");
+        let Symbol::UsedItem(i8) = i8 else {
+            panic!("Expected i8 to be a UsedItem");
+        };
+        assert!(i8.alias_of().is_some());
+        let i8_alias_of = i8.alias_of().unwrap();
+        assert!(i8_alias_of.as_str() == "int8");
+        let i8_module = i8.module_name();
+        assert!(i8_module.as_str() == "iso_fortran_env");
+
+        let real32 = symbol_table.get("real32");
+        assert!(real32.is_some());
+        let real32 = real32.unwrap();
+        assert!(real32.is_used_item());
+        assert!(real32.name().as_str() == "real32");
+        let Symbol::UsedItem(real32) = real32 else {
+            panic!("Expected real32 to be a UsedItem");
+        };
+        assert!(real32.alias_of().is_none());
+        let real32_module = real32.module_name();
+        assert!(real32_module.as_str() == "iso_fortran_env");
+
+        let iso_fortran_env = real32.decl_statement();
+        assert!(iso_fortran_env.name().as_str() == "iso_fortran_env");
+        assert!(iso_fortran_env.is_intrinsic());
+        assert!(iso_fortran_env.has_colon());
+        assert!(iso_fortran_env.has_only());
+
+        let my_subroutine = symbol_table.get("my_subroutine");
+        assert!(my_subroutine.is_some());
+        let my_subroutine = my_subroutine.unwrap();
+        assert!(my_subroutine.is_used_item());
+        let Symbol::UsedItem(my_subroutine) = my_subroutine else {
+            panic!("Expected my_subroutine to be a UsedItem");
+        };
+        assert!(my_subroutine.alias_of().is_none());
+        let my_subroutine_module = my_subroutine.module_name();
+        assert!(my_subroutine_module.as_str() == "my_module");
+
+        let my_module = my_subroutine.decl_statement();
+        assert!(my_module.name().as_str() == "my_module");
+        assert!(!my_module.is_intrinsic());
+        assert!(my_module.has_colon());
+        assert!(my_module.has_only());
+
+        let symbol_table = symbol_table.pop_table().unwrap();
+        let another_module = symbol_table.iter_use_statements().last().unwrap();
+        assert!(another_module.name().as_str() == "another_module");
+        assert!(!another_module.is_intrinsic());
+        assert!(!another_module.has_colon());
+        assert!(!another_module.has_only());
+
+        Ok(())
+    }
+
+    #[test]
     fn all_symbols() -> Result<()> {
         let mut parser = Parser::new();
         parser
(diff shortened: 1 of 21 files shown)


crates/fortitude_linter/src/ast/symbol_table.rs
```diff
@@ -84,22 +84,50 @@ impl<'a> SymbolTable<'a> {
...
             .filter_map(|decl| VariableDeclaration::try_from_node(&decl, src).ok())
-            .for_each(|line| new_table.insert_from_decl_line(line));
+            .for_each(|line| new_table.insert_from_decl_line(line, &dummy_vars));
 
         // The `function` statement itself _may_ also be the declaration line if
         // it has a type as a procedure attribute. If it doesn't, then it will
         // either have an explicit decl line, which is handled above, or it's
         // implicitly typed, which we don't currently handle here at all
-        if scope.is_named() && scope.kind() == "function" {
+        if scope.kind() == "function" {
```
ZedThree, line 131: I think I added `scope.is_named()` to guard against triggering on identifiers called `function`, but maybe that isn't actually necessary.

This might be a reason to use `.kind_id()`, but I recall that being a little tricky for some reason -- probably worth revisiting.

EDIT: Ahh, I see you've just moved it up the higher scope, sensible!

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
```diff
@@ -0,0 +1,145 @@
...
+/// ## Why is this bad?
+/// Shadowing variables can lead to confusion, as it can be unclear which
+/// variable is being referenced in a given context. Shadowing may be
+/// unintentional, which is a common source of bugs.
+///
+/// There are contexts in which shadowing is acceptable. For instance,
+/// if a procedure dummy argument has the same name as a module variable, this is acceptable,
+/// this indicates that the programmer likely intends to use the dummy argument
+/// in preference to the module variable. However, shadowing a module variable
+/// with a local variable is generally considered poor practice.
```
ZedThree: Nice, I think this is a very sensible exception! Might be worth having a setting like `strict` to turn this off though?

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
```diff
@@ -0,0 +1,145 @@
...
+///
+/// It is also very common to reuse loop variables such as `i`, `j`, and `k` or error
+/// flags such as `err` in different scopes. Fortitude will ignore integers with common
```
ZedThree: Also a very sensible exception! Maybe another setting, or have this under `strict` too?

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
```diff
@@ -0,0 +1,145 @@
...
+const ALLOWED_INTEGERS: &[&str] = &[
+    "i", "j", "k", "ii", "jj", "kk", "idx", "index", "err", "ierr", "ioerr", "stat", "iostat",
+    "status",
+];
+
+fn is_allowed_integer(var: &Variable) -> bool {
+    let t = var.type_();
+    let name = var.name().as_str().to_ascii_lowercase();
+    t.is_intrinsic()
+        && t.as_str().eq_ignore_ascii_case("integer")
```
ZedThree: Hmm, maybe we need an `Intrinsic` enum to make this easier?

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
```diff
@@ -0,0 +1,145 @@
...
+        && ALLOWED_INTEGERS.contains(&name.as_str())
+}
+
+impl Violation for ShadowedVariable {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::None;
+
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        let name = &self.name;
+        // TODO: Use new diagnostic annotations to point to higher scope variable
```
ZedThree, line 97: :+1:

crates/fortitude_linter/src/ast/symbol_table.rs
LiamPattinson: I was getting horrendous test failures for `function` nodes that didn't have a `child(0)`. It took me ages to work out it was trying to build a `SymbolTable` from a variable called `function`! Moving it up a level seems to fix it here too, but I'll have to remember to use `.is_named()` in other contexts. There are probably some rules that should be using it but currently aren't.

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
LiamPattinson: I think an `Intrinsic` enum would be really useful, as it could also catch things like kind and len. I haven't implemented it here, as I'm already touching quite a lot of other code, but I'd be happy to raise an issue.

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
LiamPattinson: Added in https://github.com/PlasmaFAIR/fortitude/pull/675/commits/d2f6b8904aa1a9208d0343cecfd407b358e0ea64. Funny to see `strict` mode return on a rule-by-rule basis, it used to be a global setting in a really early build!

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
LiamPattinson: Also added under https://github.com/PlasmaFAIR/fortitude/pull/675/commits/d2f6b8904aa1a9208d0343cecfd407b358e0ea64

LiamPattinson:
I think this is now ready for review. It would be good to know if there are other really common loop counter/error flag names out there. Can you think of any way we could poll the corpus for that?

ZedThree:
Here's the top 25 `do` counters:

```console
$ rg -t fortran -ioNI "^ *do +([a-z][a-z0-9_]*)" -r '$1' | tr '[:upper:]' '[:lower:]' | sort | uniq -c | sort -nr | head -n 25
 255728 i
 112044 j
  74477 k
  19174 n
  19049 while
  14253 ii
  13639 l
   9746 m
   7741 h2b
   7667 itime
   7105 p5b
   6752 h1b
   6002 p3b
   5917 p1b
   5823 istring
   5751 jj
   5598 h4b
   5564 i1
   5355 h9b
   5194 h10b
   4725 h8b
   4310 h7b
   4301 jk
   4016 h6b
   3993 h12b
```

(although `while` is not actually a counter, of course)

And for identifiers following `stat=`:

```console
$ rg -t fortran -ioNI "stat *= *([a-z][a-z0-9_]*)" -r '$1' | tr '[:upper:]' '[:lower:]' | sort | uniq -c | sort -nr | head -n 25
  16288 ierr
   8189 stat
   5533 ios
   4230 ier
   3773 info
   3403 iostat
   3385 status
   2676 istat
   2071 e_io
   2055 err
   1154 rc
    891 alloc_stat
    889 stats
    821 alloc
    820 elementinfo
    771 io_stat
    652 ierror
    642 alloc_status
    625 iores
    588 ist
    490 astat
    474 input_status
    469 izero
    428 k
    421 i
```

From looking at these, some of them are likely due to individual projects being large enough to affect the results, so take them with a pinch of salt.

crates/fortitude_linter/src/ast/symbol_table.rs
```diff
@@ -84,22 +89,56 @@ impl<'a> SymbolTable<'a> {
...
+                .expect("First child must be function/subroutine statement");
+            if let Some(params) = stmt.child_by_field_name("parameters") {
+                for child in params.named_children(&mut params.walk()) {
+                    if child.kind() == "identifier" {
+                        dummy_vars.push(
+                            child
+                                .to_text(src)
+                                .unwrap_or("<unknown>")
+                                .to_ascii_lowercase(),
+                        );
```
ZedThree: I know that this is bad habit of mine, but I'm sure there's a way to write this using `.map`, `.filter`, and `.collect` to turn this whole bit into a single statement so that we don't need to have `mut dummy_vars`.

Not something to actually worry about unless you're bored!

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
```diff
@@ -0,0 +1,187 @@
...
+/// See [check.shadowed-variables](../settings.md#checkshadowed-variables)
+#[derive(ViolationMetadata)]
+pub struct ShadowedVariable {
+    name: String,
+}
+
+const ALLOWED_INTEGERS: &[&str] = &[
+    "i", "j", "k", "ii", "jj", "kk", "idx", "index", "err", "ierr", "ioerr", "stat", "iostat",
+    "status",
+];
```
ZedThree: Hmm, I'm not overly keen on having this list be completely hardcoded. Currently it can be extended with `allow` or completely ignored with `strict`.

Is it going to be disgusting if we have:
- `allow`, with a default of `ALLOWED_INTEGERS`
- `extend-allow`, current behaviour of `allow`
- `strict`, which disables the dummy args exception

Ahh, except that's not quite right, because `allow` allows anything, and `ALLOWED_INTEGERS` only applies to `integer` dummy args.

We don't want an explosion of options, and I think the current behaviour is a really nice balance, but I do think it would be nice to be able to override this set specifically (for example, maybe a project wants to always use a global `status` for logging)

ZedThree (changes requested):
LGTM! Only real change that I think is necessary is some way to override `ALLOWED_INTEGERS` for dummy args (that is, not just appending using `allow`, but removing default allowed names)

crates/fortitude_linter/src/ast/symbol_table.rs
LiamPattinson: I think it's the `&mut params.walk()` in `named_children` that pushed me towards this implementation. Perhaps we can implement some way around it, or maybe something using `scan` would work?

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
LiamPattinson: What if we have two lists?

- `allow` 
- `allow_integer_only`

The general `allow` will ignore typing, while `allow_int_only` will additionally check that both the shadower and shadowee are integers, and take the current list of allowed ints (plus others you've found) as a default.

LiamPattinson:
I've simplified the implementation. Rather than checking for specific integers, it now just has a general whitelist, and users are free to overwrite this with their own preferences. Strict mode now doesn't overrule the whitelist, but does target dummy variables. It might be a good idea to rename that option. How about `allow_dummy_args`, and set it to be true by default?

crates/fortitude_linter/src/ast/types.rs
```diff
@@ -833,3 +845,127 @@ impl<'a> Program<'a> {
...
+        let has_only = node.child_with_name("included_items").is_some();
+
+        let mut intrinsic = false;
+        let mut has_colon = false;
+        for child in node.children(&mut node.walk()) {
+            if child.kind() == "intrinsic" {
+                intrinsic = true;
+            }
+            if child.kind() == "::" {
+                has_colon = true;
```
ZedThree, line 882: Could add
```suggestion
                has_colon = true;
                break;
```
here because this should always appear after `intrinsic`, so there's no need to check any more children?

crates/fortitude_linter/src/rules/correctness/shadowed_variable.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// ## What it does
+/// Checks for variables that shadow those in a higher scope.
+///
+/// ## Why is this bad?
+/// Shadowing variables can lead to confusion, as it can be unclear which
+/// variable is being referenced in a given context. Shadowing may be
+/// unintentional, which is a common source of bugs.
+///
+/// There are contexts in which shadowing is acceptable. For instance, if a
```
ZedThree:
```suggestion
/// There are contexts in which shadowing may be acceptable. For instance, if a
```

crates/fortitude_workspace/src/options.rs
```diff
@@ -448,6 +452,36 @@ impl ExitUnlabelledLoopOptions {
...
+    /// common loop variables and error flags are allowed to be shadowed.
+    /// Variables in this list will be ignored in addition to the default list.
+    #[option(
+        default = r#"["i", "j", "k", "l", "m", "n", "ii", "jj", "kk", "ll", "mm", "nn", "idx", "index", "err", "ierr", "ioerr", "ios", "info", "stat", "iostat", "istat", "status"]"#,
+        value_type = "list[str]",
+        example = r#"allow = ["array", "x"]"#
+    )]
+    pub allow: Option<Vec<String>>,
+
+    /// Strict mode will also include dummy arguments.
```
ZedThree: I think this is slightly ambiguously worded, given that it will appear immediately after `allow`? Maybe:
```suggestion
    /// Strict mode will also check dummy arguments for violations.
```

crates/fortitude_workspace/src/options.rs
```diff
@@ -448,6 +452,36 @@ impl ExitUnlabelledLoopOptions {
...
 
+/// Options for the `shadowed-variable` rule
+#[derive(
+    Clone, Debug, PartialEq, Eq, Default, OptionsMetadata, CombineOptions, Serialize, Deserialize,
+)]
+#[serde(deny_unknown_fields, rename_all = "kebab-case")]
+pub struct ShadowedVariableOptions {
+    /// A list of variable names that are allowed to be shadowed. By default,
+    /// common loop variables and error flags are allowed to be shadowed.
+    /// Variables in this list will be ignored in addition to the default list.
```
ZedThree: This is no longer true
```suggestion
```

ZedThree (changes requested):
LGTM! Just some nitpicking on the docs

crates/fortitude_linter/src/ast/symbol_table.rs
ZedThree: That's not a problem because there's no reference to it that escapes. I had a crack at it for fun:

```rust
        let dummy_vars = if matches!(scope.kind(), "function" | "subroutine") {
            scope
                .named_child(0)
                .expect("First child must be function/subroutine statement")
                .child_by_field_name("parameters")
                .map_or(vec![], |params| {
                    params
                        .named_children(&mut params.walk())
                        .filter(|child| child.kind() == "identifier")
                        .map(|child| {
                            child
                                .to_text(src)
                                .unwrap_or("<unknown>")
                                .to_ascii_lowercase()
                        })
                        .collect_vec()
                })
        } else {
            vec![]
        };
```
I'll let you decide if that's actually worth it or not!

LiamPattinson: Added in https://github.com/PlasmaFAIR/fortitude/pull/675/commits/baf10bbd270b59c168f56449e9bebaf257e57d9f, thanks!

LiamPattinson:
Thanks @ZedThree , I've added your suggestions in https://github.com/PlasmaFAIR/fortitude/pull/675/commits/52d694f01b5b37e659ef528ef9be44a545ebe91b

ZedThree (approved):
LGTM!

Issue #472, opened 2025-06-17
Clarify criteria for fixes being 'unsafe'


LiamPattinson:
Following on from a discussion in https://github.com/PlasmaFAIR/fortitude/pull/471#discussion_r2149905809, we haven't been following a strict rule about 'safe' vs 'unsafe' fixes. The argument in that PR is that a fix should not be considered safe if there's a chance that the user has written their code in error. For example, the fix to add parentheses to `a * b / c * d` to improve readability would be something like `(a * b / c) * d`, but the user might have intended `(a * b) / (c * d)`, and therefore we shouldn't fix it automatically.

I had previously been working under the assumption that a fix should be considered safe only if it doesn't change the behaviour of the code. This means that rules like `missing-intrinsic` is considered unsafe, as the user might have created their own module called `iso_fortran_env` as a drop-in replacement for the intrinsic module, even though that's very unlikely to occur in practice. Meanwhile, `misleading-inline-if-statement` is considered safe, as even though it's very likely that the user has made an error, the fix doesn't change the code's actual behaviour in any way.

The opposite interpretation is taken by `split-escaped-quote`, which is considered a safe fix but does change the behaviour of the code. This was justified at the time as being almost certainly a bug.

I think we should set some explicit rules regarding when a fix is or isn't safe, and then ensure they're being applied consistently. I personally lean towards the idea that safeness should only depend on changes to code behaviour, and so the new rule in #471 should be considered safe and `split-escaped-quote` should be changed to unsafe -- although this is primarily just because it's easier to apply that interpretation consistently.

ZedThree:
Ruff's docs on fix safety: https://docs.astral.sh/ruff/linter/#fix-safety

TL;DR: safe fixes don't impact runtime behaviour, unsafe fixes might.

I agree this is the most reasonable distinction, although we have the added complication of compile-time behaviour -- adding `implicit none` might cause a compile error. This is probably fine though, because runtime behaviour wouldn't be impacted.

PR #679, opened 2026-06-18
Add rule for redundant whole-array indexing

### LINKED ISSUE #678: Avoid whole array indexing with `(:)`
[From Doctor Fortran](https://stevelionel.com/drfortran/2008/03/31/doctor-it-hurts-when-i-do-this/):

> So, Doctor Fortran’s advice if you put `(:)` on your arrays? Don’t do that!

It almost certainly doesn't hurt anything, apart from making the compiler work harder, so this is most likely a `style` rule

koriyoshi2041:
## Summary
- add preview style rule S312 for redundant whole-array indexing
- flag and safely fix whole-array references like `x(:)`, `y(:, :)`, and `item%values(:)`
- cover meaningful slices like `1:`, `:5`, `::2`, and `1:5` as non-violations

Fixes #678.

## Tests
- `INSTA_UPDATE=always cargo test -p fortitude_linter rules::style::tests::rules -- --nocapture`
- `cargo clippy -p fortitude_linter --all-targets -- -D warnings`
- `target/debug/fortitude check --preview --select S312 --diff crates/fortitude_linter/resources/test/fixtures/style/S312.f90`


crates/fortitude_linter/src/rules/style/whole_array_indexing.rs
```diff
@@ -0,0 +1,75 @@
...
+/// ```f90
+/// x = x(:)
+/// y = y(:, :)
+/// ```
+///
+/// Use instead:
+/// ```f90
+/// x = x
+/// y = y
+/// ```
```
ZedThree, line 31: Please could you also add https://stevelionel.com/drfortran/2008/03/31/doctor-it-hurts-when-i-do-this/ to a `## References` section here?

crates/fortitude_linter/src/rules/style/whole_array_indexing.rs
```diff
@@ -0,0 +1,75 @@
...
+    let Some(first) = arguments.next() else {
+        return false;
+    };
+
+    first.kind() == "extent_specifier"
+        && first.to_text(src).is_some_and(|text| text.trim() == ":")
+        && arguments.all(|argument| {
+            argument.kind() == "extent_specifier"
+                && argument.to_text(src).is_some_and(|text| text.trim() == ":")
+        })
```
ZedThree: We could avoid the repetition here with something like:

```rust
    let is_extent_specifier = |arg: &Node| { 
        arg.kind() == "extent_specifier"
            && arg.to_text(src).is_some_and(|text| text.trim() == ":")
    };

    is_extent_specifier(first) && arguments.all(is_extent_specifier)
```

I'm not sure it makes a great deal of difference though

ZedThree (changes requested):
Thanks for the PR @koriyoshi2041, it's much appreciated! Please could you add the reference to the docstring and then run `cargo dev generate-all` to generate the markdown docs?

koriyoshi2041:
Thanks, added the reference from the issue docstring and generated the rule docs in 737d5ba.

Local checks:
- cargo dev generate-all --mode=check
- INSTA_UPDATE=always cargo test -p fortitude_linter rules::style::tests::rules -- --nocapture
- git diff --check

crates/fortitude_linter/src/rules/mod.rs
```diff
@@ -165,6 +165,7 @@ pub fn code_to_rule(category: Category, code: &str) -> Option<(RuleGroup, Rule)>
         (Style, "291") => (RuleGroup::Preview, Ast, Default, style::literals::BareDecimal),
         (Style, "301") => (RuleGroup::Preview, Ast, Optional, style::superfluous_while_true::SuperfluousWhileTrue),
         (Style, "311") => (RuleGroup::Preview, Ast, Default, style::keywords::KeywordReuse),
+        (Style, "312") => (RuleGroup::Preview, Ast, Optional, style::whole_array_indexing::WholeArrayIndexing),
```
ZedThree: I'm afraid our numbering system is a bit strange. Please could you just change this to `321` instead, and then re-generate the docs with `cargo dev generate-all`?

```suggestion
        (Style, "321") => (RuleGroup::Preview, Ast, Optional, style::whole_array_indexing::WholeArrayIndexing),
```

ZedThree:
I've tested this out on some real world code and it works a treat

crates/fortitude_linter/src/rules/style/whole_array_indexing.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// ## References
+/// - [Doctor Fortran: "Doctor, it hurts when I do this!"](https://stevelionel.com/drfortran/2008/03/31/doctor-it-hurts-when-i-do-this/)
+#[derive(ViolationMetadata)]
+pub(crate) struct WholeArrayIndexing;
+
+impl AlwaysFixableViolation for WholeArrayIndexing {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Avoid redundant whole-array indexing".to_string()
```
LiamPattinson: When writing error messages, we normally omit words like 'avoid' following our [naming guide](https://github.com/PlasmaFAIR/fortitude/blob/main/CONTRIBUTING.md#naming-and-categorising-rules).


```suggestion
        "Redundant whole-array indexing".to_string()
```

crates/fortitude_linter/src/rules/style/whole_array_indexing.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    fn fix_title(&self) -> String {
+        "Remove redundant whole-array indexing".to_string()
```
LiamPattinson: To provide a better fix hint, we could capture the variable name in the `Violation` struct, and write this as:


```suggestion
        let name = self.name;
        format!("Write as `{name}`");
```

koriyoshi2041:
Thanks for testing it on real-world code. The requested reference/docs update is in the latest head, and the PR is green across ReadTheDocs, rustfmt, and the full platform test matrix now.

ZedThree (changes requested):
Very sorry, I missed a comment off from the last review. I think it's done after this

crates/fortitude_linter/resources/test/fixtures/style/S321.f90
LiamPattinson, line 1: Please could we add some examples where `x(:)` is on the left-hand-side of an equation?

LiamPattinson (commented):
Hi @koriyoshi2041, thanks for your contribution! This looks good to me, and beyond @ZedThree's comments I only have a few minor suggestions of my own.

imciner2:
I don't think a fix for the LHS of an expression would be a safe fix. Based on that Doctor Fortran article, it would affect the runtime behavior of the program because now there would be no reallocation, and I think based on https://github.com/PlasmaFAIR/fortitude/issues/472 safe fixes shouldn't impact runtime behavior. The RHS part would be a safe fix though I think.

LiamPattinson:
Maybe the solution could be to ensure the array is on the RHS of an assignment or passed to a subroutine in this rule, and to add a separate `correctness` rule that looks at the left hand side of an assignment.

koriyoshi2041:
Thanks, that makes sense. I pushed 2b440e1 to skip this rule on the left-hand side of assignment statements, so it still fixes RHS whole-array references and subroutine arguments but leaves cases like `x(:) = y` alone.\n\nI also added fixture coverage for both cases. Local checks:\n- `INSTA_UPDATE=always cargo test -p fortitude_linter rules::style::tests::rules -- --nocapture`\n- `cargo fmt --check`\n- `cargo dev generate-all --mode=check`\n- `cargo clippy -p fortitude_linter --all-targets -- -D warnings`\n- `target/debug/fortitude check --preview --select S312 --diff crates/fortitude_linter/resources/test/fixtures/style/S312.f90`

ZedThree:
Well, the lhs is a little bit more complicated, because it matters if the variable is `allocatable` or not. If it's not, then this should still be a safe fix. If it _is_ `allocatable`, then it could still be an unsafe fix, but I suppose it's also possible the programmer used it on purpose to disable reallocation.

Unfortunately, we can't 100% detect attributes on variables, because they may come from other modules (a separate violation!), so disabling the fix entirely seems reasonable. We should add something about this to the docstring though.

Maybe we need a separate rule for `prevented-reallocation` or similar. That would let developers use `allow(prevented-reallocation)` to disable it and also document why

koriyoshi2041:
Thanks, agreed. I added that caveat to the rule docstring in b42e962, and regenerated the rule docs so it shows up in `docs/rules/whole-array-indexing.md` too.

Checks run:
- `cargo fmt --check`
- `cargo dev generate-all`
- `cargo dev generate-all --mode=check`
- `INSTA_UPDATE=always cargo test -p fortitude_linter rules::style::tests::rules -- --nocapture`

koriyoshi2041:
Quick follow-up after the last review round: the requested reference/docs generation, rule number rename, message/hint cleanup, and LHS assignment fixture/caveat updates are all in the current head.

The PR is green across rustfmt, platform tests, cargo install, pip install, and docs. I think it is ready for another look when you have a chance.

LiamPattinson:
Hi @koriyoshi2041, there are still some comments I left that haven't yet been addressed:

- https://github.com/PlasmaFAIR/fortitude/pull/679#discussion_r3441905136
- https://github.com/PlasmaFAIR/fortitude/pull/679#discussion_r3441911819
- https://github.com/PlasmaFAIR/fortitude/pull/679#discussion_r3441939470

You'll also need to rebase on `main` and update the entrypoints function, as we've updated that system -- sorry for the clash! It should be as simple as this:

```rust
    fn entrypoints() -> Vec<u16> {
        kind_ids!["call_expression"]
    }
```

koriyoshi2041:
Thanks for the pointers. I rebased on current main and pushed the review follow-up:

- updated the rule code/docs/fixture from S312 to S321
- switched `entrypoints()` to `kind_ids!["call_expression"]`
- adjusted the diagnostic message and fix title to include the replacement name
- simplified the extent check and kept the LHS assignment coverage

Local checks:
- `cargo fmt --check`
- `cargo dev generate-all --mode=check`
- `INSTA_UPDATE=always cargo test -p fortitude_linter rules::style::tests::rules -- --nocapture`
- `cargo clippy -p fortitude_linter --all-targets -- -D warnings`
- `git diff --check`

ZedThree (approved):
Lovely, thanks @koriyoshi2041!

Issue #670, opened 2026-06-12
fortitude check doesnot respect rules defined in .fotitude.toml or pyproject.toml


yaswant:
Example repository: <https://github.com/MetOffice/lfric_apps>

## zero configuration

```bash
fortitude check . --statistics
Discarding all fixes. Some violations from the AST may be unreliable.
2104    C082    [ ] pointer-initialisation-in-declaration
 578    S241    [*] bad-quote-string
 417    C003    [ ] implicit-external-procedures
 397    C031    [ ] magic-number-in-array-size
 397    S001    [ ] line-too-long
 115    MOD011  [*] old-style-array-literal
  56    C081    [ ] initialisation-in-declaration
  49    S231    [*] keywords-missing-space
  47    C011    [ ] missing-default-case
  42    C152    [*] misleading-inline-if-continuation
  26    S061    [*] unnamed-end-statement
  19    C121    [ ] use-all
  16    C043    [ ] missing-action-specifier
  15    C131    [ ] missing-accessibility-statement
   7    S101    [*] trailing-whitespace
   6    C002    [ ] interface-implicit-typing
   5    C072    [ ] assumed-size-character-intent
   4    MOD021  [*] deprecated-relational-operator
   3    S002    [*] missing-newline-at-end-of-file
   2    C051    [ ] trailing-backslash
   2    S261    [ ] inconsistent-array-declaration
fortitude: 1821 files scanned.
Number of errors: 4307
```

## `.fortitude.toml` ❌
```toml
[check]
ignore = ["C082",]
```

```
fortitude check . --statistics

2086    C082    [ ] pointer-initialisation-in-declaration
 578    S241    [*] bad-quote-string
 417    C003    [ ] implicit-external-procedures
 397    C031    [ ] magic-number-in-array-size
 397    S001    [ ] line-too-long
 115    MOD011  [*] old-style-array-literal
  56    C081    [ ] initialisation-in-declaration
  49    S231    [*] keywords-missing-space
  47    C011    [ ] missing-default-case
  42    C152    [*] misleading-inline-if-continuation
  26    S061    [*] unnamed-end-statement
  19    C121    [ ] use-all
  16    C043    [ ] missing-action-specifier
  15    C131    [ ] missing-accessibility-statement
   7    S101    [*] trailing-whitespace
   6    C002    [ ] interface-implicit-typing
   5    C072    [ ] assumed-size-character-intent
   4    MOD021  [*] deprecated-relational-operator
[...]

PR #680, opened 2026-06-19
Clarify nested configuration discovery

koriyoshi2041:
## Summary
- clarify that nested configuration files do not inherit or merge parent configuration settings
- call out that rule selection settings like `select` and `ignore` come from the closest config under automatic discovery
- distinguish this from `--config-file`, which applies one config to all files

Related to #670.

## Tests
- `git diff --check`


ZedThree (approved):
Thanks @koriyoshi2041!

PR #684, opened 2026-06-25
Compile time checking of node names

ZedThree:
Currently we rely on `Node::kind()` and string comparisons. There's a couple of instances where we need to distinguish between named and unnamed nodes with the same kind (the named `function` node is the whole function, whereas the unnamed `function` is just the keyword). 

This PR lets us start switching to using `Node::kind_id() -> u16`. Unfortunately, tree-sitter doesn't auto-generate any kind of mapping. There is a runtime function, `Language::id_for_node_kind()` that does this, but it would be quite messy to use it everywhere, plus we can't use it in e.g. `match` expressions (because it's not a constant).

This PR provides three proc macros, `kind!()`, `kw!()`, `field!()`, which do this look-up at compile-time, so we get compile-time checking of node names, plus the ability to use them in constant expressions. There may even be some performance benefit to not doing string comparisons everywhere.

We can also use these to avoid case-insensitive comparisons:

```diff
     node.children(&mut node.walk())
-        .find(|child| child.to_text(src.source_text()).unwrap().to_lowercase() == "external")
+        .find(|child| child.kind_id() == kw!("external"))
```

I've made a few changes like this to give a feel of how ergonomic these macros are to use.

The biggest change is for `AstRule::entrypoints()` to return `Vec<u16>`, using a macro `kind_ids![]` to make this simple. Use `"node_name" | kw` to get an unnamed node:

```rust
kind_ids![
    "function",          // named node
    "function" | kw,     // unnamed node, such as a keyword
    "math_expression",
    "+" | kw,
    "end_program_statement",
    "end" | kw,
];
```

### DIFF
--- a/crates/fortitude_linter/src/rules/mod.rs
+++ b/crates/fortitude_linter/src/rules/mod.rs
@@ -53,6 +53,41 @@ impl PartialEq<&str> for NoqaCode {
     }
 }
 
+/// Use `kind_ids!` in `AstRule::entrypoints` to convert a `Vec` of string
+/// literals into a `Vec` of compiletime kind IDs. For unnamed keywords, append
+/// ``| kw`` to the literal.
+///
+/// # Example
+/// ```
+/// use fortitude_linter::kind_ids;
+/// use fortitude_macros::{kind, kw};
+///
+/// fn main() {
+///     let ids = kind_ids![
+///         "function",
+///         "end" | kw
+///     ];
+///     let expected = vec![kind!("function"), kw!("end")];
+///     assert_eq!(ids, expected);
+/// }
+/// ```
+#[macro_export]
+macro_rules! kind_ids {
+    ($($kinds:literal $(| $modifier:tt)?),* $(,)?) => {
+        vec![
+            $(
+                kind_ids!(@kind $kinds $(| $modifier)?),
+            )*
+        ]
+    };
+    (@kind $kind:literal) => {
+        fortitude_macros::kind!($kind)
+    };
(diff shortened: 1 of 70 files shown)


crates/fortitude_linter/src/rules/style/keywords.rs
```diff
@@ -163,7 +164,7 @@ impl AstRule for KeywordsMissingSpace {
             "end_subroutine_statement",
             "end_type_statement",
             "end_where_statement",
-            "inout",
+            "inout" | kw,
```
LiamPattinson, line 167: I feel this rule could be rewritten to use keywords only. Not for this PR, but maybe worth raising as an issue afterwards.

crates/fortitude_linter/src/rules/mod.rs
```diff
@@ -53,6 +53,41 @@ impl PartialEq<&str> for NoqaCode {
...
+///     let ids = kind_ids![
+///         "function",
+///         "end" | kw
+///     ];
+///     let expected = vec![kind!("function"), kw!("end")];
+///     assert_eq!(ids, expected);
+/// }
+/// ```
+#[macro_export]
+macro_rules! kind_ids {
```
LiamPattinson, line 75: Loving this macro, it does exactly what I'd hoped it would do.

LiamPattinson (commented):
Awesome, I love this change! I've been wanting to do something like this for a long time, but always thought it would be too challenging. I'm surprised by how elegant the solution is!

LiamPattinson (approved):

PR #686, opened 2026-06-25
Add new unary-following-arithmetic rule.

### LINKED ISSUE #495: Unary negation following arithmetic operator is compiler dependent
https://community.intel.com/t5/Intel-Fortran-Compiler/Negative-exponential-should-be-compiler-error-instead-gives/td-p/1710610

```
> gfortran negative_exponent.f90
negative_exponent.f90:2:37:

    2 |   write(*,*) '10.0**-2 * 2=', 10.0**-2 * 2
      |                                     1
Warning: Extension: Unary operator following arithmetic operator (use parentheses) at (1)
> a.out
 10.0**-2 * 2=   1.99999996E-02

> nvfortran negative_exponent.f90
> a.out 
 10.0**-2 * 2=   2.0000000E-02

> ifx negative_exponent.f90
> a.out
 10.0**-2 * 2=  9.9999997E-05
```

I think this is just catching that `unary_expression` with a parent of `unary_expression` is non-standard and should use parentheses

avadean:
**Add new unary-following-arithmetic rule.**

- Checks if a `unary_expression` (e.g negating a number `-3`) follows a `math_expression` (or alternatively named arithmetic expression) (e.g adding two numbers `2+3`). For example, `2 + -3` should be replaced with `2 + (-3)`.
- Works with all `unary_expression` and `math_expression` operator possibilities ~including user-defined~.

Have not implemented as `AlwaysFixable` as I think(?) it can be ambigious what the user would want.
E.g do we replace `10 ** -2 * 3 - 5` with A) `10 ** (-2) * 3 - 5` or B) `10 ** (-2 * 3) - 5` or C) `10 ** (-2 * 3 - 5)`? We could default to showing some unsafe fix where we parenthesise the first expression only (option A)? Or parenthesise everything (option B)?

Closes #495.

### DIFF
--- a/crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
+++ b/crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
@@ -1,0 +1,156 @@
+use crate::diagnostics::{Diagnostic, FixAvailability, Violation};
+use crate::traits::TextRanged;
+use crate::{AstRule, CheckContext, kind_ids};
+use fortitude_macros::ViolationMetadata;
+use ruff_diagnostics::{Edit, Fix};
+use ruff_macros::derive_message_formats;
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for use of a unary expression following an arithmetic operator.
+///
+/// ## Why is this bad?
+/// The use of a unary operator (`+`, `-` but *not* user-defined) following an arithmetic operator (
+/// `+`, `-`, `*`, `/`, `**` but *not* user-defined) is not part of the Fortran standard. In some
+/// cases, this may be ambiguous as the order of operations does not necessarily follow typical
+/// mathematical order. The computed output may therefore vary between compilers.
+///
+/// Many compilers, Intel Fortran for example, will let you do this as an extension; sometimes with
+/// a warning on compilation. Use parentheses to remove ambiguity of user expected output; the fix
+/// may change prior behaviour on some compilers.
+///
+/// ## Example
+/// ```f90
+/// x = 10 ** -2 * 2
+/// ! Would expected x = 0.02 but some compilers may give x = 0.0001.
+/// ```
+///
+/// Use instead:
+/// ```f90
+/// x = 10 ** (-2) * 2
+/// ! Result is unambiguously x = 0.02.
+/// ```
+///
+/// ## References
+/// * [Doctor Fortran in "Order! Order!"](https://stevelionel.com/drfortran/2021/04/03/doctor-fortran-in-order-order/)
+#[derive(ViolationMetadata)]
+pub(crate) struct UnaryFollowingArithmetic;
+
+impl Violation for UnaryFollowingArithmetic {
+    const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes;
+
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Unary operator following an arithmetic expression".to_string()
+    }
+
+    fn fix_title(&self) -> Option<String> {
+        Some("Add parenthesese around the desired argument of the unary operator".to_string())
+    }
+}
+
+impl AstRule for UnaryFollowingArithmetic {
+    fn check(context: &CheckContext, node: &Node) -> Option<Vec<Diagnostic>> {
+        // Check if we have a unary expression on the RHS of this math expression.
+        let unary_exp = node.child_by_field_name("right")?;
+
+        if unary_exp.kind() != "unary_expression" {
+            return None;
+        }
+
+        // User defined unary operators have precedence so don't cause problems.
+        let unary_op = unary_exp.child_by_field_name("operator")?;
+
+        if unary_op.kind() == "user_defined_operator" {
+            return None;
+        }
+
+        // If for some strange reason we have multiple unary operators, e.g `10 ** -+2`, then we
+        // need the right-most one so its argument can be the only thing parenthesised.
+        let unary_arg = right_most_unary(unary_exp).child_by_field_name("argument")?;
+
+        let parenthesise_end_node = if unary_arg.kind() == "math_expression" {
+            unary_arg.child_by_field_name("left")?
+        } else {
+            unary_arg
+        };
+
+        // All fixes are safe except for those of the form `a ** -b * c` where the `-` can also be a
+        // `+` and the `*` can also be a `/`.
+        let fix = if is_unsafe_fix(node, &unary_arg, &unary_op) {
+            Fix::unsafe_edits(
+                Edit::insertion("(".to_string(), unary_exp.start_textsize()),
+                [Edit::insertion(
+                    ")".to_string(),
+                    parenthesise_end_node.end_textsize(),
+                )],
+            )
+        } else {
+            Fix::safe_edits(
+                Edit::insertion("(".to_string(), unary_exp.start_textsize()),
+                [Edit::insertion(
+                    ")".to_string(),
+                    parenthesise_end_node.end_textsize(),
+                )],
+            )
+        };
+
+        some_vec![
+            context
+                .create_diagnostic(UnaryFollowingArithmetic, unary_op)
+                .with_fix(fix)
+        ]
+    }
+
+    fn entrypoints() -> Vec<u16> {
+        kind_ids!["math_expression"]
+    }
+}
+
+fn is_unsafe_fix(node: &Node, unary_arg: &Node, unary_op: &Node) -> bool {
+    // An unsafe fix is when we have something of the form `a ** -b * c` where the `-` can also be a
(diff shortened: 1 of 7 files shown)


crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
```diff
@@ -0,0 +1,55 @@
...
+/// x = 10 ** (-2) * 2
+/// ! Result is unambiguously x = 0.02.
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct UnaryFollowingArithmetic;
+
+impl Violation for UnaryFollowingArithmetic {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Unary operator following arithmetic operator (use parentheses)".to_string()
```
LiamPattinson: Rather than adding `(use parentheses)` here, you can implement a function `fix_title` that gives advice on how to correct the problem.

crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
```diff
@@ -0,0 +1,55 @@
...
+        let rhs = node.child_by_field_name("right")?;
+
+        if rhs.kind() != "unary_expression" {
+            return None;
+        }
+
+        let unary_operator = rhs.child(0)?;
+
+        some_vec![context.create_diagnostic(UnaryFollowingArithmetic, unary_operator)]
+    }
```
LiamPattinson, line 130: Nice, this is a really clean implementation!

LiamPattinson (changes requested):
Thanks @avadean, this looks really good! There's only one small change that I think will be needed.

I'm also not 100% sure of the name for this one, but naming rules is always really hard. All I've got is `ambiguous-unary-operator`, but I'm not sure that's much better.

ZedThree:
We could use `Fix::display_only_edit` here, which should just show the edit as a suggestion without actually applying it as a fix. 

In most cases this should really be a safe fix, I think it's only when we have something like `x ** -y * z` where there's a real non-portability issue due to the combination of a) exponentiation have right-to-left associativity vs multiplication's left-to-right, and b) the extra `math_expression` to the right of the exponent. `x ** -y` should always be fine because there's actually only one possible way to evaluate that. Probably we do still want to trigger on that though, because gfortran raises a warning. We could have a safe fix for those cases, and an unsafe fix for the `x ** -y * z` cases.

Also, I think the user-defined unary operators might also be fine, because they have the highest precedence. I've tweaked the test file and put it into [godbolt](https://godbolt.org/z/e3z5Gb34E) to compare different compilers, and indeed for `ifx`, the version with `.negate.` is the same as `gfortran`'s answer.

crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
```diff
@@ -0,0 +1,55 @@
...
+/// ```f90
+/// x = 10 ** -2 * 2
+/// ! Would expected x = 0.02 but some compilers may give x = 0.0001.
+/// ```
+///
+/// Use instead:
+/// ```f90
+/// x = 10 ** (-2) * 2
+/// ! Result is unambiguously x = 0.02.
+/// ```
```
ZedThree, line 32: Please could you add a reference to the [Doctor Fortran](https://stevelionel.com/drfortran/2021/04/03/doctor-fortran-in-order-order/) article?

ZedThree:
@avadean We've changed the output format in `main`, as well as the return type of `entrypoints`, so you'll need to fix those

avadean:
@ZedThree thanks for the comments!

gfortran will throw a warning for _any_ unary expression that follows an arithmetic operator. So for example, `2 + -3` will throw a warning because the unary operator `-` follows the addition `+`. Fix is `2 + (-3)`.

@LiamPattinson and I had a chat and we think we should throw a warning if gfortran does, even in the trivial example above.

I agree that all of these cases there is a safe fix, except for the `a ** -b * c`, which I can implement an unsafe fix.

So course of action:
1) Throw warning for any unary operator following math expression. The code I've committed does this and it follows gfortran exactly.
2) Safe fix for everything except the `a ** -b * c` case where we do unsafe fix of `a ** (-b) * c`.

Sound ok? :)

ZedThree:
> 1. Throw warning for any unary operator following math expression. The code I've committed does this and it follows gfortran exactly.

I don't think gfortran triggers on user defined operators though? I think the fix is as simple as just check for literal `+`/`-`

ZedThree:
> 2\. Safe fix for everything except the `a ** -b * c` case where we do unsafe fix of `a ** (-b) * c`.

Yes, this is reasonable. Maybe we just want to make the language in the docstring even stronger, and really hammer home the point that the fix may change behaviour on certain compilers?

avadean:
> > 1. Throw warning for any unary operator following math expression. The code I've committed does this and it follows gfortran exactly.
> 
> I don't think gfortran triggers on user defined operators though? I think the fix is as simple as just check for literal `+`/`-`

Yes you are right, sorry forgot about that!

avadean:
@LiamPattinson @ZedThree 

**Changes implemented**
- Raises warning if unary operator follows a math expression i.e if `gfortran` has an issue then so do we!
- Does _not_ raise on user-defined unary operators.
- Generally have a safe fix except unsafe fix for the case of `a ** -b * c` (`-` can also be `+` and `*` can also be `/`).
- Added [Doctor Fortran article](https://stevelionel.com/drfortran/2021/04/03/doctor-fortran-in-order-order/).


**Outstanding issues/questions**
1. Have not changed the name. @LiamPattinson suggested `ambiguous-unary-operator`. What do we think? Most of the time we are dealing with the unary operator positioning not being part of the Fortran standard, _not_ ambiguity. For example, `a + -b` is not ambiguous but is technically not standard. The only time we have both lack of standard and an ambiguity is `a ** -b * c`.
2. Have changed the docstring only slightly, @ZedThree did you have any or additions/changes in mind?
3. I've not used `AlwaysFixable`. Should it be? It _is_ always fixable but not necessarily always a safe fix. With that in mind, I suppose `const FIX_AVAILABILITY: FixAvailability = FixAvailability::Sometimes` is not quite accurate(?). Please advise :)
4. Kinda a P.S... I feel my code is very Pythonic as its what I'm used to - maybe a lot of the code can be completed much more neatly(?) If so, can change on advice, or merge and others change later :)

crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// ## What it does
+/// Checks for use of a unary expression following an arithmetic operator.
+///
+/// ## Why is this bad?
+/// The use of a unary operator (`+`, `-` but *not* user-defined) following an arithmetic operator (
+/// `+`, `-`, `*`, `/`, `**` but *not* user-defined) can be ambiguous and is not part of the Fortran
+/// standard. The order of operations does not necessarily follow typical mathematical order. See
+/// (the Doctor Fortran
+/// article)[https://stevelionel.com/drfortran/2021/04/03/doctor-fortran-in-order-order/].
```
LiamPattinson: I think it would be best to move this to a `## References` section at the bottom of the doc-comment

crates/fortitude_linter/src/rules/portability/unary_following_arithmetic.rs
```diff
@@ -0,0 +1,155 @@
...
+///
+/// Some compilers may warn users of this code smell, but only via extensions. The use of a unary
```
LiamPattinson: I'm not sure it's the case that the warning comes from an extension, but rather that the warning is to let the user know that the compiler is applying a non-standard extension to the language.

LiamPattinson (commented):
Thanks @avadean, this looks good to me. I only have some thoughts on the docs. I've come around to the current name, so unless we can think of something better I'd lean towards keeping it as-is.

I don't see anything wrong with your code style. With Rust, there's usually a way to rewrite a series of `if` statements as a chain of `map`/`filter` on `Option`/`Result` types, but it's not necessarily better to do it that way, and I find your code really easy to understand at a glance.

As for the violation type, I think it should be `FixAvailability::Sometimes` in this case, as the rules where we've used `AlwaysFixableViolation` all seem to have simple stylistic fixes that can be run safely.

LiamPattinson (approved):
It's hard to tell exactly what changed since last time due to the rebase, but I think this all looks good. Thanks @avadean!

avadean:
> It's hard to tell exactly what changed since last time due to the rebase, but I think this all looks good. Thanks @avadean!

Yeah I was half realising that might be the case. In future perhaps I will just commit a bit more chronologically rather than worrying too much about a 'perfect' commit history! Sorry about that! The only change was rewording of the docs.

PR #687, opened 2026-06-30
Enable passing arbitrary config settings on the command line

ZedThree:
Deprecates `--config-file` in favour of `--config`

Feature and docs completely taken from Ruff and adapted


LiamPattinson (approved):
Thanks @ZedThree, this all looks great!

LiamPattinson:
We should probably make a new release after getting this and a couple of outstanding rules in.

ZedThree:
Yes, although it would be good to investigate performance beforehand too. If there's any particularly slow rules, we might just want to mark them as such

PR #688, opened 2026-06-30
Correctness upgrades

LiamPattinson:
Modernises and upgrades `correctness` rules in various ways:

- Use `kind_id` over `kind` wherever possible (and similar changes)
- Use annotations where appropriate
- Prefer `.eq_ignore_ascii_case` to manual lowercase conversions

Contains some drive-by fixes to make diagnostic messages more consistent.

I'd push the changes through the rest of the rule set, but this already took much longer than I expected!


crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__exit-or-cycle-in-unlabelled-loop_C142.f90.snap
```diff
@@ -1,33 +1,39 @@
...
 C142 'cycle' statement in unlabelled 'do' loop
-  --> ./resources/test/fixtures/correctness/C142.f90:27:7
+  --> ./resources/test/fixtures/correctness/C142.f90:26:5
    |
+24 |   label5: do i = 1, 2
 25 |     ! should warn
 26 |     do j = 1, 2
+   |     ----------- Unlabelled loop
 27 |       cycle
    |       ^^^^^
```
ZedThree: Maybe we should add some text to this annotation? I like this change, but I feel it draws a bit too much attention away from the problem statement?

Maybe something along these lines?
```
26 |     do j = 1, 2
   |     ----------- Unlabelled loop referenced by...
27 |       cycle
   |       ^^^^^ ...control flow statement here
```

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__nonportable-shortcircuit-inquiry_C161.f90.snap
```diff
@@ -1,37 +1,36 @@
...
 ---
 C161 variable inquiry `present(arg1)` and use in same logical expression
-  --> ./resources/test/fixtures/correctness/C161.f90:13:29
+  --> ./resources/test/fixtures/correctness/C161.f90:13:9
    |
 11 |     end if
 12 |
 13 |     if (present(arg1) .and. arg1 > 1) then
-   |                             ^^^^
+   |         ^^^^^^^^^^^^^       ^^^^
```
ZedThree: These are both `primary` annotations; is this a deliberate choice? Do we want some annotation text on these maybe, or is it obvious enough without?

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__unchecked-stat_C181.f90.snap
```diff
@@ -22,17 +26,29 @@ C181 stat argument 'status' is overwritten before being checked in this scope.
...
-59 |   allocate (x(10), stat=STATUS_X)
-   |                    ^^^^^^^^^^^^^
-60 |   if (.true.) then
-61 |     allocate (y(10), STAT=status_y)
-   |
+   --> ./resources/test/fixtures/correctness/C181.f90:59:20
+    |
+ 58 |   ! Repeat, but using a custom error handler
+ 59 |   allocate (x(10), stat=STATUS_X)
+    |                    ^^^^^^^^^^^^^
```
ZedThree: Maybe this should have some text, "Set here..."?

crates/fortitude_linter/src/rules/correctness/exit_labels.rs
```diff
@@ -219,21 +225,22 @@ impl AstRule for MissingEndLabel {
...
             .to_text(src)?
             .trim_end_matches(':');
 
-        let end = match node.kind() {
-            "select_case_statement" | "select_type_statement" | "select_rank_statement" => {
-                "end_select_statement".to_string()
-            }
-            "do_loop" => "end_do_loop_statement".to_string(),
+        let end = match node.kind_id() {
+            #[allow(clippy::manual_range_patterns)]
```
ZedThree, line 238: Is this because they have consecutive IDs or something?

ZedThree (dismissed):
Looks great, thanks @LiamPattinson! This must've taken a while to go through!

I'm happy for this to go in as-is, but I have a couple of bikeshedding comments about some of the new annotations.

Also, have you benchmarked this at all? I didn't really see much improvement with just using kind IDs on the entrypoints

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__nonportable-shortcircuit-inquiry_C161.f90.snap
LiamPattinson: Yes, I intended these to both be primary as it isn't clear to me which is the mistake and which is the extra information. Maybe the `present` call should be secondary and the usage primary? I agree that labelling them would be helpful.

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__exit-or-cycle-in-unlabelled-loop_C142.f90.snap
LiamPattinson: I'm not really sure what the best way is to add text to the standard primary annotation. I know you can get a mutable iterator over all annotations, so I could add it that way, but it feels clunky. We perhaps need some more wrapper functions to help generate these things.

crates/fortitude_linter/src/rules/correctness/exit_labels.rs
LiamPattinson: Yes, and clippy doesn't seem to realise that they're the result of macro expansions. It's odd, because there are many other scenarios where it's capable of working that out. I could set this `allow` globally to avoid these confusing statements everywhere, but then I worry we'll miss something elsewhere. There might be a way to set it within the macro itself, but I couldn't figure it out.

crates/fortitude_linter/src/rules/correctness/snapshots/fortitude_linter__rules__correctness__tests__exit-or-cycle-in-unlabelled-loop_C142.f90.snap
LiamPattinson: Never mind, I figured it out!

crates/fortitude_linter/src/rules/correctness/exit_labels.rs
ZedThree: Just `allow`ing it here seems reasonable. This is actually a little bit annoying, because the warning may come and go depending on how tree-sitter decides to number them. It might be possible to suppress this with a strong type, but maybe we just hope it doesn't crop up much for now

ZedThree (approved):
Really nice! I love the new annotations

LiamPattinson:
Bad news on the benchmarking front. If I build it locally rather than using the PyPI release, v0.9.0 takes about twice as long to finish on the usual GS2 test -- the same thing you experienced with your local builds. However, I can get a moderate speed up by using `--output-format=concise`:

```console
❯ hyperfine --warmup 3 -i './fortitude_0.9.0 check --select=ALL --preview --file-extensions="f90" src'
Benchmark 1: ../fortitude_0.9.0 check --select=ALL --preview --file-extensions="f90" src
  Time (mean ± σ):     487.6 ms ±  24.2 ms    [User: 2144.0 ms, System: 35.8 ms]
  Range (min … max):   450.8 ms … 527.5 ms    10 runs
 
❯ hyperfine --warmup 3 -i './fortitude_0.9.0 check --select=ALL --preview --file-extensions="f90" src --output-format=concise'
Benchmark 1: ./fortitude_0.9.0 check --select=ALL --preview --file-extensions="f90" src --output-format=concise
  Time (mean ± σ):     390.5 ms ±  24.5 ms    [User: 2047.9 ms, System: 36.8 ms]
  Range (min … max):   342.8 ms … 416.2 ms    10 runs
 
```

Running the new version, I find the concise output is basically the same as before despite us implementing a few new rules, but the regular output is a _lot_ slower now that we have the new diagnostics:

```console
❯ hyperfine --warmup 3 -i './fortitude_new check --select=ALL --preview --file-extensions="f90" src'
Benchmark 1: ./fortitude_new check --select=ALL --preview --file-extensions="f90" src
  Time (mean ± σ):      3.133 s ±  0.047 s    [User: 4.561 s, System: 0.180 s]
  Range (min … max):    3.085 s …  3.240 s    10 runs
 
❯ hyperfine --warmup 3 -i './fortitude_new check --select=ALL --preview --file-extensions="f90" src --output-format=concise'
Benchmark 1: ./fortitude_new check --select=ALL --preview --file-extensions="f90" src --output-format=concise
  Time (mean ± σ):     390.6 ms ±  18.5 ms    [User: 2096.2 ms, System: 40.9 ms]
  Range (min … max):   364.8 ms … 423.0 ms    10 runs
 
```

We were probably selling ourselves short in the JOSS paper benchmark by not using concise output. In future benchmarks, I should leave a note to keep it switched on.

ZedThree:
Dang, that is pretty disappointing.

Oh, I thought that you always used `concise` in the benchmarks! Yes, definitely wise to make a note to always use that. It's also worth noting that actually printing to the terminal costs a _lot_ more, so it might be interesting to just compare running without `hyperfine` to see the difference.

This is with a `release` build, yeah? Also, you should probably compare to `main`, rather than `0.9.0` to rule out any other rule changes (no pun intended).

Can you also compare for a single file?

Also, I managed to get a bit closer to the PyPI build speed by using `cibuildwheel` to build in the `manylinux` container -- there's definitely something about that environment that makes it much faster!

We should also investigate doing some proper profiling and trying to work out if there are any particularly slow rules. I've not looked into if/how ruff handles this

LiamPattinson:
Yes, I've been using release builds. Comparing to `main` gives essentially no difference compared to this PR:

```console
❯ hyperfine --warmup 3 -i '../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" src'
Benchmark 1: ../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" src
  Time (mean ± σ):      3.151 s ±  0.080 s    [User: 4.716 s, System: 0.144 s]
  Range (min … max):    3.056 s …  3.302 s    10 runs
 
❯ hyperfine --warmup 3 -i '../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" src --output-format=concise'
Benchmark 1: ../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" src --output-format=concise
  Time (mean ± σ):     370.9 ms ±   8.5 ms    [User: 1944.7 ms, System: 38.7 ms]
  Range (min … max):   359.4 ms … 390.2 ms    10 runs
 
```

The concise output is marginally faster on `main`, but I think that's just due to random fluctuations. I guess the parsing is the most expensive part, and the rules running over the tree is relatively cheap in comparison, so swapping a few string comparisons for int comparisons isn't going to make much difference overall.

I'd argue that switching to `kind!` and `kw!` is more important from a safety perspective than performance though. Being able to match on specific named vs unnamed nodes without accidentally picking up both will be very useful.

ZedThree:
Ok, then the massive slowdown is due to something else, and we can investigate that separately

PR #689, opened 2026-07-01
Fix links to relevant settings from rule docstrings

### LINKED ISSUE #682: `incorrect-keyword-case` docs missing link to relevant settings
[`incorrect-keyword-case`](https://fortitude.readthedocs.io/en/stable/rules/incorrect-keyword-case/) should link to [`check.settings.keyword-case`](https://fortitude.readthedocs.io/en/stable/settings/#check_incorrect-keyword-case_keyword-case)

ZedThree:
Closes #682

Try to be more consistent in how we link to rule-specific settings from the
docstrings, make it a compile-time error if the wrong section title is used, and
add something to the template for making new rules.

### DIFF
--- a/crates/fortitude_linter/src/rules/correctness/exit_labels.rs
+++ b/crates/fortitude_linter/src/rules/correctness/exit_labels.rs
@@ -108,8 +108,8 @@ impl AstRule for MissingExitOrCycleLabel {
 /// end do
 /// ```
 ///
-/// ## Settings
-/// See [check.exit-unlabelled-loops](../settings.md#checkexit-unlabelled-loops)
+/// ## Options
+/// - `check.exit-unlabelled-loops.allow-unnested-loops`
 #[derive(ViolationMetadata)]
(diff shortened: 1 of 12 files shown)


ZedThree:
Links to settings trigger the `broken_intra_doc_links` warning (which errors on
CI), but we handle these links via `cargo dev generate-docs`.

This is a feature taken from Ruff, and rather than explicitly silencing the
lint, they seem to just rely on the rule structs being `pub(crate)`, so I've done the same here.

LiamPattinson (dismissed):

ZedThree:
6dbf6910 is maybe a little silly and could be dropped perhaps

LiamPattinson:
> [6dbf691](https://github.com/PlasmaFAIR/fortitude/commit/6dbf6910fb621073de39688ddf86cf72a8058130) is maybe a little silly and could be dropped perhaps

I don't see how it's an issue, I think it'll help to keep things standardised.

LiamPattinson (approved):

PR #690, opened 2026-07-03
Fix performance regression in `line-too-long`

ZedThree:
Collecting the comment positions while walking the AST gives a ~30% speed up (with concise output) over using `named_descendant_for_byte_range` inside the check. The latter method has to walk the entire AST for each character, and caused the rule to take 5x as long as all of the AST rules combined.


LiamPattinson (approved):
Nice! We should really look into some kind of automatic benchmarking to catch this sort of thing before it becomes a problem.

PR #691, opened 2026-07-06
Fix diff render performance

ZedThree:
Improve the performance of rendering diffs with the `full` output-format in preview mode, by only diffing a small context window around the actual edit. Brings the performance to the same level as the 0.9.0 release.

Testing on GS2, as is tradition:

- `v0.9.0` tag (as built on my machine using the `profiling` profile in this PR):
```console
$ hyperfine -w 3 -i "/home/peter/Codes/fortitude/fortitude_0.9.0 check --select=ALL --preview"
Benchmark 1: /home/peter/Codes/fortitude/fortitude_0.9.0 check --select=ALL --preview
  Time (mean ± σ):      2.713 s ±  0.496 s    [User: 12.870 s, System: 0.126 s]
  Range (min … max):    2.100 s …  3.382 s    10 runs
```

- Current `main`:

```console
$ hyperfine -w 3 -i "/home/peter/Codes/fortitude/fortitude_main check --select=ALL --preview"
Benchmark 1: /home/peter/Codes/fortitude/fortitude_main check --select=ALL --preview
  Time (mean ± σ):     17.662 s ±  0.170 s    [User: 25.306 s, System: 0.725 s]
  Range (min … max):   17.476 s … 17.951 s    10 runs
```

- This PR:

```console
$ hyperfine -w 3 -i "/home/peter/Codes/fortitude/target/profiling/fortitude check --select=ALL --preview"
Benchmark 1: /home/peter/Codes/fortitude/target/profiling/fortitude check --select=ALL --preview
  Time (mean ± σ):      2.377 s ±  0.117 s    [User: 8.914 s, System: 0.106 s]
  Range (min … max):    2.082 s …  2.478 s    10 runs
```

Only one test had a change in the snapshot due to the change in context (also includes a drive-by fix to the test name, which was a copy-pasted leftover from an unrelated test)


LiamPattinson (dismissed):
Thanks @ZedThree, all looks good! Is this something that can feed back into Ruff?

I was also looking into how Ruff handles benchmarking in the CI and got a bit lost. Is performance regression something we could set up? Given we have a Python package, maybe `asv` would help?

ZedThree:
Way ahead of you :) I actually have some feedback from ruff to merge into this

I'll look into how they do benchmarking. `asv` is definitely a reasonable option

LiamPattinson (approved):

PR #702, opened 2026-07-24
Don't trigger `trailing-backslash` if the next node is a comment 

### LINKED ISSUE #694: `trailing-backslash` erroneously consideres it a violation if the next line is also a comment
The mentioned rule considers it a violation if a comment ends on a backslash because the next line may be masked by the preprocessor. This should not be an issue if a subsequent lines all of them ending on a backslash is followed by a comment without it. An example:
```
! This is invalid:
! My comment1 \
a = 1
! This should be valid according to the justification for the rule but is not
! Another comment \
! Subsequent comment
a = 2
! A more complex example
! My super long \
! and extensive \
! and informative comment
a = 3
```
What do you think?

ZedThree:
Fixes #694

Includes #701 to keep clippy happy

### DIFF
--- a/crates/fortitude_linter/resources/test/fixtures/correctness/C051.F90
+++ b/crates/fortitude_linter/resources/test/fixtures/correctness/C051.F90
@@ -8,4 +8,6 @@ program t
     A = 2.0
     print *, A
 
+    ! this is fine \
+    ! if this is a comment
 end program t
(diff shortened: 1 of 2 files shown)


LiamPattinson (dismissed):
Thanks @ZedThree, all looks good to me

LiamPattinson (approved):

PR #703, opened 2026-07-28
Don't raise `C072` for UDTIO routines

### LINKED ISSUE #697: C072 flags derived type IO
In fortitude 0.9.2 derived type IO interfaces are flagged due to usage of assumed size characters with wrong intent with rule C072. Would it make senses to exclude procedures for the `write(formatted)` / `write(unformatted)` interface since the user can't decide on the intent in this interface and whether the character variable is assumed size or not?

Examples:

- [stdlib `string_type` unformatted write](https://github.com/fortran-lang/stdlib/blob/9a15c7772f1a76a6c497b9f3abb793841fc81f74/src/strings/stdlib_string_type.fypp#L1115-L1128)
- [stdlib `string_type` formatted write](https://github.com/fortran-lang/stdlib/blob/9a15c7772f1a76a6c497b9f3abb793841fc81f74/src/strings/stdlib_string_type.fypp#L1130-L1154)
- [stdlib `string_type` formatted read](https://github.com/fortran-lang/stdlib/blob/9a15c7772f1a76a6c497b9f3abb793841fc81f74/src/strings/stdlib_string_type.fypp#L1175-L1223)
- [stdlib `string_type` unformatted read](https://github.com/fortran-lang/stdlib/blob/9a15c7772f1a76a6c497b9f3abb793841fc81f74/src/strings/stdlib_string_type.fypp#L1157-L1173)

````f90
interface write(formatted)
   module procedure :: write_formatted
end interface

contains

subroutine write_formatted(self, unit, iotype, v_list, iostat, iomsg)
   class(demo_type), intent(in) :: self
   integer, intent(in) :: unit
   character(len=*), intent(in) :: iotype
   integer, intent(in) :: v_list(:)
   integer, intent(out) :: iostat
   character(len=*), intent(inout) :: iomsg

   ! ...
end subroutine write_formatted
````

### PR DESCRIPTION
**Pull request by ZedThree**
Fixes #697 

Also parse tree-sitter `intrinsic_type` nodes into Rust enums. I started trying to do `kind` nodes as well, but I was getting really bogged down as they're surprisingly complicated, and it wasn't immediately obvious how to exploit them anyway.

Also some drive-by refactoring because there were some todos.

### DIFF
--- a/crates/fortitude_linter/src/rules/correctness/assumed_size.rs
+++ b/crates/fortitude_linter/src/rules/correctness/assumed_size.rs
@@ -1,4 +1,5 @@
 use crate::ast::FortitudeNode;
+use crate::ast::symbol_table::SymbolTable;
 use crate::ast::types::{AttributeKind, HasName, Intent, Type};
 use crate::diagnostics::{Annotation, Diagnostic, Span, Violation};
 use crate::settings::FortranStandard;
@@ -6,6 +7,7 @@ use crate::traits::HasNode;
 use crate::{AstRule, CheckContext, kind_ids};
 use fortitude_macros::{ViolationMetadata, kind};
 use itertools::Itertools;
+use log::debug;
 use ruff_macros::derive_message_formats;
 use tree_sitter::Node;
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: crates/fortitude_linter/src/rules/correctness/assumed_size.rs [hunk 1]
```diff
@@ -55,29 +59,24 @@ impl Violation for AssumedSize {
...
+            return None;
         }
 
         // Assumed size ok for parameters
-        if declaration
-            .children_by_field_id(field!("attribute"), &mut declaration.walk())
-            .filter_map(|attr| attr.to_text(src))
-            .any(|attr_name| attr_name.eq_ignore_ascii_case("parameter"))
-        {
+        if declaration.has_attribute(AttributeKind::Parameter) {
```
LiamPattinson, line 79: Not necessary in this PR, but at some point it'd be great to have a bunch of functions like `is_parameter`, `is_allocatable`, etc.

LiamPattinson (approved):
Thanks @ZedThree, all looks good, and I'm glad to see more additions to `ast`!

PR #704, opened 2026-07-29
Add rules for missing default in both select type and select rank

imciner2:
This is a follow-on to https://github.com/PlasmaFAIR/fortitude/pull/240 and https://github.com/PlasmaFAIR/fortitude/issues/213 to add two more rules to flag when the `class default` and `rank default` cases of the `select type` and `select rank` statements, respectively, are missing.


LiamPattinson (approved):
Thanks @imciner2, honestly no notes, this all looks great to me!

PR #705, opened 2026-08-03
Add wrapper around `tree-sitter`; stop passing source around so much

ZedThree:
Inspired by [`yak-sitter`](https://github.com/Jakobeha/type-sitter/blob/main/yak-sitter/), this is a mostly drop-in replacement for `tree-sitter`, except that like `yak-sitter`, it stores the source text in the `Tree` which allows each `Node` to immediately access its text. The cost of this is that it's now unsafe to make edits to the `Tree` -- but we modify the source text and re-parse, so that's not a problem for us.

This means that instead of stuff like `node.to_text(src).context("expected text")?`, we can just write `node.text()`. This often means we then don't need to pass `src` down a whole stack of functions.

Unfortunately, this touches almost everything. Thankfully, [`rust-analyzer` can remove unused parameters trivially](https://rust-analyzer.github.io/book/assists.html#remove_unused_param), so this was actually pretty easy to implement.


Cargo.toml
```diff
@@ -97,13 +98,13 @@ tracing-subscriber = { version = "0.3.18", default-features = false, features =
...
-tree-sitter = "~0.25.0"
-tree-sitter-cli = "~0.25.0"
+tree-sitter = "~0.26.0"
+tree-sitter-cli = "~0.26.0"
 tree-sitter-fortran = { git = "https://github.com/stadelmanma/tree-sitter-fortran.git", rev = "32a3e63" }
 toml = "1.0.0"
 unicode-width = "0.2.0"
 url = { version = "2.5.0" }
-
+yak-sitter = { path = "/home/peter/Tools/tree-sitter/type-sitter/yak-sitter/" }
```
LiamPattinson: Should this point to a git repo?

ZedThree: Oops, that was just there for testing, we don't use it at all now (or rather, it's been entirely vendored)

crates/fortitude_sitter/src/ast/types.rs
```diff
@@ -1068,13 +1062,161 @@ impl<'a> HasNode<'a> for UsedItem<'a> {
...
 }
 
+/// Returns true if the type passed to it is number-like, and of a kind that can be modified using
+/// kinds. 'double precision' and 'double complex' are not included.
+pub fn dtype_is_plain_number(dtype: &str) -> bool {
+    matches!(
+        dtype.to_lowercase().as_str(),
+        "integer" | "real" | "logical" | "complex"
+    )
+}
```
LiamPattinson, line 1072: Not worth fixing in this PR, but I'm sure this can be replaced with `Symbol` checking nowadays

LiamPattinson (commented):
All looks good! I think it may need to be rebased before merging for the new rule that was just added

ZedThree:
Rebased to fix the new rule

LiamPattinson (approved):
Thanks @ZedThree!

PR #707, opened 2026-08-04
Add rule: NonPortableSystemCall

### LINKED ISSUE #282: Suggest that `system` be replaced by `execute_command_line`
Fortitude could suggest that the `system` subroutine, a gfortran extension, be replaced by `execute_command_line` in code such as

```fortran
implicit none
call system("dir")  ! non-standard
call execute_command_line("dir")
end
```

drewlongdin-dev:
Resolved #282

### DIFF
--- a/crates/fortitude_linter/src/rules/portability/non_portable_system_call.rs
+++ b/crates/fortitude_linter/src/rules/portability/non_portable_system_call.rs
@@ -1,0 +1,60 @@
+use crate::ast::FortitudeNode;
+use crate::diagnostics::{Diagnostic, Fix, Violation};
+use crate::{AstRule, CheckContext, kind_ids};
+use fortitude_macros::{ViolationMetadata, field};
+use ruff_macros::derive_message_formats;
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for use of the non-portable `system` call.
+///
+/// ## Why is this bad?
+/// `system` is a GFortran extension and isn't available as part of other compilers.
+///
+/// ## Example
+/// ```f90
+/// call system("dir")
+/// ```
+///
+/// Use instead:
+/// ```f90
+/// call execute_command_line("dir")
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct NonPortableSystemCall;
+
+impl Violation for NonPortableSystemCall {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Use of non-portable system call".to_string()
+    }
+}
+
+impl AstRule for NonPortableSystemCall {
+    fn check(context: &CheckContext, node: &Node) -> Option<Vec<Diagnostic>> {
+        let identifier = node.child_by_field_id(field!("subroutine").into()).unwrap();
+
+        if !(identifier
+            .to_text(context.source_text())?
+            .eq_ignore_ascii_case("system"))
+        {
+            return None;
+        }
+
+        if context.symbol_table().get("system").is_some() {
+            return None;
+        }
+
+        let edit = identifier.edit_replacement(&context.file, "execute_command_line".to_string());
+
+        some_vec!(
+            context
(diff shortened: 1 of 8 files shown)


crates/fortitude_linter/src/rules/portability/non_portable_system_call.rs
```diff
@@ -0,0 +1,60 @@
...
+/// call execute_command_line("dir")
+/// ```
+#[derive(ViolationMetadata)]
+pub(crate) struct NonPortableSystemCall;
+
+impl Violation for NonPortableSystemCall {
+    #[derive_message_formats]
+    fn message(&self) -> String {
+        "Use of non-portable system call".to_string()
+    }
```
LiamPattinson, line 34: If we add a `fix_title`, it'll give a little more help text to users in the terminal. This might require updating the snapshots.

```suggestion
    }
    
    fn fix_title(&self) -> Option<String> {
        Some("Replace with `execute_command_line`".to_string())
    }
```

crates/fortitude_linter/src/rules/portability/non_portable_system_call.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl AstRule for NonPortableSystemCall {
+    fn check(context: &CheckContext, node: &Node) -> Option<Vec<Diagnostic>> {
+        let identifier = node.child_by_field_id(field!("subroutine").into()).unwrap();
+
+        if !(identifier
```
LiamPattinson, line 42: Adds some explanatory comments
```suggestion
        // Skip subroutines with other names
        if !(identifier
```

crates/fortitude_linter/src/rules/portability/non_portable_system_call.rs
```diff
@@ -0,0 +1,60 @@
...
+            .to_text(context.source_text())?
+            .eq_ignore_ascii_case("system"))
+        {
+            return None;
+        }
+
+        if context.symbol_table().get("system").is_some() {
```
LiamPattinson, line 50: Ditto
```suggestion
        // Exit early if there is a user-defined symbol called 'system'
        if context.symbol_table().get("system").is_some() {
```

crates/fortitude_linter/src/rules/portability/non_portable_system_call.rs
```diff
@@ -0,0 +1,60 @@
+use crate::ast::FortitudeNode;
+use crate::diagnostics::{Diagnostic, Fix, Violation};
+use crate::{AstRule, CheckContext, kind_ids};
+use fortitude_macros::{ViolationMetadata, field};
+use ruff_macros::derive_message_formats;
+use tree_sitter::Node;
+
+/// ## What it does
+/// Checks for use of the non-portable `system` call.
```
LiamPattinson: Adds a little more context to the docs:

```suggestion
/// Checks for use of the non-portable `system` call for running programs
/// on the command line.
```

You'll need to re-run `cargo dev generate-all` after applying this.

LiamPattinson (commented):
Thanks @drewlongdin-dev, and congrats on your first contribution! :tada:

I have a few suggestions, but it's just a few edits to the docs/comments and the addition of an optional function. You'll probably need to rebuild the docs and snapshots after some of these changes, so it's probably better to implement them locally than to accept the changes via GitHub.

LiamPattinson:
Thanks @drewlongdin-dev, all looks good! If you like, you can add your details to the file `CITATION.cff` at the top level of the repo. Don't worry if you don't have an ORCID; it's a unique reference ID for academics, and we can always add this later if you make one.

drewlongdin-dev:
@LiamPattinson I don't have a ORCID yet so I left that field empty. I will update in the event I get one. Thanks for the opportunity, I had great fun!

LiamPattinson (approved):

PR #708, opened 2026-08-04
Parse bind attributes on variables

### LINKED ISSUE #706: Unexpected `E000` when using `bind`
I'm working on developing Fortran bindings for the [Enzyme automatic differentiation framework](https://github.com/EnzymeAD/Enzyme) and was setting up Fortitude for linting. The report is generally what I would expect, except for an `E000 Error opening file` error. Here's a reproducer:
```fortran
module test
  use iso_c_binding, only: c_int
  implicit none
  private
  integer(c_int), public, bind(C, name="enzyme_const") :: enzyme_const
end module
```
gives
```console
E000 Error opening file: unknown attribute 'bind'
  Cause: unknown attribute 'bind'
  Cause: Matching variant not found
--> tmp.f90:1:1

fortitude: 0 files scanned, 1 could not be read.
Number of errors: 1
```
using latest Fortitude at eb5e7fcfc1cb6f76d8fd97ade654a97df3f96b56.

I was under the impression what I'd done was valid Fortran. Has `bind` not been included as a possible attribute? We use it for C-binding functions and subroutines in [FTorch](https://github.com/Cambridge-ICCS/FTorch) and Fortitude doesn't complain about its usage there.

imciner2:
After some bumbling with the AST, I managed to get the `bind()` part parsed into a struct (the standard says the name is optional, but the language is not, but it is in effect always `C`).

It would be great if there was an equivalent to the `print-ast` command for the Fortitude node tree to verify this though.

Fixes https://github.com/PlasmaFAIR/fortitude/issues/706

### DIFF
--- a/crates/fortitude_sitter/src/ast/types.rs
+++ b/crates/fortitude_sitter/src/ast/types.rs
@@ -214,29 +214,48 @@ impl<'a> Dimension<'a> {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, EnumIs, IntoStaticStr, PartialEq)]
+pub enum BindLanguage {
+    #[default]
+    C,
+    Other,
+}
+
+impl BindLanguage {
+    pub fn from_node(node: &Node) -> Self {
+        if node.text().to_uppercase().eq("C") {
+            Self::C
+        } else {
+            Self::Other
+        }
+    }
+}
+
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct Bind<'a> {
-    language: &'a str,
+    language: BindLanguage,
     name: Option<&'a str>,
 }
 
 impl<'a> Bind<'a> {
-    pub fn try_from_node(node: &Node<'a>, src: &'a str) -> Result<Self> {
+    pub fn try_from_node(node: &Node<'a>) -> Result<Self> {
         // Identifier is a required node in a bind() attribute
         let lang_node = node
             .child_with_id(kind!("identifier"))
             .expect("must have identifier child");
-        let lang = lang_node.to_text(src).expect("must have text");
+        let lang = BindLanguage::from_node(&lang_node);
 
         // The name node is an optional one
-        if let Some(name_kw_node) = node.child_with_id(kind!("keyword_argument")) {
-            if let Some(name_node) = name_kw_node.child_with_id(kw!("value")) {
-                let name = name_node.to_text(src).expect("must have text for name");
-                return Ok(Self {
-                    language: lang,
-                    name: Some(name),
-                });
-            }
+        if let Some(name_kw_node) = node.child_with_id(kind!("keyword_argument"))
+            && let Some(name_node) = name_kw_node.child_with_id(kind!("string_literal"))
+        {
+            // Remove the quotes from the string literal
+            let name = name_node.text().trim_matches('\"');
+
+            return Ok(Self {
+                language: lang,
+                name: Some(name),
+            });
         }
 
         Ok(Self {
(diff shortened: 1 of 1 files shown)


LiamPattinson (commented):
Thanks @imciner2, this all looks good to me! Please could you add the code snippet in #706 to a test at the bottom of `types.rs` and check that the `Bind` object is parsed correctly? If you're busy, I could push to this branch if you're happy for me to do so.

ZedThree:
Sorry, before this goes in, could we rebase this onto #705?

imciner2:
> Sorry, before this goes in, could we rebase this onto https://github.com/PlasmaFAIR/fortitude/pull/705?

Do you want me to retarget it to that branch then, or is that PR going to be merged soon?

LiamPattinson:
#705 is going to be merged soon, I'm just taking my time reviewing it (I'm probably about 70% done). If you rebase on top of there, we'll get that merged, and then this can go in immediately afterwards.

LiamPattinson:
@imciner2 You should now be able to rebase on `main`. The primary change will be to replace:

```rust
node.to_text(src).expect("must have text")
```

with simply:

```rust
node.to_text()
```

ZedThree:
Correction: just `node.text()` :)

imciner2:
I have just rebased this to be on top of the new node text handling, changed how the language part is handled (it is technically supposed to be only `C`, but I put an unknown option just in case), and also added the two testcases that handle both named and unnamed `bind()` attributes.

crates/fortitude_sitter/src/ast/types.rs
```diff
@@ -214,6 +214,57 @@ impl<'a> Dimension<'a> {
...
+#[derive(Clone, Copy, Debug, Default, EnumIs, IntoStaticStr, PartialEq)]
+pub enum BindLanguage {
+    #[default]
+    C,
+    Other,
+}
+
+impl BindLanguage {
+    pub fn from_node(node: &Node) -> Self {
+        if node.text().to_uppercase().eq("C") {
```
ZedThree: Not super important, but for consistency elsewhere:

```suggestion
        if node.text().eq_ignore_ascii_case("C") {
```

crates/fortitude_sitter/src/ast/types.rs
```diff
@@ -214,6 +214,57 @@ impl<'a> Dimension<'a> {
...
+            .child_with_id(kind!("identifier"))
+            .expect("must have identifier child");
+        let lang = BindLanguage::from_node(&lang_node);
+
+        // The name node is an optional one
+        if let Some(name_kw_node) = node.child_with_id(kind!("keyword_argument"))
+            && let Some(name_node) = name_kw_node.child_with_id(kind!("string_literal"))
+        {
+            // Remove the quotes from the string literal
+            let name = name_node.text().trim_matches('\"');
```
ZedThree: We need to also handle single quotes, which I think should just be:
```suggestion
            let name = name_node.text().trim_matches(&['\"', '\'']);
```

crates/fortitude_sitter/src/ast/types.rs
```diff
@@ -214,6 +214,57 @@ impl<'a> Dimension<'a> {
...
+    language: BindLanguage,
+    name: Option<&'a str>,
+}
+
+impl<'a> Bind<'a> {
+    pub fn try_from_node(node: &Node<'a>) -> Result<Self> {
+        // Identifier is a required node in a bind() attribute
+        let lang_node = node
+            .child_with_id(kind!("identifier"))
+            .expect("must have identifier child");
```
ZedThree: My current thinking is that it makes most sense to bubble up errors and catch them when we make the symbol table, so please could you switch to this:

```suggestion
            .context("must have identifier child")?;
```

ZedThree (changes requested):
LGTM, thanks @imciner2! A couple of minor points to fix

crates/fortitude_sitter/src/ast/types.rs
imciner2: Thanks! I thought there must have been something that would do this, but just couldn't find it.

ZedThree (approved):
Lovely, thanks @imciner2!

joewallwork:
Thanks for the quick fix!

PR #710, opened 2026-08-05
Fix case-sensitive `SymbolTable` lookups

### LINKED ISSUE #709: `SymbolTable` lookups are case-sensitive
Discovered while working with @drewlongdin-dev yesterday: `SymbolTable` lookups are sometimes case sensitive. The `Symbol` itself should retain its original capitalisation, but it should be possible to look up variables with any case. The following test fails:

```rust
    #[test]
    fn mixed_case() -> Result<()> {
        let mut parser = Parser::new();
        parser
            .set_language(&tree_sitter_fortran::LANGUAGE.into())
            .context("Error loading Fortran grammar")?;

        let code = r#"
module mod
  use :: some_module, only: foo, BaR => baz
  implicit none (type, external)
  Integer :: Idx, count
contains
  subroutine sub(x)
    integer, intent(in) :: x
  end subroutine sub
  function FuNc(x) result(y)
    integer, intent(in) :: x
  end function FuNc
end module mod
}
"#;
        let tree = parser.parse(code, None).context("Failed to parse")?;
        let root = tree.root_node().child(0).context("Missing child")?;

        let symbol_table = SymbolTable::new(&root, code)?;

        let foo = symbol_table.get("Foo");
        assert!(foo.is_some());
        let foo = foo.unwrap();
        assert!(foo.name().as_str() == "foo");

        let bar = symbol_table.get("bar");
        assert!(bar.is_some());
        let bar = bar.unwrap();
        assert!(bar.name().as_str() == "BaR");

        let idx = symbol_table.get("idx");
        assert!(idx.is_some());
        let idx = idx.unwrap();
        assert!(idx.name().as_str() == "Idx");

        let count = symbol_table.get("COUNT");
        assert!(count.is_some());
        let count = count.unwrap();
        assert!(count.name().as_str() == "count");

        let sub = symbol_table.get("SUB");
        assert!(sub.is_some());
        let sub = sub.unwrap();
        assert!(sub.name().as_str() == "sub");

        let func = symbol_table.get("func");
        assert!(func.is_some());
        let func = func.unwrap();
        assert!(func.name().as_str() == "FuNc");
        Ok(())
    }
```

I've already written the fix, just raising this for record keeping!

LiamPattinson:
Fixes #709

### DIFF
--- a/crates/fortitude_sitter/src/ast/symbol_table.rs
+++ b/crates/fortitude_sitter/src/ast/symbol_table.rs
@@ -171,12 +171,12 @@ impl<'a> SymbolTable<'a> {
     pub fn insert_from_decl_line(&mut self, decl: VariableDeclaration<'a>, dummy_vars: &[String]) {
         let decl = Rc::new(decl);
         for name in decl.names().iter() {
-            let name_lower = name.name().as_str().to_ascii_lowercase();
-            let is_dummy_var = dummy_vars.contains(&name_lower);
-            self.inner.insert(
-                name_lower,
-                Symbol::Variable(Variable::new(name.clone(), is_dummy_var, decl.clone())),
-            );
+            let is_dummy_var = dummy_vars.contains(&name.name().as_str().to_ascii_lowercase());
+            self.insert_symbol(Symbol::Variable(Variable::new(
+                name.clone(),
+                is_dummy_var,
+                decl.clone(),
+            )));
         }
         self.decl_lines.push(decl);
     }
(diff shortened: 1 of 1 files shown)


ZedThree:
Please could you rebase on top of #705? That PRs moves this file

LiamPattinson:
I've just approved #705, so once that's merged I'll rebase on `main`

LiamPattinson:
@ZedThree This should now be ready to review

crates/fortitude_sitter/src/ast/symbol_table.rs
```diff
@@ -171,12 +171,12 @@ impl<'a> SymbolTable<'a> {
     pub fn insert_from_decl_line(&mut self, decl: VariableDeclaration<'a>, dummy_vars: &[String]) {
         let decl = Rc::new(decl);
         for name in decl.names().iter() {
-            let name_lower = name.name().as_str().to_ascii_lowercase();
-            let is_dummy_var = dummy_vars.contains(&name_lower);
-            self.inner.insert(
-                name_lower,
-                Symbol::Variable(Variable::new(name.clone(), is_dummy_var, decl.clone())),
-            );
+            let is_dummy_var = dummy_vars.contains(&name.name().as_str().to_ascii_lowercase());
```
ZedThree: Can we do something like:

```suggestion
            let is_dummy_var = dummy_vars.iter().find(|val| val.eq_ignore_ascii_case(name.name().as_str()));
```

to save lowercasing the name twice?

ZedThree (dismissed):
LGTM!

It looks like there's a couple of crates for case-insensitive hashmaps that might also be useful?

crates/fortitude_sitter/src/ast/symbol_table.rs
```diff
@@ -171,12 +171,15 @@ impl<'a> SymbolTable<'a> {
...
-            let name_lower = name.name().as_str().to_ascii_lowercase();
-            let is_dummy_var = dummy_vars.contains(&name_lower);
-            self.inner.insert(
-                name_lower,
-                Symbol::Variable(Variable::new(name.clone(), is_dummy_var, decl.clone())),
-            );
+            let is_dummy_var = dummy_vars
+                .iter()
+                .find(|val| val.eq_ignore_ascii_case(name.name().as_str()))
+                .is_some();
```
ZedThree: Sorry, should've spotted this before, is this just:

```suggestion
                .any(|val| val.eq_ignore_ascii_case(name.name().as_str()));
```

ZedThree (approved):

PR #711, opened 2026-08-05
Run clippy on PR workflows

imciner2:
@ZedThree this change will make Clippy run on the PR checks and not just in the fork's push I believe. So then the results will show in the PR flow.


ZedThree (approved):
Thanks @imciner2!

PR #715, opened 2026-08-07
Add `name` and `count` output formats

### LINKED ISSUE #698: Option for reporting name only
As far as I can see, `fortitude check` doesn't have an option for only reporting the names of the files that are problematic under the given configuration. I use `git diff --name-only` and `git log --name-only` a lot and feel that something similar would be a useful addition.

ZedThree:
Similar to `grep --files-with-matches` and `grep --count`, respectively.

Closes #698

I've styled this as bold red, to match other elements:

image

but maybe it should be purple, to match `grep`/`rg` etc?

image


LiamPattinson (approved):
Nice, all looks good!

joewallwork:
This is just what I was after, thanks!

PR #722, opened 2026-08-10
Add some missing nodes to S105

### LINKED ISSUE #716: Where blocks do not get indented with S105
The `where` block constructs that create a new multi-line scope don't get flagged by the S105 style rule for improper indentation.

Take this example:
```fortran
    dimension pressure(1000), temp(1000), precepitation(1000)

    where(pressure .GE. 1.0)
    pressure = pressure + 1.0
    temp = temp - 10.0
    elsewhere
    precepitation = .TRUE.
    endwhere
```
it will not give any warnings for S105, saying it is correctly indented. However the `where` clauses should introduce a new indentation level, and it should be:
```fortran
    dimension pressure(1000), temp(1000), precepitation(1000)

    where(pressure .GE. 1.0)
        pressure = pressure + 1.0
        temp = temp - 10.0
    elsewhere
        precepitation = .TRUE.
    endwhere
```

ZedThree:
Adds `where`, `select type`, `select rank` to S105

Fixes #716 

Also adds test case names, this doesn't affect anything except the test name that gets printed when run

### DIFF
--- a/crates/fortitude_linter/src/rules/style/whitespace.rs
+++ b/crates/fortitude_linter/src/rules/style/whitespace.rs
@@ -847,12 +847,14 @@ mod tests {
     use std::fs;
     use std::process::Command;
     use tempfile::TempDir;
+    use test_case::test_case;
 
-    #[test_case::test_case(
+    #[test_case(
         2,
         "!> My program\nprogram test\n        implicit none\nend program test"
+        ; "indent 2"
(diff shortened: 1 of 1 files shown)


LiamPattinson (approved):

imciner2:
Thanks @ZedThree! That was fast.

PR #724, opened 2026-08-10
Remove several markdown files for old rules

imciner2:
This was generated by deleting all files in the rules directory, and then doing a fresh `cargo dev generate-all` run.


ZedThree (approved):
Thanks @imciner2! We should probably add a test or CI check for extraneous files in generated dirs

imciner2:
> We should probably add a test or CI check for extraneous files in generated dirs

That's one way, but actually what I initially tried was just deleting the entire `docs` folder and generating it. When that failed (it complained about missing directories), I just deleted everything in the `rules` folder. So another thing might be to just have a clean run before it does the generate where it only cleans out the generated files.

PR #727, opened 2026-08-11
Display rule name

### LINKED ISSUE #713: Switch to reporting rule names in output?
In ruff 0.16, preview mode now reports the human-readable rule names instead of rule codes, with [a setting](https://docs.astral.sh/ruff/settings/#output-prefer-rule-codes) to switch back to codes.

This is something we've talked about before, and if ruff are doing it, that's a good indication that it's probably sensible.

(They're also finally letting you select rules by name with `--preview` too now, something we've had from the beginning!)

ZedThree:
Builds on #726 

Closes #713 

Preview mode output now looks like this by default:

```
error[implicit-external-procedures][*]: 'implicit none' missing 'external'
 --> crates/fortitude_linter/resources/test/package/src/package/app.f90:2:3
  |
1 | program app
2 |   implicit none
  |   ^^^^^^^^^^^^^
3 | end program app
  |
help: Add `(external)` to 'implicit none'
1 | program app
  -   implicit none
2 +   implicit none (type, external)
3 | end program app
note: This is an unsafe fix and may change runtime behavior
```

but now you can use `check.output-rule-id-format = "both"` or `"code"`. `"both"`:

```
error[implicit-external-procedures (C003)][*]: 'implicit none' missing 'external'
...
```
`"code"`:
```
error[C003][*]: 'implicit none' missing 'external'
...
```

For reference, current `main` without preview:
```
C003 [*] 'implicit none' missing 'external'
 --> crates/fortitude_linter/resources/test/package/src/package/app.f90:2:3
  |
1 | program app
2 |   implicit none
  |   ^^^^^^^^^^^^^
3 | end program app
  |
help: Add `(external)` to 'implicit none'
```


LiamPattinson:
Thanks @ZedThree! I'm not sure about the inclusion of `error` at the start, but if we're moving to a model where rule severities can be modified, I guess that's information we'll need to output going forward. The use of so many square brackets also looks like it could be cleaned up. How about just two layers of colons?

- `error: C003 [*]: 'implicit none' missing 'external'`
- `error: implicit-external-procedures [*]: 'implicit none' missing 'external'`
- `error: C003 implicit-external-procedures [*]: 'implicit none' missing 'external'`

ZedThree:
That formatting is handled by the `annotate-snippets`, so we have less control over it without reimplementing some things ourselves.

LiamPattinson (dismissed):

LiamPattinson (dismissed):

LiamPattinson:
Fair enough, I'll defer to Ruff for now!

LiamPattinson:
I've tried to approve twice, but it says the merge-base keeps changing. Could you let me know when you've finished what you're doing?

ZedThree:
Er, not sure what's going on there, I haven't touched anything! Let me just rebase on `main` and see if that makes it happy

ZedThree:
Try now?

LiamPattinson (approved):

LiamPattinson:
Sorry, no idea why this didn't show in my notifications!

PR #729, opened 2026-08-12
Hide `fortitude explain...` note in summary text in preview mode

### LINKED ISSUE #291: Should be able to toggle the summary
How about a `--summary` flag? I was thinking about the following options:

- `--summary=full`: Files scanned, files failed to scan (if any), time taken, errors found (or 'all checks passed')
- `--summary=concise`: Only errors found or 'all checks passed' (same as Ruff)
- `--summary=off`: Only print diagnostics, no summary.

I'm unsure if this should be its own thing or just something controlled by the log levels.

---

As an aside, is it worth keeping this bit?

```bash
For more information about specific rules, run:

    fortitude explain X001,Y002,...

```

I added this as I thought it would be good for the tool to direct users towards other helpful features, as I imagine most users won't read the docs and I think this one is quite important (if you've been using `double precision` without issue for the last 40 years only for this tool to start telling you off, I can understand why an explanation might be warranted!). The main downside is that it adds quite a lot of visual noise at the end of the output, and after you've seen it a hundred times it stops being useful.

ZedThree:
Closes #291

~~Builds on #727 to avoid snapshot conflicts~~

Just disables the note in the summary text when running with `--preview`. We've talked about possibly removing it before, this seems like a reasonable time to do it


LiamPattinson:
I think this closes https://github.com/PlasmaFAIR/fortitude/issues/291

LiamPattinson:
Sorry for not reviewing this sooner! It seems there are fresh conflicts since it was raised. I can take a look if it they aren't too difficult to resolve.

ZedThree:
No worries, I'll fix them up

LiamPattinson (approved):

Issue #718, opened 2026-08-07
Semicolons in comments block S105 indentation


imciner2:
If there is a semicolon in an inline comment, then that entire line is not indented properly, e.g.

```fortran
if( present(iStatus) ) then
iStatus = 0 ! = 0 - full stuck, 1 - slip, 2 - separation; 3 - full separation
endif
```
prevents the inner part of the if statement from getting indented.

While I realize that the initial implementation of S105 didn't handle lines with semicolons on them, I think that should be changed to still indent lines that include a semicolon in the comment.

Issue #717, opened 2026-08-07
do...continue loops mis-indented in S105


imciner2:
This is mostly a problem for putting old code through S105 to give it a consistent format. In one file, I have a copy of some BLAS functions, and those still tend to include `do...continue` loops. The `continue` statement is not detected as the end of the loop, which means the indentation level is not decreased after the loop.

For example, take this part of the BLAS:
```fortran
      IF (BETA.NE.ONE) THEN
          IF (INCY.EQ.1) THEN
              IF (BETA.EQ.ZERO) THEN
                  DO 10 I = 1,N
                      Y(I) = ZERO
   10             CONTINUE
              ELSE
                  DO 20 I = 1,N
                      Y(I) = BETA*Y(I)
   20             CONTINUE
              END IF
          ELSE
              IY = KY
              IF (BETA.EQ.ZERO) THEN
                  DO 30 I = 1,N
                      Y(IY) = ZERO
                      IY = IY + INCY
   30             CONTINUE
              ELSE
                  DO 40 I = 1,N
                      Y(IY) = BETA*Y(IY)
                      IY = IY + INCY
   40             CONTINUE
              END IF
          END IF
      END IF
```

putting it into S105, its indentation gets broken.

```fortran
    IF (BETA.NE.ONE) THEN
        IF (INCY.EQ.1) THEN
            IF (BETA.EQ.ZERO) THEN
                DO 10 I = 1,N
                    Y(I) = ZERO
10                  CONTINUE
                ELSE
                    DO 20 I = 1,N
                        Y(I) = BETA*Y(I)
20                      CONTINUE
                    END IF
                ELSE
                    IY = KY
                    IF (BETA.EQ.ZERO) THEN
                        DO 30 I = 1,N
                            Y(IY) = ZERO
                            IY = IY + INCY
30                          CONTINUE
                        ELSE
                            DO 40 I = 1,N
                                Y(IY) = BETA*Y(IY)
                                IY = IY + INCY
40                              CONTINUE
                            END IF
                        END IF
                    END IF
```

ZedThree:
Hmm, this might be tricky in general with the current algorithm, because non-block `do` loops can have shared termination:

```f90
do 10 i = 1, n
  do 10 j = 1, m
     ...
10 continue
```

which looks like the following AST:

```
(do_loop
  ...
  (do_loop
    ...
    (end_do_label_loop_statement
      do_label: (do_label_virtual)))
  (end_do_label_loop_statement
[...]

PR #734, opened 2026-08-14
Add `LogicalLines`

LiamPattinson:
Raising this as a draft so I can get some feedback before continuing.

Adds a `LogicalLines` view into each file. This differs from the 'physical' lines in the following ways:

- Empty lines are skipped
- A physical line can contain multiple logical lines separated by semicolons
- A logical line can stretch over multiple continued physical lines (`&`)

`LogicalLines` is built up while Fortitude passes over the AST as part of the standard checks. I've simulated this in the tests, but haven't plugged it in yet.

I've also added machinery to track the expected indent as this structure is built. I'm hoping to use this to simplify the implementation of `incorrect-indent` and catch some bugs:

- This can handle shared continues in nested do loops! https://github.com/PlasmaFAIR/fortitude/issues/717
- It can handle semicolons in comments too https://github.com/PlasmaFAIR/fortitude/issues/718

TODO:

- [x] Hook into the standard checking logic.
- [x] Add `impl` methods for `LogicalLines` and `LogicalLine` for ease of use.
- [ ] Update implementation of `incorrect-indent`, check it still passes current tests.
- [ ] Add further tests for #717 and #718 bugs.

Once I've gotten that done, it would be handy if this could be stress-tested against some cursed code and a bunch of old constructs.

It should also be useful for simplifying the implementation of `superfluous-semicolon` and `multiple-statements-per-line`, but I'll save that for a future PR.


LiamPattinson:
I've refactored `S105` and `S106` to use `LogicalLines`. This outsources a lot of the complexity to the construction of `LogicalLines`, and keeps the rules themselves simple. I'm currently doing a performance comparison too to see if it runs any faster.

There are some major changes:

- `S105` no longer impacts semicolons. Lines with semicolons are handled as usual, but the indentation stacks, so the contents of a doubly nested do loop like `do i=1, 10; do j=1, 10` will be indented twice. I'm not sure how to handle this beyond telling the users to also switch on the `multiple-statements-per-line` rule, which will nest things properly.
- Continuation lines are ignored, as the user will probably want to set those themselves for readability. However, the first line _is_ adjusted, which might lead to some messy results.
- You can no longer choose multiple indent widths in the settings; only whether you indent something or not.

It's also currently failing tests, as I haven't yet implemented the ability to switch off indentation for modules/programs/control flow, etc. This would need to be handled at the `LogicalLines` level, as that keeps track of the expected indents on each line. Before implementing this, I thought I'd ask: are these the settings we want?  @ZedThree @connoraird @imciner2 I'd be grateful for your input.

A further issue is that it's indenting numbered statement labels rather than keeping them on the left-most column. We could add an option to always put them in the leftmost column, but then we run into issues if the users request an indent width of 1 or 2, and the statement label is `10`, e.g:
```f90
function f
  integer :: i, j
  do 10 i = i, 10
    do 10 j = 1, 10
      print *, i, j
  10 continue
end function f

! Becomes:

function f
  integer :: i, j
  do 10 i = i, 10
    do 10 j = 1, 10
      print *, i, j
10continue
end function f
```

If we always indent statement labels, as I'm doing now, this problem goes away. Again, feedback would be welcome here.

---

TODO, probably in another PR:
- Adjust all continued lines when the first continued line is edited
- Get `indent-width` from `Stylist` if not set by the user

LiamPattinson:
@ZedThree I might need some help figuring out how to get useful info from samply, as I'm currently struggling to filter out all the rayon noise.

Otherwise though, the performance of this method versus the original is pretty much the same as far as I can tell, both are completing the standard GS2 test I've been using in about 200ms.

This branch:

```
❯ hyperfine --warmup 3 -i '../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" --output-format=name src'
Benchmark 1: ../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" --output-format=name src
  Time (mean ± σ):     203.2 ms ±  11.7 ms    [User: 1006.9 ms, System: 60.5 ms]
  Range (min … max):   186.6 ms … 230.3 ms    16 runs
 
```

`main`:

```
❯ hyperfine --warmup 3 -i '../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" --output-format=name src'
Benchmark 1: ../fortitude/target/release/fortitude check --select=ALL --preview --file-extensions="f90" --output-format=name src
  Time (mean ± σ):     211.4 ms ±  16.9 ms    [User: 984.7 ms, System: 58.2 ms]
  Range (min … max):   187.6 ms … 256.8 ms    14 runs
 
```

Meanwhile, turning off the rule...

This branch:

```
❯ hyperfine --warmup 3 -i '../fortitude/target/release/fortitude check --select=ALL --ignore=S105,S106 --preview --file-extensions="f90" --output-format=name src'
Benchmark 1: ../fortitude/target/release/fortitude check --select=ALL --ignore=S105,S106 --preview --file-extensions="f90" --output-format=name src
  Time (mean ± σ):     150.0 ms ±   8.0 ms    [User: 828.6 ms, System: 34.6 ms]
  Range (min … max):   139.3 ms … 165.9 ms    21 runs
 
```

`main`:

```
❯ hyperfine --warmup 3 -i '../fortitude/target/release/fortitude check --select=ALL --ignore=S105,S106 --preview --file-extensions="f90" --output-format=name src'
Benchmark 1: ../fortitude/target/release/fortitude check --select=ALL --ignore=S105,S106 --preview --file-extensions="f90" --output-format=name src
  Time (mean ± σ):     158.1 ms ±   8.7 ms    [User: 840.5 ms, System: 39.3 ms]
  Range (min … max):   140.4 ms … 170.8 ms    18 runs
 
```

This one eats up a lot of run-time regardless of how you implement it, but as usual it's a tiny fraction of the tiny needed to write to terminal.

imciner2:
I think we really should keep the statement labels starting at the beginning of the line and not indent them, because it is very easy for them to get lost in the code when indented. In the case where the label is too close to the statement, then I think just putting 1 space in there is fine (maybe that could also generate a diagnostic warning this is happening).

crates/fortitude_linter/src/logical_lines.rs
```diff
@@ -0,0 +1,700 @@
...
+            // is zero.
+            let first_non_whitespace = line
+                .chars()
+                .find(|c| !c.is_ascii_whitespace())
+                .unwrap_or('\0');
+            let expected_indent = if first_non_whitespace == '#' {
+                0
+            } else {
+                self.current_expected_indent
+            };
```
connoraird, line 174: Would we be able to handle unindented line labels if we add some logic here for lines which start with an integer and have zero indentation to begin with?

crates/fortitude_linter/src/logical_lines.rs
```diff
@@ -0,0 +1,700 @@
...
+            self.lines.push(logical_line);
+            self.current_start_byte = end_byte;
+            self.current_expected_indent = self.next_expected_indent;
+        }
+
+        // Dedentors and post-indentors affect the expected indentation of the
+        // next logical line, so this must be performed after the current
+        // logical line has been finalised.
+        if DEDENTORS.contains(&kind_id) {
```
connoraird, line 287: Do we also need to add the other types of node (i.e. scoped zero indent nodes such as `contains`)? I think this shouldn't be too difficult to add. It would be something like

```rs
const SCOPED_ZERO_INDENTOR: [&str; 7] = [
    "contains_statement",
    "case_statement",
    "type_statement",
    "rank_statement",
    "else_clause",
    "elseif_clause",
    "elsewhere_clause",
];
...
if SCOPED_ZERO_INDENTOR.contains(&kind_id) {
	self.current_expected_indent = self.current_expected_indent.saturating_sub(1)
}
```

connoraird: Oh I see this has been done by dedenting and then re-indenting.

crates/fortitude_linter/src/rules/style/whitespace.rs
```diff
@@ -404,418 +389,96 @@ impl AlwaysFixableViolation for IndentedPreprocessorStatement {
...
+            continue;
+        }
 
-            // Include previous semicolon if present
-            line_segment_start = if !is_first_segment || line.starts_with(';') {
-                line_segment_start - TextSize::new(1)
-            } else {
-                line_segment_start
-            };
+        let expected_indent = line.expected_indent(indent_width.as_u8());
```
connoraird, line 417: Could we maintain the ability to specify specific indentations for each node type here if we extract the starting node type for the logical line and then match it's kind to the setting just like we are doing in the code atm?

connoraird: We could add a `node_kind` attribute to the definition of LogicalLine which can be set when constructing the LogicalLines.

connoraird (commented):
This is a cool implementation. I think there should be some logic in the current implementation to help with setting the indentation of labelled lines.

crates/fortitude_linter/src/rules/style/whitespace.rs
LiamPattinson: (Currently on leave so won't be working on this for a bit)

I'm planning on reimplementing your number-of-indents thing here, as I think when I wrote this I'd forgotten what the intent of it actually was. I'll aim to restore it exactly to what you wrote via an interactive rebase. And I think you're right, that's probably how I'll implement it.

Issue #735, opened 2026-08-20
[Bug] Incorrect cyclomatic complexity calculation for `select`/`case` statements


cyrilgandon:
The current cyclomatic complexity rule for Fortran `switch`/`case` statements in Fortitude [[introduced in #596](https://github.com/PlasmaFAIR/fortitude/issues/596)](https://github.com/PlasmaFAIR/fortitude/issues/596) incorrectly increments complexity **per `case` statement** rather than per `select` structure. This leads to:

- A `select` with 10 cases → **+10** to complexity (should be **+1**)
- A `select` with 100 cases → **+100** to complexity (should be **+1**)
- False positives for large `select` statements

This behavior:
1. **Is inconsistent with modern tools** (SonarQube, PMD, Visual Studio, etc., all count +1 per `select/switch`).
2. **Is counterintuitive**—a `select/switch` with 100 cases is not 100x more complex than an `if/else if` with 100 conditions.

### **Proposed Fix**
Update the rule to count **+1 for the entire `select` structure** instead of per `case`:
```rust
// Current (incorrect):
"case_statement" => complexity += 1,

// Proposed (correct):
"select_statement" => complexity += 1,  // +1 for the entire structure
```

Référence: Cognitive Complexity a new way of measuring understandability. By G. Ann Campbell.
> Switches
> A switch and all its cases combined incurs a single structural increment.
> Under Cyclomatic Complexity, a switch is treated as an analog to an if-else
> if chain. That is, each case in the switch causes an increment because it
> causes a branch in the mathematical model of the control flow.
> But from a maintainer’s point of view, a switch - which compares a single
> variable to an explicitly named set of literal values - is much easier to
> understand than an if-else if chain because the latter may make any
> number of comparisons, using any number of variables and values.
> In short, an if-else if chain must be read carefully, while a switch can often
> be taken in at a glance.

ZedThree:
Thanks @cyrilgandon, sounds like a reasonable change and in line with other tools. Do you want to make a PR?

cyrilgandon:
Yes I'm on it !

PR #736, opened 2026-08-20
fix: count select/case as single complexity increment, not per-branch

cyrilgandon:
Fixing https://github.com/PlasmaFAIR/fortitude/issues/735


ZedThree (approved):
Thanks @cyrilgandon!

PR #738, opened 2026-08-27
Add missing-kind-argument rule to check for explicit kind in intrinsic

imciner2:
This is a new rule that flags the use of certain intrinsic functions where the default return kind parameter is not based on the kind of the numeric variable on the input - e.g., `real` is a single precision return by default, even if it gets a double precision passed into it. In these cases, there is an optional `kind=` argument that specifies the kind to use for the output, so this rule enforces the use of that argument to always make clear what kind should be used by the return.

It seemed like `C023` seemed like the best place (next to other ones about kinds), but we could update the number if it might fit in a different place better.

In addition to these intrinsics, there are also a few more that have a `kind` argument that affects the return, like `ACHAR`, `CHAR`, `COSHAPE`, `FIND_LOC`, and others. The ones I highlighted here are the ones that I think are the most problematic.

---
Generated with help from GitHub Copilot (with lots of manual review and tweaking).


crates/fortitude_linter/resources/test/fixtures/correctness/C023.f90
```diff
@@ -0,0 +1,20 @@
...
+
+  integer :: i
+  real :: x, y
+  complex :: z
+
+  x = real(i)                 ! catch
+  i = int(x)                  ! catch
+  z = cmplx(x)                ! catch
+
+  x = real(i, 8)              ! catch
```
ZedThree, line 12: Shouldn't this be ignored because the second argument is the kind? Unless the rule is that these functions should explicitly use the `kind=` keyword argument?

imciner2: > Shouldn't this be ignored because the second argument is the kind? Unless the rule is that these functions should explicitly use the kind= keyword argument?

This was my intent with the rule - IMO optional arguments should be specified like that to be clearer, so I wrote the rule to basically enforce it that way.

We could make it support the syntax without `kind=` in the argument list, but that will require knowing where it falls in each intrinsic (e.g., it is the second argument for `real`, but the third in `cmplx` if there are no labels given), so this way was also slightly easier to implement.

ZedThree: I mean, I do like that as a rule, requiring optional arguments to use a keyword arg, but that feels like a separate rule to this one?

LiamPattinson: I agree with @ZedThree, I think it should be a separate `style` rule. The case of `real(5, 8)` is unclear but not a correctness problem, the meaning of `real(5, dp)` should be self-evident, while `real(5, kind=dp)` is better again.

PR #739, opened 2026-09-02
Change how `incorrect-indentation` options work

ZedThree:
Two main changes:

- Previously, `incorrect-indentation.*-indents` specified the number of _indents_ to apply to each node.
  Now, `incorrect-indentation.*-indent` (no `s`) specifies the number of _spaces_ to indent.

  This gives a more consistent behaviour: the user can set `indent-width` to 4, and then change, say, `control-flow-indent = 3` to indent control flow nodes by 3 spaces, and leave everything else at 4.

- Deprecate `check.invalid-tab.indent-width` in favour of `check.indent-width`, and use the latter as the default value for `incorrect-indentation.*-indent`. Previously, there was some hierarchy of fallback values. This is more straightforward, and gives a deprecation warning encouraging users to change if the old key is set

This doesn't add the file-local indentation discovered by `Stylist` -- there was some complicated behaviour that I didn't quite understand, so I've left it for the time being.


LiamPattinson (approved):
Thanks @ZedThree, all looks good to me. I'll work on rebasing https://github.com/PlasmaFAIR/fortitude/pull/734 on top of this when I have a spare moment.

